PR #48, opened 2022-10-07
Automating semver PATCH bump of helm chart

mschenck:
Created `helm_bump_version` job to automate patch-level version bumping of helm chart

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -59,15 +59,11 @@ jobs:
           version: "${{ steps.helm_chart.outputs.version }}"
 
       - name: Update Chart version
-        id: replace
         uses: jacobtomlinson/gha-find-replace@v2
         with:
           include: "helm/ingress-controller/Chart.yaml"
           find: '(?m)^version: .*$'
           replace: "version: ${{ steps.next_version.outputs.patch }}"
-
-      - name: debugging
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -32,11 +32,48 @@ jobs:
...
+          version: "${{ steps.helm_chart.outputs.version }}"
+
+      - name: Update Chart version
+        uses: jacobtomlinson/gha-find-replace@v2
+        with:
+          include: "helm/ingress-controller/Chart.yaml"
+          find: '(?m)^version: .*$'
+          replace: "version: ${{ steps.next_version.outputs.patch }}"
+
+      - name: Commit Chart.yaml version bump
```
alex-bezek, line 68: Won't this commit cause another invocation of the github action and make a cyclical loop? 
I didn't realize we couldn't push over the same version, but i guess it wouldn't be too hard to just include a bump in that version as a part of the pr or as a separate pr? That could be annoying though with merge conflicts I guess 🤔 . I guess helm just doesn't really have the same `:latest` concept as docker. 

If it doesn't make a loop, this seems good for now. It will let us easily test the latest constantly, and later on I think russ and alan probably want to unify the release process for the open source repos.

alex-bezek (approved):

PR #50, opened 2022-10-10
Only helm release on Chart.yaml changes

mschenck:
separated docker and helm release, helm now only triggers when a merge to main includes a change to the Chart.yaml


mschenck:
You can see the result [here](https://github.com/ngrok/ngrok-ingress-controller/blob/gh-pages/index.yaml)

alex-bezek (approved):

PR #52, opened 2022-10-12
Make assorted changes to make code more idiomatic

euank:
This makes the following changes:
1. Switches to cobra, which is better than the goflags package
2. Removes some refs/derefs which felt unidiomatic. Generally, `*x` shouldn't happen very often, and so I pretty much shuffled around types to make that less common.
3. Updates domain list iteration to be more idiomatic
4. Other misc stuff I'm sure

I've commented on most of the changes below to indicate why they're there.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
@@ -30,9 +31,9 @@ import (
 
 	"github.com/ngrok/ngrok-ingress-controller/internal/controllers"
 	"github.com/ngrok/ngrok-ingress-controller/pkg/ngrokapidriver"
+	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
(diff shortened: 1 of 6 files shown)


cmd/manager/main.go
```diff
@@ -49,52 +50,76 @@ var (
 )
 
 func init() {
-	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	if err := clientgoscheme.AddToScheme(scheme); err != nil {
+		panic(err)
+	}
 
-	podNamespace, ok := os.LookupEnv("POD_NAMESPACE")
```
euank, line 54: Configuration loading should typically happen somewhere below `main`, with `init` being as small as possible. This was removed to reduce the size of `init`, and put that loading in a more idiomatic location.

Notably, this also removes some global variables (like `namespace`), which are also best avoided when possible.

cmd/manager/main.go
```diff
@@ -49,52 +50,76 @@ var (
...
+		setupLog.Error(err, "error running manager")
+		os.Exit(1)
 	}
-	namespace = podNamespace
+}
 
-	k, ok := os.LookupEnv("NGROK_API_KEY")
-	if !ok {
-		panic("NGROK_API_KEY is not set")
+type managerOpts struct {
```
euank, line 67: flags/config got moved into this struct, which feels slightly cleaner. :man_shrugging:

cmd/manager/main.go
```diff
@@ -105,8 +130,7 @@ func main() {
 		Namespace:      namespace,
 		NgrokAPIDriver: ngrokapidriver.NewNgrokApiClient(ngrokApiKey),
 	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "ingress")
-		os.Exit(1)
+		return fmt.Errorf("unable to create ingress controller: %w", err)
```
euank, line 133: logging got moved up to `main`, where it just prints the error and exits.
This removes a little boilerplate, but I'm not totally tied to that decision. It does seem slightly cleaner code-wise, but it means we can't easily provide context (key-value logging pair context I mean) to the messages. Still, we could add additional messages if we wanted that.

cmd/manager/main.go
```diff
@@ -115,35 +139,32 @@ func main() {
...
 	// For controller level configs, this may be the recommended way though https://book.kubebuilder.io/reference/markers.html
 	// We can't use this though to write config maps, and the mgr.GetClient() doesn't work because the cache isn't initialized
 	config := &v1.ConfigMap{}
 	if err := mgr.GetAPIReader().Get(ctx, types.NamespacedName{Name: configMapName, Namespace: namespace}, config); client.IgnoreNotFound(err) != nil {
-		setupLog.Error(err, fmt.Sprintf("error getting config map named '%s'", configMapName))
-		os.Exit(1)
+		return err
 	} else {
-		setupLog.Info(fmt.Sprintf("Found config map named '%s' %+v", configMapName, config))
+		setupLog.Info(fmt.Sprintf("Found config map named %q %+v", configMapName, config))
```
euank, line 152: total nit thing, but the `%q` fmt verb does quoting for you.

internal/controllers/ingress_controller.go
```diff
@@ -64,7 +64,7 @@ func (irec *IngressReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	// The object is being deleted
 	if !ingress.ObjectMeta.DeletionTimestamp.IsZero() && controllerutil.ContainsFinalizer(ingress, finalizerName) {
-		return irec.DeleteIngress(ctx, *edge, ingress)
+		return irec.DeleteIngress(ctx, edge, ingress)
```
euank, line 67: I did type-shuffling to get rid of the dereference here. It feels unidiomatic to deref, and there didn't seem to be a reason not to just make it a pointer throughout.

pkg/agentapiclient/tunnel.go
```diff
@@ -9,52 +9,59 @@ import (
 	"time"
 )
 
-const tunnelApiUrl = "http://localhost:4040/api/tunnels"
+const tunnelAPIURL = "http://localhost:4040/api/tunnels"
```
euank, line 12: Api -> API and Url -> URL are the idiomatic spellings. Go generally prefers acronyms to be capitalized, which is weird and different from every other language, but that's the idiom.

See https://github.com/golang/go/wiki/CodeReviewComments#initialisms

pkg/agentapiclient/tunnel.go
```diff
@@ -9,52 +9,59 @@ import (
...
+func NewAgentApiClient() AgentAPIClient {
+	return agentAPIClient{
+		client: &http.Client{Timeout: 1 * time.Second},
 	}
 }
 
-func (ac agentApiClient) CreateTunnel(_ context.Context, t TunnelsApiBody) error {
-	resp, err := ac.client.Get(tunnelApiUrl + "/" + t.Name)
+func (ac agentAPIClient) CreateTunnel(ctx context.Context, t TunnelsAPIBody) error {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, tunnelAPIURL+"/"+t.Name, nil)
```
euank, line 30: This is context plumbing, every http request should have a context ideally.

pkg/ngrokapidriver/driver.go
```diff
@@ -56,27 +57,34 @@ func (nc ngrokAPIDriver) FindEdge(ctx context.Context, id string) (*ngrok.HTTPSE
...
 		Metadata:    napi.metadata,
 	})
 	// Swallow conflicts, just always try to create it and don't delete them upon ingress deletion
 	if err != nil {
-		if strings.Contains(err.Error(), "ERR_NGROK_413") || strings.Contains(err.Error(), "ERR_NGROK_7122") {
-			log.Info("Reserved domain already exists, skipping creation", "domain", domain)
-		} else {
-			return nil, err
+		var nerr *ngrok.Error
+		if errors.As(err, &nerr) {
```
euank, line 76: Using `ngrok.Error.ErrorCode` is better than substring matching on the whole error text.

pkg/ngrokapidriver/driver.go
```diff
@@ -183,28 +191,24 @@ func (nc ngrokAPIDriver) GetReservedDomains(ctx context.Context, edgeID string)
 	}
 	hostPortDomains := []string{}
 	for _, hostport := range *edge.Hostports {
-		hostPortDomains = append(hostPortDomains, strings.Split(hostport, ":")[0])
+		domain, _, err := net.SplitHostPort(hostport)
```
euank, line 194: I mean `net.SplitHostPort` exists so I guess we should use it why not. :man_shrugging:

pkg/ngrokapidriver/driver.go
```diff
@@ -183,28 +191,24 @@ func (nc ngrokAPIDriver) GetReservedDomains(ctx context.Context, edgeID string)
...
 	// Loop while there are more domains and check if they match any of the hostPortDomains. If so add it to the reservedDomains
-	for {
-		if !domainsItr.Next(ctx) {
-			err := domainsItr.Err()
-			if err != nil {
-				return nil, err
-			}
-			break
-		}
+	for domainsItr.Next(ctx) {
```
euank, line 204: this `for` construct, and then doing a single `if iter.Err() != nil` after the loop, is the more idiomatic way to use an iterator than relying on a `break` to escape a `while true` loop.

pkg/ngrokapidriver/types.go
```diff
@@ -15,6 +15,3 @@ type Route struct {
 	Modules   []any
 	Labels    map[string]string
 }
-
-type Module interface {
```
euank, line 19: unused code I think?

cmd/manager/main.go
alex-bezek: Nice! I _think_ long term we want to move these docs to this https://book.kubebuilder.io/component-config-tutorial/tutorial.html based on what they say in their docs, but i haven't dug into it much. This is better for now though

cmd/manager/main.go
```diff
@@ -49,52 +50,76 @@ var (
...
+		RunE: func(c *cobra.Command, args []string) error {
+			return runController(c.Context(), opts)
+		},
 	}
-	ngrokApiKey = k
 
-	//+kubebuilder:scaffold:scheme
+	c.Flags().StringVar(&opts.metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to")
+	c.Flags().StringVar(&opts.probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	c.Flags().BoolVar(&opts.enableLeaderElection, "leader-elect", false, "Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
```
alex-bezek, line 90: Not relevant to your PR, but @mschenck maybe we should just remove this? There shouldn't be a case when you can disable leader election, things would break i believe. I'll make a follow up note

cmd/manager/main.go
alex-bezek: This seems fine to me. we can still add context at the manager level, which probably is most appropriate since its the manager failing to add the new reconciler.

cmd/manager/main.go
alex-bezek: oooh nice

internal/controllers/ingress_controller.go
alex-bezek: On an internal separate pr, I was just asking about this in a general sense. Curious if you have some info and i can go do some reading on my own.

> Can you explain why you changed this? Not that its right or wrong, I just don't know best practices on when to use pointers for objects vs the object itself. Admittedly, sometimes i go with whatever makes the red line go away but I think there's some deeper reasons people have sometimes like the size of the object. If anyone has some knowledge to drop, i'd be curious

alex-bezek (approved):
Thanks! i definitely learned a couple of things :)

cmd/manager/main.go
alex-bezek: is there a reason we keep the schema setup in the init vs the main? My vague understanding of init is that you use it mostly for running code whenever a package is installed (like i think pulling in a mysql package has an init that registers the driver or something so its ready before you use it?). In our case, its a main package, so i wonder if it should be used at all?

cmd/manager/main.go
euank: Yup, indeed.

internal/controllers/ingress_controller.go
euank: I'll put my opinion on the general thing here, but you may find other people have slightly different opinions.

In general, if something has methods on it, it ends up _often_ being less confusing to use a pointer receiver for those methods, and if it has pointer receiver methods, I personally think it also makes sense to use the struct as a reference since I don't really like the implicit `t.foo()` reference that happens for a non-pointer struct w/ a pointer-receiver method.

The reason it's often less confusing to have a pointer receiver is the following:

https://go.dev/play/p/hRB_lcPMJmz

If you're going to have a struct that is just immutable data or where you know you're doing interior mutability right, then it's nice to use it as a non-pointer value in order to encourage the compiler to let it live on the stack, and so people know it can't be nil. In reality, the compiler will take pointer/non-pointer as more of a hint than anything else, and internally may allocate it on the stack/heap as appropriate based on escape analysis for the value.

The size of the struct also matters, as you alluded to, but you can find plenty of people micro-optimizing and writing about that online already.

Since pointers have less confusing mutation semantics, and heap-allocation is often better for sufficiently large structs, I personally find myself defaulting to using references fairly often.

-------

Note that all the talk of pointer receiver methods above is moot in this case since the `ngrokapidriver.Edge` type has no methods on it.
In this case, it was about making it consistently a pointer or not, and me defaulting to pointer since that's the easier default, especially since it was already a pointer at the place I removed the deref.

cmd/manager/main.go
euank: Done

PR #61, opened 2022-12-06
Add custom user agent

### LINKED ISSUE #49: Set a user agent specific to the ingress controller
we should set the user agent on api requests so we can look up the edges that were created by this

jonstacks:
Fixes #49 

Currently setting it `ngrok-ingress-controller/v1-alpha` but open to any other suggestions or if we want to tackle compiling it with a version at this time.

### DIFF
--- a/pkg/ngrokapidriver/driver.go
+++ b/pkg/ngrokapidriver/driver.go
@@ -6,11 +6,11 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/ngrok/ngrok-api-go/v4"
-	tgb "github.com/ngrok/ngrok-api-go/v4/backends/tunnel_group"
-	edge "github.com/ngrok/ngrok-api-go/v4/edges/https"
-	edge_route "github.com/ngrok/ngrok-api-go/v4/edges/https_routes"
-	"github.com/ngrok/ngrok-api-go/v4/reserved_domains"
+	"github.com/ngrok/ngrok-api-go/v5"
+	tgb "github.com/ngrok/ngrok-api-go/v5/backends/tunnel_group"
+	edge "github.com/ngrok/ngrok-api-go/v5/edges/https"
+	edge_route "github.com/ngrok/ngrok-api-go/v5/edges/https_routes"
+	"github.com/ngrok/ngrok-api-go/v5/reserved_domains"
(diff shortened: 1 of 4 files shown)


pkg/ngrokapidriver/driver.go
```diff
@@ -34,8 +34,7 @@ type ngrokAPIDriver struct {
 }
 
 func NewNgrokAPIClient(apiKey string, region string) NgrokAPIDriver {
-	// TODO: Add in a unique user agent here
-	config := ngrok.NewClientConfig(apiKey)
+	config := ngrok.NewClientConfig(apiKey, ngrok.WithUserAgent("ngrok-ingress-controller/v1-alpha"))
```
alex-bezek, line 37: In ruby I remember we used to pull the version from a constant that was used throughout the code, any chance there is something similar we could do here? Using our milestones until GA makes sense though in the meantime

alex-bezek (approved):

pkg/ngrokapidriver/driver.go
```diff
@@ -90,7 +89,7 @@ func (napi ngrokAPIDriver) CreateEdge(ctx context.Context, edgeSummary *Edge) (*
 	}
 
 	newEdge, err := napi.edges.Create(ctx, &ngrok.HTTPSEdgeCreate{
-		Hostports:   &[]string{edgeSummary.Hostport},
```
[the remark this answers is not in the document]
jonstacks: This was correct when we were using `ngrok-api-go/v4`. The latest release introduced some breaking changes to be more idiomatic since slices are pointers to underlying arrays. Its called out in the [v5 release notes](https://github.com/ngrok/ngrok-api-go/releases/tag/v5.0.0)

pkg/ngrokapidriver/driver.go
jonstacks: A clean way that I've seen to do this is to have it be injected at build time as described [here](https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-version-information-for-go-applications) with the use of `ldflags`. I was thinking this would be a good thing to implement once we get some more automation on the release process and using a tag to inject this at build time. If you feel differently though, let me know.

Issue #62, opened 2022-12-07
Various review Feedback updates


alex-bezek:
From our product review meeting, we should make the following updates:
* switch replica count to 1: free
* remove log stdout option
* credential secret should name should just be 1 value
* add server address as a config option
* Should we explicitly require opting into the default credentials or explicitly specify a name. Could also tell them post install if that secret doesn't exist
* allow seting default ingress class via helm file

PR #66, opened 2022-12-09
A few cleanup items

alex-bezek:
Resolves: #62

These are a few items we wanted to clean up from a review session:
* change up the helm values api for setting the secret so its more explicit when its creating a secret or not
* remove the ability to configure the agent logging
* change replicas to 1 for a more default/demo setup
* handle non 200/404 error codes for the agent tunnel api so errors are requeued


helm/ingress-controller/templates/credentials-secret.yaml
```diff
@@ -1,10 +1,12 @@
-{{- if or (not (empty .Values.apiKey)) (not (empty .Values.authtoken)) }}
+{{- if eq .Values.credentials.secret.create true }}
```
alex-bezek, line 1: thinking about this a little bit more, I probably want to do a requires for this create value if either of the 2 credentials are set so the user gets a really obvious error that you passed in credentials but didn't specify create to avoid the footgun of having this just do nothing

helm/ingress-controller/values.yaml
```diff
@@ -14,17 +14,19 @@ image:
...
 
 # The name of the K8S secret that contains the credentials for the
 # ingress controller.
-credentialsSecret:
-  name: ngrok-ingress-controller-credentials
-apiKey: ""
-authtoken: ""
+credentials:
+  secret:
+    create: false
```
alex-bezek, line 26: @inconshreveable we are updating how the credential is created via helm values based on some a review session we had a bit ago. I recall we discussed how it should be more explicit when its creating the secret or not. Am I remembering correctly, is this what you were getting at? Working through this a bit more, I do find it a bit cumbersome to have to specify the credentials *AND* have to say to create the secret since if I'm providing the values, it feels like a given that those should be used. Let me know. your thoughts on this, how its used can bee seen in the readme docs https://github.com/ngrok/ngrok-ingress-controller/pull/66/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R69

helm/ingress-controller/templates/credentials-secret.yaml
```diff
@@ -1,10 +1,12 @@
...
+{{- if or (not (empty .Values.credentials.apiKey)) (not (empty .Values.credentials.authtoken)) }}
```
jonstacks, line 2: If both of these are required when creating the secret, I think its fine to remove this line. Otherwise, if you set `secrete.create=true` and leave both `apiKey` and `authToken` empty, you wouldn't hit the `required` checks.

alex-bezek: ahh thanks, yeah they should both be required if the create is set to true. Will remove this

helm/ingress-controller/values.yaml
```diff
@@ -14,17 +14,19 @@ image:
...
+    name: ngrok-ingress-controller-credentials
+  apiKey: ""
+  authtoken: ""
```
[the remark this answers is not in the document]
alex-bezek: i can't find the thread or PR, but I believe i tried that at one point and got corrected that we treat it as a single word not two everywhere, so we should continue for consistency

.gitattributes
```diff
@@ -0,0 +1 @@
+helm/**/__snapshot__/**  linguist-generated=true
```
jonstacks, line 1: This is cool :sunglasses:

README.md
```diff
@@ -66,8 +66,9 @@ helm repo add ngrok https://ngrok.github.io/ngrok-ingress-controller
 helm install ngrok-ingress-controller ngrok/ngrok-ingress-controller \
   --namespace ngrok-ingress-controller \
   --create-namespace \
-  --set apiKey=$(NGROK_API_KEY) \
-  --set authtoken=$(NGROK_AUTHTOKEN)
+  --set credentials.secret.create = true \
```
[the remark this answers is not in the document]
alex-bezek: added this and used the overall name helper instead of just the release. let me know what you think

helm/ingress-controller/scripts/gen-agent-config.sh
```diff
@@ -1,14 +1,11 @@
 #!/bin/bash
 
-NGROK_LOG="${NGROK_LOG:-stdout}"
-NGROK_REGION="${NGROK_REGION:-us}"
```
alex-bezek, line 4: this is hard to test and we are moving away from the agent so i didn't add much here. But what i did do was ran the e2e tests, checked in the dashboard that the edge was defaulted still using US and catted out my agent file in the pod to see its set to empty and the agent is fine with that

```
k exec -it pod/ngrok-ingress-controller-manager-6d7dbd47b7-7gpgb -c ngrok-cli -- /bin/bash

ngrok@ngrok-ingress-controller-manager-6d7dbd47b7-7gpgb:/$ cat /var/lib/ngrok/agent-template.yaml
version: 2
authtoken: REDACTED
console_ui: false
region:
remote_management: true
update_check: false
```

helm/ingress-controller/tests/credentials-secret_test.yaml
```diff
@@ -1,10 +1,35 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/quintush/helm-unittest/master/schema/helm-testsuite.json
```
alex-bezek, line 1: https://github.com/quintush/helm-unittest#test-suite-code-completion-and-validation I noticed this and it helped a bit while writing tests. it didn't seem easy to setup the other way in vscode

helm/ingress-controller/values.yaml
```diff
@@ -22,7 +22,7 @@ podAnnotations: {}
 ## @param replicaCount The number of controllers and agents to run.
 ## A minimum of 2 is recommended in production for HA.
 ##
-replicaCount: 2
+replicaCount: 1
```
alex-bezek, line 25: a side affect of this is that locally now we are running 1 container. I can override this easily in the makefile but i was curious if others had opinions first. On one hand, running a single container locally is simpler and faster to debug and work with, but on the other hand, it doesn't exercise our leader election important codepaths and some things may work in the context of a single agent but not multiple.

jonstacks (approved):

PR #69, opened 2022-12-12
Setup testing for helm

### LINKED ISSUE #36: Helm Lint & docs generation
We should lint our helm chart as a commonly good practice. 

Additionally, if we include the right comments in our values.yaml file, there are github actions and other projects that can generate markdown table docs of our values. This issue should set that up as well to publish to a markdown file in the repo.

jonstacks:
Resolves #36 

Adds helm `unittest` plugin to project with snapshots so we can validate helm generates what we want it to

### DIFF
--- a/helm/ingress-controller/.helmignore
+++ b/helm/ingress-controller/.helmignore
@@ -1,0 +1,25 @@
+# Source: https://github.com/helm/helm/blob/main/pkg/repo/repotest/testdata/examplechart/.helmignore
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+
+# helmtest plugin tests
+tests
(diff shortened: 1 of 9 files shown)


helm/ingress-controller/.helmignore
```diff
@@ -0,0 +1,25 @@
+# Source: https://github.com/helm/helm/blob/main/pkg/repo/repotest/testdata/examplechart/.helmignore
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
```
alex-bezek, line 5: thoughts on converting this to an allow list instead of block list?
https://github.com/ngrok/ngrok-ingress-controller/blob/main/.dockerignore

helm/ingress-controller/tests/__snapshot__/controller-rbac_test.yaml.snap
```diff
@@ -0,0 +1,103 @@
+Should generate rbac:
```
alex-bezek, line 1: nit ^ the `.keep` file doesn't seem necessary anymore

helm/ingress-controller/.helmignore
jonstacks: I'd probably err on the side of having extra tests files getting included in a release and us fixing it in a dot release than accidentally publishing something that doesn't include a new support files folder. We could move to an allow list, but we'd need some way to test helm packaging locally and make sure we are installing from the package first IMO so we catch not allowing files through that need to be bundled in the chart

helm/ingress-controller/tests/__snapshot__/controller-rbac_test.yaml.snap
jonstacks: Fixed in https://github.com/ngrok/ngrok-ingress-controller/pull/69/commits/ae6973830a6227706075691ce69609a8b700ea5b

PR #71, opened 2022-12-12
Fail CI if Kubebuilder generated files aren't up to date

### LINKED ISSUE #34: Have CI run gen commands and fail on diffs
Kubebuilder supports auto generating some files like rbac permissions based on annotations in the [code](https://github.com/ngrok/ngrok-ingress-controller/blob/main/internal/controllers/ingress_controller.go#L30-L35). We should have CI run these generation commands and fail CI if there are any diffs so we don't forget to generate and check those in.

A job or step in our workflows should run the gen commands and ensure no git diffs are produced. If there are diffs produced, the CI job should fail

jonstacks:
Fixes #34 

Will fail CI if Kubebuilder generated files aren't up to date. I moved the other tests to depend on this one so that if the kubebuilder generated files aren't up to date, we will skip the other tests. Also moved the docker buildx builds to depend on it as well. This should help save runner minutes when there are small issues that need to be fixed first.


.github/workflows/ci.yaml
```diff
@@ -5,9 +5,54 @@ on:
...
+    name: Kubebuilder Diff
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.19 # Use the latest
+    - name: Generate manifests
+      run: make manifests
```
alex-bezek, line 23: there is technically another generate command too https://github.com/ngrok/ngrok-ingress-controller/blob/main/Makefile#L61 it was just created with kubebuilder though and I'm not sure we are even leveraging it until we make CRDs

.github/workflows/ci.yaml
```diff
@@ -5,9 +5,54 @@ on:
...
+      run: |
+        git diff --exit-code
+        git diff --cached --exit-code
+
+  build-images:
+    name: Build Images
+    needs:
+    - kubebuilder-diff
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
```
alex-bezek, line 34: these are pretty arbitrary I assume. Are they just a good practice safeguard to prevent an action from running rogue and burning through your build minutes?

alex-bezek (approved):

.github/workflows/ci.yaml
jonstacks: Yep. I had a bad experience so now do it as a safeguard against accidents

Issue #39, opened 2022-10-05
Setup Unit and Integration tests


alex-bezek:
While we have our e2e suite to test overall functionality, we should also setup lower level tests. 
* Plain go unit tests where they provide value
* Kubebuilder tests frameworks
  * https://betterprogramming.pub/write-tests-for-your-kubernetes-operator-d3d6a9530840
  * https://book.kubebuilder.io/reference/envtest.html?highlight=envtest#configuring-envtest-for-integration-tests
 * Other test frameworks like k6s or Ginkgo for integration style tests

jonstacks:
Closing this as resolved for now since we have unit tests and the e2e tests. Will re-open when we decide we need kubebuilder or ginko style tests.

PR #74, opened 2022-12-14
Add unit tests for ingress controller & tunnel controller

jonstacks:
Related: #39 

## What

This change adds in some tests for both the ingress controller and tunnel controller to validate the edge and tunnel API objects we are creating.

## How

Test that given certain ingress objects, they return the respective edge and tunnel API objects.


mschenck (approved):
this looks great to me, thanks for getting this test coverage effort going!

Issue #60, opened 2022-12-05
Formalize release process for artifacts


alex-bezek:
Currently we have some github actions that will release new `latest` docker images on each push to main and it will publish a new release of our helm chart on each new version bump in its chart.yaml. 

We need to formalize a release process for these artifacts, make sure its fully automated using common community best practices, and document it in our readme.

PR #76, opened 2022-12-15
Add initial Release Process documentation

jonstacks:
Contributes to: #60 

## What

Documentation on our release process

## How

Describes what artifacts we will build and release, where they are published to, how we will version them, and their how it maps to git tags. We can then automate our release process to conform with our documentation.

## Breaking Changes
No


mschenck (approved):
🔥

PR #78, opened 2022-12-19
Automate docker image release

### LINKED ISSUE #60: Formalize release process for artifacts

jonstacks:
Resolves #60 

## What

Automates building and releasing tagged docker image artifacts when the VERSION file is changed.

## How

When a commit is pushed to `main` that changes the `VERSION` file, the following will happen:
* A tagged docker image matching the version will be pushed to Dockerhub.
* A new GitHub release will be created as described in the release documentation.

## Breaking Changes
No, but the `VERSION` file is being bumped to test the automation and to get a tagged version out that has recent changes.


.github/workflows/ci.yaml
```diff
@@ -138,22 +142,3 @@ jobs:
       run: make helm-lint
     - name: Helm Unit Tests
       run: make helm-test
-
-  # Action that will automatically comment on PRs based on changes
-  pr-comments:
```
jonstacks, line 143: Moved to a separate pr-comments workflow since we need `on: pull_request_target` so we can comment on PRs across forks.

alex-bezek (approved):

PR #79, opened 2022-12-19
Fix for pr-commenting

jonstacks:
Related: #60 

## What

Fixes PR comments not working as seen in #78.

## How

Due to how `GITHUB_TOKEN` permissions are scoped for pull_requests, we need to check this in first and run it with the `pull_request_target` event type so that it can comment on PRs

## Breaking Changes

No


mschenck (approved):
👍

PR #81, opened 2022-12-19
Trigger update when annotations change

jonstacks:
## What

Updates predicate filters so we also capture changes to annotations

## How

Annotations are outside of the `Ingress`'s `spec` field so they don't trigger a change in the generation. This compares the old annotations to the new annotations to figure out if we need to do a reconcile.

## Breaking Changes
No


mschenck (approved):
I'm not totally sure we want these unified, but I see no issue with it today.

PR #82, opened 2022-12-19
Fix for creating Github release

jonstacks:
Contributes to: #60 

## What

Missed the `.rest` being added in newer versions of github-script. 

## How

Update to match what github-script@v5+ expects

## Breaking Changes

No, but bumping the `VERSION` once again, since pre-release versions are cheap, to get a git tag out there.


mschenck (approved):
🔥

Issue #55, opened 2022-12-05
Helm Common Values


alex-bezek:
We should audit and support the common helm values other controllers support.

jonstacks:
Going to pick this one up next since it should help test the helm chart release automation.

PR #83, opened 2022-12-19
Add Helm Common Values

jonstacks:
Resolves: #55 

## What

Adding more helm common values like overriding the `image.registry` and adding `image.pullSecrets`. See changes in the `values.yaml` file.

## How

* Added additional helm template snapshots for future testing. About half of this PR are tests.
* Updated the command line args of the controller to support taking an `election-id` since we support overriding the name now that is used to prefix resources.

## Breaking Changes
Yes, specifically the names of resources are now prefixed to support overriding with `nameOverride` and `fullnameOverride` as is common in a lot of helm charts.


mschenck (approved):
Nice!  I really like the leader election revamp

Issue #63, opened 2022-12-07
Make a little magic happen


alex-bezek:
as a part of the notes.txt that runs after the helm installation, we should output some useful information about next steps the user could take to see some useful functionality of ngrok very quickly. This is open to other ideas but the one that was brought up was to query the cluster for the k8s dashboard and if present output an example of the ingress object needed to expose it behind oauth

alex-bezek:
Another option would be to look for services on port 80 and 443 as good potential options to output an example for. Or even a generic example

mschenck:
I'm happy to close this ticket for now, but I'm leaving it open while looking for some additional ideas for _magic_.

mschenck:
closing for now.  More enhancements down the road

PR #84, opened 2022-12-20
Added example Ingress manifest to NOTES.txt

mschenck:
Selecting first service (other than "kubernetes") that is listening on 80 or 443 and generating a basic Ingress object manifest

related: #63

## What
Providing a ready-to-go example `Ingress` manifest so that new users can easily start to leverage the ingress controller, right after installation.

## How
The template now queries all K8s Services (across all namespaces) and selects the first Service that is listening on either 80 or 443 (that is not the "kubernetes" API service).  Then it generates an `Ingress` manifests, from the selected service. 

## Breaking Changes
No

### DIFF
--- a/helm/ingress-controller/templates/NOTES.txt
+++ b/helm/ingress-controller/templates/NOTES.txt
@@ -1,6 +1,51 @@
-The ngrok Ingress controller has been deployed as a Deployment type to your cluster.
+================================================================================
+The ngrok Ingress controller has been deployed as a Deployment type to your
+cluster.
 
-If you haven't yet, create some Ingress resources in your cluster and they will be automatically configured on the internet using ngrok.
+If you haven't yet, create some Ingress resources in your cluster and they will
+be automatically configured on the internet using ngrok.
+
+
+{{- $found := false }}
+{{- range $svcIndex, $service := (lookup "v1" "Service" "" "").items }}
+    {{- range $portMapIdx, $portMap := $service.spec.ports }}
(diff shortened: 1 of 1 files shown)


mschenck:
The output on my machine (which has a prometheus deploy) looks like this:

```
================================================================================
The ngrok Ingress controller has been deployed as a Deployment type to your
cluster.

If you haven't yet, create some Ingress resources in your cluster and they will
be automatically configured on the internet using ngrok.}}

One example, taken from your cluster, is the Service:
   "prometheus-server"

You can make this accessible via Ngrok with the following manifest:
--------------------------------------------------------------------------------
kind: Ingress
metadata:
  name: prometheus-server
  namespace: default
spec:
  ingressClassName: ngrok
  rules:
  - host: prometheus-server-1DcE8sFA.ngrok.io
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: prometheus-server
            port:
              number: 80
--------------------------------------------------------------------------------
Applying this manifest will make the service "prometheus-server"
available on the public internet at "https://prometheus-server-1DcE8sFA.ngrok.io/".

Once done, view your edges in the Dashboard https://dashboard.ngrok.com/cloud-edge/edges
Find the tunnels running in your cluster here https://dashboard.ngrok.com/tunnels/agents

This controller is under active development and is in Alpha. It is not suggested to use this in production.
If you have any questions or feedback, please join us in https://ngrok.com/slack and let us know!
```

jonstacks (approved):

Issue #35, opened 2022-10-05
Migrate to ngrok-go


alex-bezek:
Remove the agent sidecar and replace the tunnel management portion with https://github.com/ngrok/ngrok-go

PR #85, opened 2022-12-22
Replace the ngrok-agent with ngrok-go

jonstacks:
Fixes: #35 

## What

This change removes the ngrok agent sidecar and replaces it with the [ngrok-go](https://github.com/ngrok/ngrok-go) library. Tunnels are now managed in the same process as the tunnel controller.

## How

The TunnelController now includes a TunnelManager which is responsible for keeping track of all tunnels and their creation, updates, and deletion.

## Breaking Changes

No.


helm/ingress-controller/CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.3.0
+### Changed
+
+- Moved from calling ngrok-agent sidecar to using the ngrok-go library in the controller process.
+- Moved `apiKey` and `authtoken` to `credentials.apiKey` and `credentials.authtoken` respectively.
+- `credentialSecrets.name` is now `credentials.secret.name`
+- Changed replicas to 1 by default to work better for default/demo setup.
```
alex-bezek, line 14: Thanks!

pkg/ngrokgodriver/driver.go
```diff
@@ -0,0 +1,121 @@
...
+	"log"
+	"net"
+	"reflect"
+
+	"golang.ngrok.com/ngrok"
+	"golang.ngrok.com/ngrok/config"
+	"golang.org/x/sync/errgroup"
+)
+
+type TunnelManager struct {
```
alex-bezek, line 15: Nit: we should probably doc our exported functions in the pkg folder? If that's something we all think we should do, maybe in a subsequent pr we can setup a linter

pkg/ngrokgodriver/driver.go
```diff
@@ -0,0 +1,121 @@
...
+	session ngrok.Session
+	tunnels map[string]ngrok.Tunnel
+}
+
+func NewTunnelManager() (*TunnelManager, error) {
+	opts := []ngrok.ConnectOption{
+		ngrok.WithAuthtokenFromEnv(),
+	}
```
alex-bezek, line 23: What do the logs look like with how you have it? I'm wondering if we need to set it up to use our logger https://github.com/ngrok/ngrok-go/blob/45a13036a1ff/session.go#L222

pkg/ngrokgodriver/driver.go
```diff
@@ -0,0 +1,121 @@
...
+	tm.tunnels[t.Name] = tun
+	go tm.startTunnel(tun, t.Addr)
+	return nil
+}
+
+func (tm *TunnelManager) DeleteTunnel(ctx context.Context, name string) error {
+	tun := tm.tunnels[name]
+	if tun == nil {
+		return nil
```
alex-bezek, line 56: This makes sense within the context of this function. If the tun is nil, there isn't anything left to do. For the controller calling this though, I wonder if there is a code path where this may happen and be valid. My assumption is that if things are all working as they are supposed to, we shouldn't hit this. And if we are, then maybe we are getting extra reconcile loops or something. Do you think hitting this case deserves a debug/info log so we might happen to see if we are hitting this case, or do you think those wouldn't be helpful and we should look at something more like tracing?

pkg/ngrokgodriver/driver.go
jonstacks: Good catch. I'll add some docs to these

pkg/ngrokgodriver/driver.go
jonstacks: I'd go with logging initially. Its hard to ask people to give traces in issue reports and requires them to have tracing infrastructure set up. If that sounds good, I'll add a log statement to this and give the tunnel manager an actual logger.

pkg/ngrokgodriver/driver.go
jonstacks:
```
1.6716860153547833e+09	DEBUG	events	Normal	{"object": {"kind":"Ingress","namespace":"default","name":"test-app-1","uid":"ef836330-9c76-4739-bd13-5660ef8e65a9","apiVersion":"networking.k8s.io/v1","resourceVersion":"11469545"}, "reason": "CreatedEdge", "message": "Created edge edghts_<redacted>"}
1.6716860154495308e+09	DEBUG	events	Warning	{"object": {"kind":"Ingress","namespace":"default","name":"test-app-2","uid":"0123939d-f9cc-411a-8070-57b6fd83867e","apiVersion":"networking.k8s.io/v1","resourceVersion":"11469294"}, "reason": "EdgeExists", "message": "Edge already exists"}
1.6716860155462549e+09	DEBUG	events	Normal	{"object": {"kind":"Ingress","namespace":"default","name":"test-app-2","uid":"0123939d-f9cc-411a-8070-57b6fd83867e","apiVersion":"networking.k8s.io/v1","resourceVersion":"11469294"}, "reason": "CreatedEdge", "message": "Created edge edghts_<redacted>"}
2022/12/22 22:44:19 Accepted connection from  <redacted>:59426
2022/12/22 22:49:22 Connection closed:  <nil>
2022/12/23 03:52:49 Accepted connection from <redacted>:46789
2022/12/23 03:57:49 Connection closed:  <nil>
```

Yeah, I'll update it to pass the logger in since having 2 different formats in there isn't great

pkg/ngrokgodriver/driver.go
alex-bezek: yep that makes sense on the tracing stuff. thanks!

alex-bezek (approved):
Discussed these offline. Sounds like the logging will be cleaned up a little bit, but otherwise it looks good

pkg/ngrokgodriver/driver.go
jonstacks: Fixed in https://github.com/ngrok/ngrok-ingress-controller/pull/85/commits/d320396e9a92e568b567762b17ea16021a1927b2

pkg/ngrokgodriver/driver.go
jonstacks: Fixed in https://github.com/ngrok/ngrok-ingress-controller/pull/85/commits/d320396e9a92e568b567762b17ea16021a1927b2

PR #91, opened 2022-12-30
Kubebuilder standardization

jonstacks:
## What

Kubebuilder expects there to be a `main.go` file at the root of the repo for api generation. Currently it fails with:

```
2022/12/30 16:39:29 failed to create API: unable to run pre-scaffold tasks of "base.go.kubebuilder.io/v3": main.go file should present in the root directory
```

## How

This change moves the `main.go` file out of the cmd directory and into the root of the project so that we can run `kubebuilder` and have it work as expected. We also pull some of the more recent changes from kubebuilders `Makefile` in and update the go dependencies.

## Breaking Changes

No


Makefile
```diff
@@ -121,6 +118,40 @@ deploy: docker-build manifests ## Deploy controller to the K8s cluster specified
...
+
+## Tool Versions
+KUSTOMIZE_VERSION ?= v3.8.7
+CONTROLLER_TOOLS_VERSION ?= v0.9.2
+
+KUSTOMIZE_INSTALL_SCRIPT ?= "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"
+.PHONY: kustomize
+kustomize: $(KUSTOMIZE) ## Download kustomize locally if necessary.
+$(KUSTOMIZE): $(LOCALBIN)
+	test -s $(LOCALBIN)/kustomize || { curl -Ss $(KUSTOMIZE_INSTALL_SCRIPT) | bash -s -- $(subst v,,$(KUSTOMIZE_VERSION)) $(LOCALBIN); }
```
alex-bezek, line 141: we don't use kustomize still though right? Are we including this just so it doesn't continually show up as a diff even though its not used today?

mschenck (approved):
LGTM

Makefile
jonstacks: Yes. Kubebuilder is using kustomize and I believe it throws an error on generation if its not present. The more we can match kubebuilder on this the better IMO.

PR #94, opened 2023-01-04
Add ReservedDomain CRD

jonstacks:
## What

This adds a new kind, `ReservedDomain`, which is used for managing Reserved Domains in the ngrok API. This will give us an easy way to see all reserved domains and their status with `kubectl get reserveddomains`. In addition, for ingress rules that have the same host across multiple ingresses, we won't need to do the reserved domain logic anymore.

It was the easiest kind to add and will provide a good path forward for implementing the `Tunnel` kind and `Edge` kind as well.

## How

Added a new CRD and controller for `ReservedDomain` and wired it into the ingress controller so the ingress controller can manage `ReservedDomain`s

## Breaking Changes

No


inconshreveable:
can we call this a `Domain`? we've wanted to rename reserveddomain -> domain for a long long time. it's already that way in most places except our public API.

and should it be namespaced to ngrok in some way, e.g. `kubectl get ngrok-domains`?

jonstacks:
@inconshreveable, thank you for pointing that out. Yes, I'll make the update to just `Domain`.

If there is only 1 kind called `Domain`, then you can run
```
kubectl get domains
```
If there are other APIs registered that have a kind of `Domain`, you'll have to do 

```
kubectl get domains.ingress.k8s.ngrok.com
```

api/v1alpha1/reserveddomain_types.go
```diff
@@ -0,0 +1,105 @@
...
+type ReservedDomainSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Domain is the domain name to reserve
+	// +kubebuilder:validation:Required
+	Domain string `json:"domain"`
+
+	// Region is the region in which to reserve the domain
+	// +kubebuilder:validation:Required
```
alex-bezek, line 44: i assume long term we want this to be optional? eventually we want no region to be the default and for it to use the global network. For now though, I think the ingress controller defaults everything to us so required is ok for now?

api/v1alpha1/reserveddomain_types.go
```diff
@@ -0,0 +1,105 @@
...
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// ID is the unique identifier of the reserved domain
+	ID string `json:"id,omitempty"`
+
+	// Domain is the domain that was reserved
+	Domain string `json:"domain,omitempty"`
+
+	// Region is the region in which the reserved domain was created
+	Region string `json:"region,omitempty"`
```
alex-bezek, line 68: why does status need these? is it default to just put everything here? or are these specifically because they are the api responses?

helm/ingress-controller/templates/rbac/reserveddomain_viewer_role.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: clusterrole
+    app.kubernetes.io/instance: reserveddomain-viewer-role
+    app.kubernetes.io/component: rbac
+    app.kubernetes.io/created-by: ngrok-ingress-controller
+    app.kubernetes.io/part-of: ngrok-ingress-controller
+    app.kubernetes.io/managed-by: kustomize
```
alex-bezek, line 11: :(

api/v1alpha1/reserveddomain_types.go
jonstacks: My thought was that this should be explicitly supplied to the CRD. The ingress controller will take care of getting the region config during install and make sure it matches. If we want to open it up to be optional, and defaulting to the ngrok-api-go client behavior, I'm good with that

helm/ingress-controller/templates/rbac/reserveddomain_viewer_role.yaml
jonstacks: Good catch on the auto-generated stuff. I'll fix those up

api/v1alpha1/reserveddomain_types.go
jonstacks: I wanted to get as much info back from the API and on here as possible. The `.Status` seems to be pretty wide open in terms of what we can put on there.

alex-bezek (approved):

PR #95, opened 2023-01-04
Add Tunnel CRD and Controller

jonstacks:
## What

This adds a `Tunnel` Kind and Controller. This will make our creation of tunnels flexible enough that we can build them out with either an `Ingress` Controller or `Gateway` Controller going forward, although we'll only be doing `Ingress` at first.

## How

Modified the existing Tunnel Controller and Reconciler to to watch the `Tunnel` kind instead of also watching ingresses. The ingress controller can now also has an ownerReference on the created tunnels as well. This should allow it to clean up tunnels as well once nothing else is using the tunnel.

## Breaking Changes

Yes. We removed the `k8s.ngrok.com/ingress-name` label on the tunnels and it will no longer be added to the Tunnel Group Backend. This is because tunnel backends map to a tuple of `<namespace,service,service-port>` in k8s since they can be reused by multiple ingresses.


api/v1alpha1/tunnel_types.go
```diff
@@ -0,0 +1,77 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// TunnelSpec defines the desired state of Tunnel
+type TunnelSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// ForwardsTo is the name and port of the service to forward traffic to
+	ForwardsTo string `json:"forwardsTo,omitempty"`
```
alex-bezek, line 40: The forwards to seems like it should be required?

api/v1alpha1/tunnel_types.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// Labels are key/value pairs that are attached to the tunnel
+	Labels map[string]string `json:"labels,omitempty"`
+}
+
+// TunnelStatus defines the observed state of Tunnel
```
alex-bezek, line 46: do you want to put a note here or log an issue with a little info about why we can't have a status here and how we could split it into separate controllers? I'm assuming we are putting that on the back burner but it would be good to not forget about it. I can spin up an issue with my understanding for you to review if you want?

go.mod
```diff
@@ -103,3 +103,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace golang.ngrok.com/ngrok => github.com/jonstacks/ngrok-go v0.0.0-20230104195048-cc46e7fe982f
```
alex-bezek, line 107: image

internal/controllers/ingress_controller.go
```diff
@@ -272,3 +315,44 @@ func (irec *IngressReconciler) ingressToDomains(ctx context.Context, ingress *ne
...
 }
+
+func ingressToTunnels(ingress *netv1.Ingress) []ingressv1alpha1.Tunnel {
+	tunnels := make([]ingressv1alpha1.Tunnel, 0)
+
+	if ingress == nil || len(ingress.Spec.Rules) == 0 {
+		return tunnels
+	}
+
+	// Tunnels should be unique on a service and port basis so if they are ferenced more than once, we
```
alex-bezek, line 326:
```suggestion
	// Tunnels should be unique on a service and port basis so if they are referenced more than once, we
```

internal/controllers/ingress_controller.go
```diff
@@ -272,3 +315,44 @@ func (irec *IngressReconciler) ingressToDomains(ctx context.Context, ingress *ne
...
+			continue
+		}
+
+		for _, path := range rule.HTTP.Paths {
+			serviceName := path.Backend.Service.Name
+			servicePort := path.Backend.Service.Port.Number
+			tunnelAddr := fmt.Sprintf("%s.%s.%s:%d", serviceName, ingress.Namespace, clusterDomain, servicePort)
+			tunnelName := fmt.Sprintf("%s-%d", serviceName, servicePort)
+
+			tunnelMap[tunnelName] = ingressv1alpha1.Tunnel{
```
alex-bezek, line 340: making sure i'm understanding this correctly: we could have key conflicts here within one ingress, but because the value is composed only of values that are a part of the key or are consistent here, those conflicts would always have the same spec value (forwards to and labels) and so its fine?

main.go
```diff
@@ -151,6 +142,15 @@ func runController(ctx context.Context, opts managerOpts) error {
 		setupLog.Error(err, "unable to create controller", "controller", "Domain")
 		os.Exit(1)
 	}
+	if err = (&controllers.TunnelReconciler{
+		Client:   mgr.GetClient(),
+		Log:      ctrl.Log.WithName("controllers").WithName("Tunnel"),
```
alex-bezek, line 147: super nit: technically the ingress controller is lower case with hyphens. Do you think we should make it human readable like this _tunnel_ -> _Tunnel_ ?

internal/controllers/ingress_controller.go
```diff
@@ -248,6 +260,37 @@ func (irec *IngressReconciler) reconcileDomains(ctx context.Context, ingress *ne
 	return nil
 }
 
+func (irec *IngressReconciler) reconcileTunnels(ctx context.Context, ingress *netv1.Ingress) error {
+	tunnels := ingressToTunnels(ingress)
+
+	for _, tunnel := range tunnels {
+		if err := controllerutil.SetControllerReference(ingress, &tunnel, irec.Scheme); err != nil {
```
alex-bezek, line 267: This isn't a change request, just me clarifying how this works and where we are going with this:

reading through https://book.kubebuilder.io/reference/watching-resources/operator-managed.html?highlight=owner#manage-the-owned-resource later on in their _SetupWithManager_ they add a `.Owns` call for the owned resource. 

I'm assuming that once we get the store up and going and we have it handling tunnels, then that is when the ingress controller will 
- get a reconcile for the ingress
- update it in the store
- calculate intended tunnel differences between the stores ingress and the stores tunnels
- do a kubectl update of the tunnel based on that diff
- receive an update reconcile event for the tunnel *because it is watching/owning that resource*
- update store with new tunnel value

Let me know if thats what you are thinking

internal/controllers/ingress_controller_test.go
```diff
@@ -108,3 +108,90 @@ func TestIngressReconcilerIngressToEdge(t *testing.T) {
...
+								HTTP: &netv1.HTTPIngressRuleValue{
+									Paths: []netv1.HTTPIngressPath{
+										{
+											Path: "/",
+											Backend: netv1.IngressBackend{
+												Service: &netv1.IngressServiceBackend{
+													Name: "test-service",
+													Port: netv1.ServiceBackendPort{
+														Number: 8080,
+													},
```
alex-bezek, line 157: It might be worth having the test case include multiple of the same service/port combination and make sure it only creates 1 and doesn't error since thats the one _interesting case_ i see

api/v1alpha1/tunnel_types.go
jonstacks: Yeah, I think we can have the status in here, but I saved it as a future enhancement. If you want to spin up an issue for that, I can review and add any additional context.

internal/controllers/ingress_controller.go
jonstacks: This might have been an early optimization, but assuming you you have an ingress where:

* `/` -> forwards to service-1 on port 80
* `/path1` -> forwards to service-2 on port 80
* `/path2` -> forwards to service-2 on port 80

The map will have:

* service-1-80 -> Tunnel for service-1 on port 80
* service-2-80 -> Tunnel for service-2 on port 80

The thinking here being we don't need to duplicate the tunnel for service-2 on port 80. Since the tunnel backends are based on selectors, we can just reference the same tunnel backend.

Let me know if that makes sense.

internal/controllers/ingress_controller_test.go
jonstacks: Good catch, I'll add one of those tests

internal/controllers/non_leader_elected_controller_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	c, err := controller.NewUnmanaged("tunnel-controller", m, controller.Options{
+		Reconciler: rec,
+	})
+	assert.NoError(t, err)
+
+	wrappedController := NonLeaderElectedController{c}
+
+	// Assert that the wrapped controller implements the LeaderElectionRunnable interface
+	assert.Implements(t, (*manager.LeaderElectionRunnable)(nil), wrappedController)
```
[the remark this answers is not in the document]
jonstacks: So I think there is a bit of a distinction between the Controller and the Reconciler. The Controller has a Reconciler. We don't really care about the Reconciler implementing `LeaderElectionRunnable`. We want the controller that we pass to `mgr.Add` to implement `manager.LeaderElectionRunnable`

internal/controllers/ingress_controller_test.go
jonstacks: Did some cleanup of the tests. Should now be covered by https://github.com/ngrok/ngrok-ingress-controller/pull/95/commits/5d2dca984d4636d3579ee29d345c2adde9053e50

main.go
jonstacks: I think the recorder is lower case with hyphens. The Ingress controller logs seems to just be:

```go
Log:            ctrl.Log.WithName("controllers").WithName("ingress"),
```

I'm fine with making the change to `Tunnel` -> `tunnel` and for `Domain` -> `domain` if you want

internal/controllers/ingress_controller.go
jonstacks: Yep, that is what we are shooting for!

api/v1alpha1/tunnel_types.go
jonstacks: Fixed in https://github.com/ngrok/ngrok-ingress-controller/pull/95/commits/b32ccc27456d4846847bd7042d19e82d18eed890

internal/controllers/ingress_controller.go
alex-bezek: yep, that makes sense. I was just clarifying that the loop at some point will have a map with `service-1-80` as a key and then override the value with a new value. But that value will be the same so it doesn't matter if the key is already present

alex-bezek (approved):

PR #96, opened 2023-01-05
Alex/configurable envs

alex-bezek:
## What
- for clients to use white label agent ingress domains, the tunnel server address needs to be configurable. This is done via a helm value
- add the ability to pass in custom environment variables
- add the ability to pass in custom volumes and volume mounts
- allow ngrok api address to be configurable via an undocumented environment variable

## How
Plumbing through helm values to the inner go code

## Breaking Changes
none


helm/ingress-controller/templates/controller-deployment.yaml
```diff
@@ -45,12 +45,19 @@ spec:
...
+        {{- if .Values.serverAddr }}
+        - --server-addr={{ .Values.serverAddr}}
+        {{- end }}
         - --health-probe-bind-address=:8081
         - --metrics-bind-address=:8080
         - --election-id={{ include "ngrok-ingress-controller.fullname" . }}-leader
         securityContext:
           allowPrivilegeEscalation: false
         env:
+        {{- range $key, $value := .Values.extraEnv }}
```
jonstacks, line 57: nit: Can we move this to the end of the `env` section since its a last value wins?

alex-bezek: we can, i think i did this because it made the test case simpler :) i'll move it to the bottom and see if i can find a way to test it without hardcoding the number of other envs

helm/ingress-controller/templates/controller-deployment.yaml
```diff
@@ -65,6 +72,12 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        {{- if .Values.extraVolumeMounts  }}
+        volumeMounts:
+        {{ toYaml .Values.extraVolumeMounts | nindent 10 }}
+        {{- end }}
+        {{- if .Values.extraVolumes }}
```
jonstacks, line 79: This looks like it might be out of place?  Should it be on line 95 instead?

alex-bezek: oh good catch, i guess i did't see this in the snapshot because we only have 1 and are providing the value. I guess maybe we should have atleast a snapshot with no overrides set and a snapshot with all options exercised. I'll add that and fix this

pkg/tunneldriver/driver.go
```diff
@@ -20,10 +20,12 @@ type TunnelDriver struct {
 }
 
 // New creates and initializes a new TunnelDriver
-func New() (*TunnelDriver, error) {
+func New(serverAddr string) (*TunnelDriver, error) {
 	opts := []ngrok.ConnectOption{
 		ngrok.WithAuthtokenFromEnv(),
 	}
+	opts = append(opts, ngrok.WithServer(serverAddr))
```
jonstacks, line 27: I think we will only want to do this if `serverAddr != ""`, right? And is there a reason not to take this from an environment variable if it exists, like we are doing for the `NGROK_API_ADDR` up above?

pkg/ngrokapidriver/driver.go
```diff
@@ -37,8 +39,18 @@ type ngrokAPIDriver struct {
 	region          string
 }
 
+// TODO:
```
alex-bezek, line 42: fixing

pkg/tunneldriver/driver.go
alex-bezek: I could add that. "" is the nil value of string i think and from looking at the code i see if its set to "" it falls back to the default. Perhaps i should code more defensively incase ngrok-go one day blows up if passed ""?

> is there a reason not to take this from an environment variable if it exists, like we are doing for the NGROK_API_ADDR up above?

an environment variable doesn't exist, I could add it from the helm value, but it felt more proper to pass it as a flag to the controller instead of via an environment variable. The ngrok_api_addr is just an environment variable because its not something we wanted to expose first class via helm and everything

jonstacks (approved):

main.go
```diff
@@ -143,10 +145,11 @@ func runController(ctx context.Context, opts managerOpts) error {
...
 	if err = (&controllers.TunnelReconciler{
-		Client:   mgr.GetClient(),
-		Log:      ctrl.Log.WithName("controllers").WithName("tunnel"),
-		Scheme:   mgr.GetScheme(),
-		Recorder: mgr.GetEventRecorderFor("tunnel-controller"),
+		Client:     mgr.GetClient(),
+		Log:        ctrl.Log.WithName("controllers").WithName("tunnel"),
+		Scheme:     mgr.GetScheme(),
+		Recorder:   mgr.GetEventRecorderFor("tunnel-controller"),
+		ServerAddr: opts.serverAddr,
```
alex-bezek, line 152: I think in a subsequent PR, i may refactor this to construct the driver here and pass it in similar to the ingress controller https://github.com/ngrok/ngrok-ingress-controller/blob/649acbaed3589e75bf560b414f53649d7261f522/main.go#L131
It looks like region isn't being plumbed through, and adding lots of options here won't be fun. I could also just create an opts struct. 

anyways, this is probably short lived

PR #97, opened 2023-01-06
pass region into ngrok go driver

alex-bezek:
## What
Plumb the region through from the helm values to the ngrok-go session

## How
We were already taking a region option from helm and passing it through to the ngrok api driver, this just also sets it up in the ngrok-go driver

## Breaking Changes
No

## Validation
image
tested making an edge in JP. Just need to make sure to change the domain to something.jp.ngrok.io


jonstacks (approved):
This feels much cleaner with wiring up the tunnel driver opts by doing it in `main.go` :+1:

PR #99, opened 2023-01-06
Fix: Domain not reconciling ID even though it exists

jonstacks:
## What

This fixes an issue with the controller for domains not reconciling a domain that already exists.

## How

If there is no `ID` in the domain status, we switch to first trying to find the domain by hostname. If that fails, we then do a create. Also, we removed the replace in go.mod since the changes have been included upstream now

## Breaking Changes

No


internal/controllers/domain_controller.go
```diff
@@ -183,16 +183,22 @@ func (r *DomainReconciler) deleteExternalResources(ctx context.Context, domain *
...
-	req := &ngrok.ReservedDomainCreate{
-		Domain:      domain.Spec.Domain,
-		Region:      domain.Spec.Region,
-		Description: domain.Spec.Description,
-		Metadata:    domain.Spec.Metadata,
-	}
-	resp, err := r.DomainsClient.Create(ctx, req)
+	// First check if the reserved domain already exists. The API is sometimes returning dangling CNAME records
+	// errors right now, so we'll check if the domain already exists before trying to create it.
+	resp, err := r.findReservedDomainByHostname(ctx, domain.Spec.Domain)
```
alex-bezek, line 188: another pattern i think we are using in other functions is to return a not found error type instead of just nil, nil. I'm curious if you have opinions on the matter, I asked in our internal chat and it sounded like it could go either way and there isn't a standard

alex-bezek (approved):

internal/controllers/domain_controller.go
jonstacks: We talked about this and we are already doing it this way in the driver. We'll continue with this for consistency for now, but may come back and implement an errors package with error checking vs returning a nil object

PR #109, opened 2023-01-18
Set ingress class default to false

alex-bezek:
## What
Changes the default helm values for whether the ngrok ingress class should be the default from true to false. We should not assume we would be the default as it could cause problems if someone installs it in a cluster with another ingress controller.

## How
Change the value in the values.yaml file

## Breaking Changes
Technically yes, this is non-passive and reverses the behavior. If its installed in a cluster and its the only ingress class though, it will not pickup things by default anymore.

Our logic for how ingress classes are handled is here https://github.com/ngrok/kubernetes-ingress-controller/blob/main/internal/controllers/main.go#L29-L37 . With these helm defaults, you would install it in the cluster and would need to specify the ingress class on their ingress objects. Alternatively, if we set the default value for `ingressClass.create` to false, then if this is the only controller installed https://github.com/ngrok/kubernetes-ingress-controller/blob/main/internal/controllers/main.go#L30 would make it handle ingress objects by default. 

Our logic for ingress classes may need to change up a bit at some point, but I haven't found any specifications on how it behaves in each scenario.

### DIFF
--- a/scripts/e2e.sh
+++ b/scripts/e2e.sh
@@ -24,6 +24,13 @@ do
   echo "kubectl get ingress -n $i -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -"
   kubectl get ingress -n $i -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -
 done
+# Remove finalizers from domains in namespace
+kubectl get domains -A -o custom-columns=NAMESPACE:metadata.namespace,NAME:metadata.name --no-headers | \
+while read -r i
+do
+  echo "kubectl get domains -n $i -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -"
+  kubectl get domains -n $i -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -
+done
 
 kubectl delete namespace $namespace --ignore-not-found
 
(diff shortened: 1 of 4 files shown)


alex-bezek:
We discussed the logic around the ingress controller. We agreed that we should prioritize these usecases:
- a user can add the ngrok ingress class to their ingress objects to have the controller pick them up
- a user can set default to true to make non-marked classes get picked up.

But we should not try to support a special case where if you install it and its the only one, then it should try to act like a default since that is error prone and somewhat surprising. @jonstacks let me know if that summarizes our conversation

scripts/e2e.sh
```diff
@@ -24,6 +24,13 @@ do
   echo "kubectl get ingress -n $i -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -"
   kubectl get ingress -n $i -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -
 done
+# Remove finalizers from domains in namespace
+kubectl get domains -A -o custom-columns=NAMESPACE:metadata.namespace,NAME:metadata.name --no-headers | \
+while read -r i
+do
+  echo "kubectl get domains -n $i -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -"
+  kubectl get domains -n $i -o=json | jq '.metadata.finalizers = null' | kubectl apply -f -
+done
```
alex-bezek, line 33: the domains hang the namespace deletion since they have finalizers

jonstacks (approved):

PR #122, opened 2023-01-26
Webhook verification route module support

jonstacks:
## What

Adds support for [Webhook Verification](https://ngrok.com/docs/cloud-edge/modules/webhook) Route Module.

## How

1. Parse annotations from ingress into Webhook Verification config
2. Pass config to HTTPSEdge controller
3. Adds support in the HTTPSEdge controller to disable webhook verification or replace the contents with new module config.

## Breaking Changes

No


internal/controllers/httpsedge_controller.go
```diff
@@ -315,6 +319,40 @@ func (r *HTTPSEdgeReconciler) setEdgeRouteResponseHeaders(ctx context.Context, e
 	return err
 }
 
+func (r *HTTPSEdgeReconciler) setEdgeRouteWebhookVerification(ctx context.Context, edge *ingressv1alpha1.HTTPSEdge, routeID string, webhookValidation *ingressv1alpha1.EndpointWebhookValidation) error {
+	client := r.NgrokClientset.EdgeModules().HTTPS().Routes().WebhookVerification()
+	if webhookValidation == nil {
+		return client.Delete(ctx, &ngrok.EdgeRouteItem{EdgeID: edge.Status.ID, ID: routeID})
+	}
+
+	// TODO: Get the secret from the store
```
alex-bezek, line 328: Discussed offline, but the edge controller won't be accessing the store, but this Client get is fine since it should rely on the controller-runtime manager's cache https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cluster/cluster.go#L49-L53

internal/controllers/httpsedge_controller.go
```diff
@@ -315,6 +319,40 @@ func (r *HTTPSEdgeReconciler) setEdgeRouteResponseHeaders(ctx context.Context, e
...
+	secret := &v1.Secret{}
+	err := r.Client.Get(ctx, types.NamespacedName{
+		Name:      webhookValidation.SecretRef.Name,
+		Namespace: edge.Namespace,
+	}, secret)
+	if err != nil {
+		return err
```
alex-bezek, line 335: so does this make the event the user see's the plain k8s get failure of "resource not found". At some point I think we'll want to review the wording and error messages for common configuration error cases I think

internal/annotations/webhook_verification/webhook_verification.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	switch provider {
+	case "sns":
+		return &EndpointWebhookValidation{Provider: provider}, nil
+	}
+
+	secretName, err := parser.GetStringAnnotation("webhook-validation-secret-name", ing)
```
alex-bezek, line 29: should we check for if the error is `ErrMissingAnnotations` and if so throw a validation error? Or have the parser do some validation in general? This is the first of the modules I think where if you supply 1 annotation, others are required. I think in this case if you add the provider but not the secret, then we'll throw a missing annotation here which then https://github.com/ngrok/kubernetes-ingress-controller/blob/1061b7961e0d3487a0e381d7a478bacc76943813/internal/annotations/annotations.go#L64 here we'd skip over. I'm wondering if we should have a more specific "you provided X annotation but Y, and Z are also required" type of message for the user

docs/annotations.md
```diff
@@ -89,3 +90,26 @@ metadata:
...
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: minimal-ingress
+  annotations:
+    k8s.ngrok.com/webhook-validation-provider: "github"
+    k8s.ngrok.com/webhook-validation-secret-name: "github-webhook-token"
+    k8s.ngrok.com/webhook-validation-secret-key: "SECRET_TOKEN"
```
alex-bezek, line 111: I think these should be called `verification` not `validation`. I see in the ngrok api go code and the docs page the type is an `EndpointWebhookValidation` but the actual key of the body is verification. I think before edges, when we used endpoint configurations, this was the name, but got changed to verification in the edge world 
https://github.com/ngrok/ngrok-api-go/blob/da8b5e25f3a1272c3106bfb06b05fd3847fa201c/datatypes.go#L2426

We may even go as far as naming these the same way the ngrok cli is named such as `https-edge-route-webhook-verification`

api/v1alpha1/ngrok_common.go
```diff
@@ -49,3 +49,21 @@ type EndpointHeaders struct {
...
+	Name string `json:"name,omitempty"`
+	// Key in the secret to use
+	Key string `json:"key,omitempty"`
+}
+
+type EndpointWebhookValidation struct {
+	// a string indicating which webhook provider will be sending webhooks to this
+	// endpoint. Value must be one of the supported providers defined at
+	// https://ngrok.com/docs/cloud-edge#webhook-verification
+	// (https://ngrok.com/docs/cloud-edge#webhook-verification)
```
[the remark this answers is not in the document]
jonstacks: Probably? I copied it from https://github.com/ngrok/ngrok-api-go/blob/da8b5e25f3a1272c3106bfb06b05fd3847fa201c/datatypes.go#L1456-L1459. We might want to go down the path of generating these sibling structs at some point.

internal/annotations/webhook_verification/webhook_verification.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (wv webhookVerification) Parse(ing *networking.Ingress) (interface{}, error) {
+	provider, err := parser.GetStringAnnotation("webhook-validation-provider", ing)
+	if err != nil {
+		return nil, err
+	}
+
+	switch provider {
+	case "sns":
```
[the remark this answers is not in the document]
jonstacks: The SNS provider is the only one that doesn't require a secret AFAICT. I was trying to push this validation into the annotation parsing to return an error, but we can defer it to the API. We'll have to handle a secret not being found and then just passing an empty string and figuring out how to bubble up that error to the end user if they forget to specify the webhook secret name or secret key for locating the value in a Kubernetes secret.

internal/controllers/httpsedge_controller.go
jonstacks: Yes, that should be the error returned. As an aside, I was also wondering if its worth it to upgrade our logging at some point to be able to do debug logging. I think we are getting close to the point where having info log for everything is too noisy, but for development or troubleshooting we'd want those debug logs.

internal/controllers/httpsedge_controller.go
jonstacks: Ahhh....yes. that is right.

internal/annotations/webhook_verification/webhook_verification.go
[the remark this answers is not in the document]
alex-bezek: I scrolled down the `S` section and didn't see sns... `Amazon SNS` :D i see here now how its short circuiting the secret lookup. Using the api for as much validation as possible makes sense

api/v1alpha1/ngrok_common.go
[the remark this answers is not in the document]
jonstacks: Removed the duplicate link in [05b43b1](https://github.com/ngrok/kubernetes-ingress-controller/pull/122/commits/05b43b1ceaaf70f88b84d08d2c1308ba2b04b42a)

internal/controllers/httpsedge_controller.go
jonstacks: Removed in [80ae704](https://github.com/ngrok/kubernetes-ingress-controller/pull/122/commits/80ae7046f0733b6afd98ade63803a40d57ac6345)

internal/annotations/webhook_verification/webhook_verification.go
jonstacks: I'm thinking we'd probably want to handle this with a [Validating Admission Webhook](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#validatingadmissionwebhook) in order to actually give a better error message back to the user in real time as opposed to making them chase it down in logs? Does that seem like the right place for it to you?

internal/controllers/httpsedge_controller.go
alex-bezek: yeah, I think that will be a good idea. I think once things stabalize we can review all the events/logs that happen in some flows and tune things from there

internal/annotations/webhook_verification/webhook_verification.go
alex-bezek: hmm, that would probably be a much better user experience. If we had that then we'd be able to give them that validation at apply time instead of at reconcile time right? Digging through logs wouldn't be fun, but we could show pretty clear/obvious k8s event warn/errors in the interim. I've heard admission webhooks are powerful but can be dangerous foot guns and feels like something we might want to look into more in the future?

alex-bezek (approved):
Discussed, the we are making the validation/verification naming consistent and then 👍

docs/annotations.md
jonstacks: Fixed in [43b3cf3](https://github.com/ngrok/kubernetes-ingress-controller/pull/122/commits/43b3cf3d377e5ed2b9983e1fa3072c738d467f8d)

PR #125, opened 2023-01-27
Allow setting minimum TLS Version

jonstacks:
## What

Allows users to set the minimum TLS Version for TLS termination at Edge. If not supplied, the module will be disabled and TLS will still be terminated at the edge using the ngrok provided default.

## How

Moves the annotation parsing outside of planning the routes to the edge reconciliation and then passes in the parsed annotations to the route planning since TLS termination is configured at the edge and not on the individual routes.

## Breaking Changes

No


api/v1alpha1/httpsedge_types.go
```diff
@@ -68,6 +68,9 @@ type HTTPSEdgeSpec struct {
 
 	// Routes is a list of routes served by this edge
 	Routes []HTTPSEdgeRouteSpec `json:"routes,omitempty"`
+
+	// TLSTermination is the TLS termination configuration for this edge
+	TLSTermination *EndpointTLSTerminationAtEdge `json:"tlsTermination,omitempty"`
```
alex-bezek, line 73: I know on the others modules we removed the "enabled" attribute in favor of just enabling it if there is a value present. This one though we could keep the enabled as without it the functionality changes a bit. If we had both enabled and min_version, then based on the API, the min_version could be optional and would default`if unspecified, ngrok will choose an industry-safe default`.

relying on that though does make the api a bit more confusing in that if you set the min_version it won't do anything (or would error) unless you also set enabled

api/v1alpha1/httpsedge_types.go
```diff
@@ -68,6 +68,9 @@ type HTTPSEdgeSpec struct {
 
 	// Routes is a list of routes served by this edge
 	Routes []HTTPSEdgeRouteSpec `json:"routes,omitempty"`
+
```
alex-bezek, line 71: I also see this terminateAt in the api client https://github.com/ngrok/ngrok-api-go/blob/da8b5e25f3a1272c3106bfb06b05fd3847fa201c/datatypes.go#L1558 and its in the api response body https://ngrok.com/docs/api/resources/https-edge-tls-termination-module#fields but not the update body. Any idea what's going on there?

internal/controllers/httpsedge_controller.go
```diff
@@ -315,6 +319,21 @@ func (r *HTTPSEdgeReconciler) setEdgeRouteResponseHeaders(ctx context.Context, e
 	return err
 }
 
+func (r *HTTPSEdgeReconciler) setEdgeTLSTermination(ctx context.Context, edgeID string, tlsTermination *ingressv1alpha1.EndpointTLSTerminationAtEdge) error {
+	client := r.NgrokClientset.EdgeModules().HTTPS().TLSTermination()
```
alex-bezek, line 323: Interestingly, it looks like you can also just set this on the edge itself instead of as an edge module. any idea if they are functionally equivalent? If so how you have it makes more sense than trying to keep it on the edge itself to me fwiw

docs/annotations.md
```diff
@@ -89,3 +91,21 @@ metadata:
...
+The `k8s.ngrok.com/tls-min-version` annotation can be used to set the minimum TLS version for all routes belonging to an Edge. *Note*: **This setting applies to all routes
+belonging to an Edge, not just the routes defined in the Ingress resource**. The annotation accepts a string value. If not specified(default), the `TLS Termination` module will be disabled in ngrok and TLS will be terminated at the edge using the ngrok default minimum TLS version.
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: minimal-ingress
+  annotations:
+    k8s.ngrok.com/tls-min-version: "1.3"
```
alex-bezek, line 108: for the names of these annotations, I assume we want to work with russ in our next product review to make sure these are consistent across products. Thoughts on leveraging what we use in the ngrok cli for these? 

```
ngrok api edge-modules
NAME:
  edge-modules -

USAGE:
  ngrok api edge-modules [flags]

DESCRIPTION:

COMMANDS:
  https-edge-mutual-tls
  https-edge-route-backend
  https-edge-route-circuit-breaker
  https-edge-route-compression
  https-edge-route-ip-restriction
  https-edge-route-oauth
  https-edge-route-oidc
  https-edge-route-request-headers
  https-edge-route-response-headers
  https-edge-route-saml
  https-edge-route-webhook-verification
  https-edge-route-websocket-tcp-converter
  https-edge-tls-termination
  tcp-edge-backend
  tcp-edge-ip-restriction
  tls-edge-backend
  tls-edge-ip-restriction
  tls-edge-mutual-tls
  tls-edge-tls-termination
```

so this would be `k8s.ngrok.com/https-edge-tls-termination-min-version`

api/v1alpha1/httpsedge_types.go
jonstacks: I believe it is just a common config type that is returned for TLSEdges and HTTPSEdges. In the case of a HTTPSEdge, it will be terminated at edge. In the case of TLSEdge, it will be terminated at the upstream. You are correct in that the update body does not allow for you setting `terminateAt`. I noticed that as well while working on this.

internal/controllers/httpsedge_controller.go
jonstacks: Yep. You can do that with all of the edge modules from what I can tell. However, from our API docs, there is no way to disable modules through the update EdgeRoute API. By updating the edge & edge route first and then handling the modules, we make it so we only have to have this logic in 1 place which seems cleaner to me and makes it so we only have to duplicate the edge logic for create/update

api/v1alpha1/httpsedge_types.go
alex-bezek: ahh i didn't see it was shared. that makes sense that the behaviour would be different between the 2 of those and the user isn't in control

docs/annotations.md
jonstacks: I'm not opposed, but would like Russ' feedback on it. Given the different section, I'd almost be more inclined to have `.` delimiting some of these: `https-edge.tls-termination.min-version` or something like that. They should be very easy to change and map though unless we change to/from embedded json on a lot of them.

internal/controllers/httpsedge_controller.go
alex-bezek: ah nice, yeah that makes sense

api/v1alpha1/httpsedge_types.go
jonstacks: From the [docs](https://ngrok.com/docs/cloud-edge/modules/tls-termination) page:

> This module is used to configure how the TLS connection is terminated at the edge. For HTTPS Edges, TLS is always terminated at the ngrok Edge which means you can only configure the TLS version enforced there. For TLS tunnels, you can configure where the TLS termination occurs.

> The default TLS version for ngrok is 1.2. You can configure the minimum TLS Version via the TLS module. The minimum TLS version ngrok supports is 1.0 up to the latest version 1.3.

it looks like for HTTPSEdges, TLS is always terminated at the ngrok Edge. Exposing `Enabled` seems extraneous to me in a configuration as code(kubernetes manifests). If values are supplied in the manifests, we'll want it enabled. If they are not supplied, the module is disabled/removed. Open to adding this back in though if there is a compelling reason.

docs/annotations.md
alex-bezek: I do like the idea of having a delimiter between the module type vs the actual configuration item. I haven't seen it used much but its not dissallowed. Accross a couple of controllers i see a few cases where people used `.` to delimit a few random words like `alb.ingress.kubernetes.io/group.name` and then some use it to reprsent a variable object like a key in a dictionary `konghq.com/headers.*`

:+1: to getting feedback from russ and they are easy to change so we can do them later

api/v1alpha1/httpsedge_types.go
alex-bezek: Ahh, I misunderstood the docs. I didn't look at the others to see enabled was on them all (i thought it was something we had done, not from the api). The fact that its always on makes sense

alex-bezek (approved):

Issue #152, opened 2023-02-15
Allow ingress class controller name to be configureable


alex-bezek:
### Description

While we offer helm values to override the ingressClass name, we don't let you change the `controllerName` which is matched against the ingress classes spec.controller name. So even though you can create multiple ingress classes, you can't deploy separate controllers to match each one specifically. 

### Use Case

Install multiple ngrok kubernetes ingress controllers watching different ingress classes

PR #159, opened 2023-02-22
Make controller name configurable so multiple can be installed in a c…

alex-bezek:
…luster at a time

closes: #152

## What
While we currently offer a way for the user to change the ingress class name, they can't change the ingress class spec controller name that the controller uses to match on. This is required if you want to install multiple ngrok ingress controllers that watch specific/different ingress classes. 

## How
This simply plumbs through a helm value that is passed to the controller and used for the ingress class filtering logic

## Breaking Changes
No, the default is the same


jonstacks:
I think the helm snapshots need to be updated for this one, but then LGTM

alex-bezek:
updated!

jonstacks (approved):

PR #164, opened 2023-02-24
Fix flakey test and add another

alex-bezek:
## What
- I wrote a flakey test that is in main and needs to be fixed. 

## How
- The test checks the status of the ingress object returned for 2 status objects at 0 and 1
- Golang (like many languages) randomizes the access to map keys, so the end result that's put in the slice isn't ordered
- This causes it to fail/pass randomly. Passing the test seed in doesn't reproduce it because it's not based on the rand seed, it's just how maps work in go. 
- fixed by ensuring the length is correct, making sure there are no duplicates, and then ensuring each shows up

## Breaking Changes
No, its just tests

### DIFF
--- a/internal/store/driver_test.go
+++ b/internal/store/driver_test.go
@@ -234,7 +234,7 @@ var _ = Describe("Driver", func() {
 			Expect(len(status)).To(Equal(0))
 		})
 
-		It("Should return multiple statuses for multiple matching domains", func() {
+		It("Should return multiple statuses for multiple different domains", func() {
 			cname1 := "cnametarget1.com"
 			cname2 := "cnametarget2.com"
 			i1 := NewTestIngressV1("test-ingress", "test-namespace")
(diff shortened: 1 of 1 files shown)


internal/store/driver_test.go
```diff
@@ -278,8 +278,56 @@ var _ = Describe("Driver", func() {
 
 			status := driver.calculateIngressLoadBalancerIPStatus(&i1, c)
 			Expect(len(status)).To(Equal(2))
-			Expect(status[0].Hostname).To(Equal(cname1))
-			Expect(status[1].Hostname).To(Equal(cname2))
+			Expect(status[0].Hostname).To(Not(Equal(status[1].Hostname)))
```
jonstacks, line 281: How do you feel about:

```go
hostnames := make([]string, len(status))
for i, s := range status {
        hostnames[i] = s.Hostname
}
Expect(hostnames).To(Equal([]string{cname1, cname2}))

```

alex-bezek: I read through https://github.com/onsi/gomega/blob/master/matchers.go and found a way to avoid the loop, let me know what you think

jonstacks (approved):

Issue #132, opened 2023-02-09
Add Support for Circuit Breaker HTTPS Edge Module


jonstacks:
### Description

Add support for [Circuit Breaker](https://ngrok.com/docs/cloud-edge/modules/circuit-breaker) Edge module. 

### Use Case

Allows users to set up circuit breakers for services exposed through the ingress controller.

PR #168, opened 2023-03-06
Refactor route module updates to be lazy

### LINKED ISSUE #150: ngrok api rate limits
### What happened

With the addition of the store/driver, its not uncommon to see ngrok rate limit errors saying we reached our 120 calls per minute. While I can look in datadog to see the api calls on our end, the controller doesn't offer the observability to determine why so many calls are being made.

We should add logging and/or metrics about the number of api calls being made so we can make sure that aren't making unnecessary calls. 
- we could log each api call at a debug level log
- create a metric for the number of api calls being made

### How to reproduce

apply a couple of ingresses and if there are enough resources, it hits rate limits

jonstacks:
Fixes #150 

## What

Right now we always either delete the route modules or update them regardless of if there are changes. This results in making unnecessary API calls. Since we get the route back after issuing a create/update, we can just check to see if there are any updates that need to happen before we do the update

## How

I refactored this so there is an `edgeRouteModuleUpdater` that we can call for each route. While moving the existing methods over to the `edgeRouteModuleUpdate`, I added checks so that we make API requests if we need to. This helps to clean things up ahead of working on #132.

## Breaking Changes
No

### DIFF
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -53,14 +53,11 @@ type HTTPSEdgeReconciler struct {
 	Scheme   *runtime.Scheme
 	Recorder record.EventRecorder
 
-	ipPolicyResolver
-
 	NgrokClientset ngrokapi.Clientset
 }
 
 // SetupWithManager sets up the controller with the Manager.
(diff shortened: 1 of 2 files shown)


alex-bezek (approved):
nice, looks like a fairly simple pattern we can follow

PR #171, opened 2023-03-08
Add Circuit Breaker Module support

### LINKED ISSUE #132: Add Support for Circuit Breaker HTTPS Edge Module

jonstacks:
Resolves #132 

## What

Adds support for adding Circuit Breaker configuration to ingresses

## How

Adds the Circuit Breaker configuration to the `NgrokModuleSet` CRD and the `HTTPSEdgeRouteSpec`. Wires it up in the driver to get the computed config for `CircuitBreaker` and add it to applicable routes. Finally, adds code for configuring the circuit breaker if it has changed.

## Breaking Changes
No


api/v1alpha1/ngrok_common.go
```diff
@@ -71,3 +73,25 @@ type EndpointWebhookVerification struct {
 	// requests from the given provider. All providers except AWS SNS require a secret
 	SecretRef *SecretKeyRef `json:"secret,omitempty"`
 }
+
+type EndpointCircuitBreaker struct {
```
alex-bezek, line 77: curiuos if these module configurations should be in the moduleset file now that there is a crd they relate to instead of the common file?

alex-bezek (approved):

api/v1alpha1/ngrok_common.go
alex-bezek: its not important, i was just wondering if is a remnant of before there was a specific CRD, but since its used in multiple, that makes more sense

PR #172, opened 2023-03-09
Use Chart appVersion if tag is not specified

jonstacks:
## What

If you follow the helm install instructions, it will install the most recent helm release `0.6.0` which defaults the docker image tag to `latest`. This causes issues since the helm chart does not have the CRDs that we recently added. For deterministic installs, we should default to using the `appVersion` in the `Chart.yaml` if  `image.tag` is not supplied.

## How

Change default `image.tag` from `latest` to `.Chart.AppVersion`
 
## Breaking Changes
No

### DIFF
--- a/helm/ingress-controller/CHANGELOG.md
+++ b/helm/ingress-controller/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.6.1
+### Fixed
+- Default the image tag to the chart's `appVersion` for predictable installs. Previously, the helm chart would default to the `latest` image tag which can have breaking changes, notably with CRDs.
+
 ## 0.6.0
 ### Changed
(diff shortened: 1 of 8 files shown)


helm/ingress-controller/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.6.1
```
alex-bezek, line 8: This release is going to pull in the unreleased crd changes as well right? 
https://github.com/ngrok/kubernetes-ingress-controller/compare/helm-chart-0.6.0...main

jonstacks: That is a good point. We might have to manually cut a release for this one unless we are okay with these CRDs going out as is

alex-bezek: i think at this stage, it would be ok if some of the crds went out as is for now and we release the rest in the next version. It should be soon and its all alpha right now anyways

alex-bezek (approved):

PR #190, opened 2023-03-17
Add OAuth Route Module

jonstacks:
Resolves #134 

## What

Adds support for `OAuth` Edge Route Module.

## How

Allows a user to create a `NgrokModuleSet` that either uses a ngrok managed OAuth Application, Ex:

```yaml
---
kind: NgrokModuleSet
apiVersion: ingress.k8s.ngrok.com/v1alpha1
metadata:
  name: managed-google-oauth
modules:
  oauth:
    provider:
      google:
        emailDomains:
        - "mydomain.com"
```

Or your own OAuth Application, Ex:

```yaml
---
kind: NgrokModuleSet
apiVersion: ingress.k8s.ngrok.com/v1alpha1
metadata:
  name: custom-google-oauth
modules:
  oauth:
    optionsPassthrough: true
    inactivityTimeout: 3h
    maximumDuration: 24h
    authCheckInterval: 20m

    provider:
      google:
        clientId: "client-id.apps.googleusercontent.com"
        clientSecret:
          name: google-oauth-secret
          key: CLIENT_SECRET
        scopes:
        - openid
        - email
```

Which can then be added to the list of module sets applied to an ingress like so:

```yaml
kind: Ingress
apiVersion: networking.k8s.io/v1
metadata:
  name: test-app-1
  namespace: default
  annotations:
    k8s.ngrok.com/modules: "compression,managed-google-oauth"
```

## Breaking Changes

No.

PR #192, opened 2023-03-18
Add OAuth Route Module

### LINKED ISSUE #134: Add Support for OAuth Edge Module
### Description

Add support for [OAuth](https://ngrok.com/docs/cloud-edge/modules/oauth/) Edge module.

### Use Case

Allow users to easily add OAuth in front of services exposed via a ngrok Ingress.

### Related issues

_No response_

jonstacks:
Resolves #134
Supersedes: #190 - Changing the CRD slightly so provider configuration is not nested.

## What

Adds support for the [OAuth Edge Route Module](https://ngrok.com/docs/cloud-edge/modules/oauth/).

## How

Allows a user to create a `NgrokModuleSet` that either uses a ngrok managed OAuth Application, Ex:

```yaml
---
kind: NgrokModuleSet
metadata:
  name: managed-google-oauth
modules:
  oauth:
    google:
      optionsPassthrough: true
      emailDomains:
      - "mydomain.com"
```

Or by using your own OAuth Application, Ex:

```yaml
---
kind: NgrokModuleSet
metadata:
  name: custom-google-oauth
modules:
  oauth:
    google:
      optionsPassthrough: true
      inactivityTimeout: 3h
      maximumDuration: 24h
      authCheckInterval: 20m
      clientId: "client-id.apps.googleusercontent.com"
      clientSecret:
        name: google-oauth-secret
        key: CLIENT_SECRET
      scopes:
      - openid
      - email
```

Either of these can then be added to the list of module sets applied to an ingress like so:

```
kind: Ingress
metadata:
  name: test-app-1
  namespace: default
  annotations:
    k8s.ngrok.com/modules: "compression,managed-google-oauth"
```

## Breaking Changes
No


api/v1alpha1/ngrok_common.go
```diff
@@ -162,3 +163,252 @@ type EndpointSAML struct {
...
+}
+
+type EndpointOAuthFacebook struct {
+	OAuthProviderCommon `json:",inline"`
+}
+
+func (facebook *EndpointOAuthFacebook) ToNgrok(clientSecret *string) *ngrok.EndpointOAuth {
+	if facebook == nil {
+		return nil
+	}
```
alex-bezek, line 283: are these checks necessary? I guess i'm just not used to adding them, but is it just a good safety check?

internal/controllers/httpsedge_controller.go
```diff
@@ -643,6 +645,71 @@ func (u *edgeRouteModuleUpdater) setEdgeRouteResponseHeaders(ctx context.Context
 	return err
 }
 
+func (u *edgeRouteModuleUpdater) setEdgeRouteOAuth(ctx context.Context, route *ngrok.HTTPSEdgeRoute, routeSpec *ingressv1alpha1.HTTPSEdgeRouteSpec) error {
+	oauth := routeSpec.OAuth
+	oauthClient := u.clientset.OAuth()
+
+	if oauth == nil {
+		if route.OAuth == nil {
+			u.log.Info("OAuth matches desired state, skipping update")
```
alex-bezek, line 654: these logs might cause more noise than good until we have flexible debug/trace logging. Its probably going to be very common for many modules to be nil. This is how all of these cookie-cutter methods work though, so we can log an issue and follow up in a second pr if we want to make that change

internal/controllers/httpsedge_controller.go
```diff
@@ -643,6 +645,71 @@ func (u *edgeRouteModuleUpdater) setEdgeRouteResponseHeaders(ctx context.Context
...
+		}
+
+		var secret *string
+		secretKeyRef := p.ClientSecretKeyRef()
+
+		// Look up the client secret key if its specified,
+		// otherwise default to nil
+		if secretKeyRef != nil {
+			secret, err = u.getSecret(ctx, *secretKeyRef)
+			if client.IgnoreNotFound(err) != nil {
```
alex-bezek, line 687: do you we need to handle the not found case here as well?

alex-bezek (approved):
Nice, the interface makes things even cleaner!

api/v1alpha1/ngrok_common.go
jonstacks: Since every type of provider is optional and can be `nil`, when we try to call `ToNgrok(*string)` if they are nil, we'll get a panic, so I think this is a good defensive check. We also check to see if its nil in the edge controller and if it is, we skip the provider.

internal/controllers/httpsedge_controller.go
jonstacks: You are right. I missed this in the refactor. We should probably get rid of the `client.IgnoreNotFound` around the `err`. If you specify a secret and its not found, we should return it rather than trying to keep going

internal/controllers/httpsedge_controller.go
jonstacks: I think it makes sense to clean them all up in one go.

internal/controllers/httpsedge_controller.go
jonstacks: Fixed in 9ebd898

PR #193, opened 2023-03-21
Add route module documentation

jonstacks:
## What

Adds documentation for `NgrokModuleSet`s along with examples of how to use them. I think we'll want to add more examples to this over time, especially for OAuth.

## Breaking Changes
No

### DIFF
--- a/docs/user-guide/route-modules.md
+++ b/docs/user-guide/route-modules.md
@@ -107,7 +107,7 @@ metadata:
     k8s.ngrok.com/modules: module-set-1,module-set-2
 ```
 
-In this example, the result is the `compression` module is enabled since `module-set-2` was supplied last.
+In this example, the result is the `compression` module is enabled since `module-set-2` was supplied last. If however, the annotation is `k8s.ngrok.com/modules: module-set-2,module-set-1 the order will result in the `compression` module being disabled since `module-set-1` overrides it from `module-set-1`
 
 ## Supported Modules
 
--- a/docs/user-guide/route-modules.md
+++ b/docs/user-guide/route-modules.md
@@ -1,3 +1,369 @@
-# Route Modules
+# Modules
 
-<img src="../assets/images/Under-Construction-Sign.png" alt="Under Construction" width="350" />
+
+ngrok's Cloud Edge [Modules](https://ngrok.com/docs/cloud-edge/modules/) allow you to configure features like compression, IP Restrictions, OAuth, adding/removing headers, and more.
+
+<!-- TOC depthfrom:2 -->
+
+- [Design](#design)
+    - [Reusable](#reusable)
+    - [Composable](#composable)
+- [Supported Modules](#supported-modules)
+    - [Circuit Breaker](#circuit-breaker)
+    - [Compression](#compression)
+        - [Enabled](#enabled)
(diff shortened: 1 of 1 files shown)


docs/user-guide/route-modules.md
```diff
@@ -1,3 +1,369 @@
...
+---
+kind: Ingress
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: example-ingress
+  annotations:
+    k8s.ngrok.com/modules: module-set-1,module-set-2
+```
+
+In this example, the result is the `compression` module is enabled since `module-set-2` was supplied last.
```
alex-bezek, line 110:
```suggestion
In this example, the result is the `compression` module is enabled since `module-set-2` was supplied last. If however, the annotation is `k8s.ngrok.com/modules: module-set-2,module-set-1 the order will result in the `compression` module being disabled since `module-set-1` overrides it from `module-set-1`
```

Nit: Maybe a bit wordy but just trying to be super clear

docs/user-guide/route-modules.md
```diff
@@ -1,3 +1,369 @@
...
+  annotations:
+    k8s.ngrok.com/modules: module-set-1
+```
+
+In this example, the `compression`, `tlsTermination`, and `headers` modules are applied to both Ingresses and the same configuration is used for both. If you change the configuration of the `NgrokModuleSet`, the change will be applied to all Ingresses that use it.
+
+
+### Composable
+
+`NgrokModuleSet`s are designed to be composable. If multiple `NgrokModuleSet`s are applied to an Ingress and a module is configured in more than one, the last one wins. Ex:
```
alex-bezek, line 80: The other neat part is technically you can use rbac to let someone manage these resources for important things like oauth/tls/etc and prevent others from modifying it while letting them still self serve ingress. Do you think thats worth noting in this doc or maybe we should have an general rbac doc eventually like https://kubernetes.github.io/ingress-nginx/deploy/rbac/ ?

alex-bezek (approved):
Nice, i think that covers the individual modules pretty well. 2 general comments we can deal with here or follow up somewhere else:
- we don't really document the types or the values they can use. We link to pages like https://ngrok.com/docs/cloud-edge/modules/oauth/ but i wonder if we should point people to the underlying values the api takes via https://ngrok.com/docs/api/resources/https-edge-route-oauth-module/ . How we deal with docs and validation of these in general is something we haven't solved I don't think
- a "kitchen sink" type of example would be nice. https://ngrok.com/docs/ngrok-agent/config/#full-example has always been super useful for me to copy and strip things out I don't want.

Neither of these are blockers though, this explains the crd and how to use it with ingress just fine

docs/user-guide/route-modules.md
jonstacks: Good call out. I'll update it

docs/user-guide/route-modules.md
jonstacks: Good call. I'll add a separate call out for that

jonstacks:
1. We could maybe generate some pretty documentation from the CRD definition in the helm chart since that defines all the specific types.
2. Another good idea. I'll add a kitchen sink example.

alex-bezek (approved):

PR #198, opened 2023-03-25
Include ngrok controller version in user agent

### LINKED ISSUE #188: Include ngrok controller version in user agent
### Description

We currently set the user agent for ngrok-api calls to `ngrok-ingress-controller/v1-alpha` so we can see edges that are being managed by this controller specifically. We should include the version of the controller that is running. 

### Use Case

Especially in this early development phase where old versions have some bugs we have fixed, its useful to know why a particular user is spamming the api in loops. 

### Related issues

_No response_

jonstacks:
Resolves #188 

## What

Includes the ingress controller version in the `useragent` string. Also prints out the build info when the controller starts.

## How

Pass build info variables to the linker when building the go binary.  Had to change the `.dockerignore` in order to get the `.git` directory, `VERSION`, `Makefile`, and a few others into the docker context so they could be added in the image during the build.

## Breaking Changes

No

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,10 +1,19 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
 # Ignore all files then explicitly whitelist ones we want
 *
+# Copy in the Makefile to re-use build command
+!Makefile
+# Used to inject version build info
+!VERSION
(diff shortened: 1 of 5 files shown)


alex-bezek (approved):
Awesome!

PR #201, opened 2023-03-29
Clean up make deploy command

alex-bezek:
## What
Make deploy has 2 problems right now:
- I left the debug flag on in a commit and it dumps a wall of yaml that's annoying as the default
- with the latest image tag changes, it gets an image pull backoff error

## How
- remove the debug flag
- set tag to latest since thats what the make file builds


jonstacks (approved):
**Non-Blocking**: I vote to keep debug since it dumps all the rendered manifests and think it would be useful, but fine if you want to remove it if you think its too noisy.

PR #203, opened 2023-04-03
Don't leak tcp connections for every tunnel dial

euank:
Before, this leaked one connection per tunnel dial, which made it very easy to run out of connections. On typical linux machines, it would mean you get whatever the ephemeral port range is of working proxy connections, which isn't a huge number. Under 65k.

Anyway, this fixes it in the naive way by closing the connections after their respective copy is complete.

This also adds a test that would have failed before, though it required some intrusive mocking changes meaning I can't actually run it on the old code.

I also chose to remove the errgroup.WithContext since, well, the ctx wasn't respected in `io.Copy` anyway, so it didn't do anything other than give a sorta false sense of security.

Oh, and while I was at mocking, I switched from `net.Dial` to `net.DialContext`. It's identical since ctx is a context.Background there, but maybe it'll nudge us to fix that too at some point.


jonstacks (approved):
Thanks for fixing this, @euank!

PR #207, opened 2023-04-07
Fix docs ingress

alex-bezek:
the ingress controller uses the ingress class stable version spec `ingressClassName` and not the old annotation syntax


russorat (approved):
lgtm

PR #212, opened 2023-04-17
Add a 'podLabels' option to the helm chart

euank:
This is an analog to `podAnnotations`.

This seems both relatively common, and like a useful thing to have.

I also made some minor wording tweaks, and I'm happy to teak any that seem off the mark.

### DIFF
--- a/helm/ingress-controller/README.md
+++ b/helm/ingress-controller/README.md
@@ -47,9 +47,9 @@ To uninstall the chart:
 
 | Name                         | Description                                                                                                           | Value                                 |
 | ---------------------------- | --------------------------------------------------------------------------------------------------------------------- | ------------------------------------- |
-| `replicaCount`               | The number of controllers to run.                                                                          | `1`                                   |
-| `podAnnotations`             | Used to apply custom annotations to the controller pods. | `{}`                                  |
-| `podLabels`             | Used to apply custom labels to the controller pods. | `{}`                                  |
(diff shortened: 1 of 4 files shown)


jonstacks (approved):

helm/ingress-controller/README.md
```diff
@@ -47,8 +47,9 @@ To uninstall the chart:
 
 | Name                         | Description                                                                                                           | Value                                 |
 | ---------------------------- | --------------------------------------------------------------------------------------------------------------------- | ------------------------------------- |
-| `podAnnotations`             | Used to inject custom annotations directly into                                                                       | `{}`                                  |
-| `replicaCount`               | The number of controllers and agents to run.                                                                          | `1`                                   |
+| `podAnnotations`             | Used to apply custom annotations to # the ingress                                                                     | `{}`                                  |
```
euank, line 50: This hash got inserted by a github action :(

https://github.com/ngrok/kubernetes-ingress-controller/blob/5c1506f1a7e95167aab2a7d85339e3ebda7f9212/.github/workflows/generate-chart-readme.yaml#L33-L37

I don't like that it picked up a '#' comment in there.

I'll change my comment in the values.yaml to not span multiple lines to hopefully fix the extra '#'.

euank: Oh, actually, it's right. I had a '#' in there myself because I'm not smart and `vim`d wrong.
Sorry for blaming you, readme-generator, turns out this one was on me :(

PR #213, opened 2023-04-17
fix: remove note about alpha and pro account

russorat:
## What
This change removes the comment about this project being in alpha from the notes section. It also removes the note about a pro account since it is no longer required.

## How
Removing the incorrect text

## Breaking Changes
No


krwenholz (approved):
That looks correct!

krwenholz (commented):
Oh and this line if you don't mind: https://github.com/ngrok/kubernetes-ingress-controller/blob/fb67e40641f0637044c30a287758e31a367f645c/docs/developer-guide/releasing.md?plain=1#L33

PR #218, opened 2023-05-07
Alex/security warning

alex-bezek:
## What
Add a comment to the top of the readme pointing out a security concern with the controller at the moment

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,9 @@
   </a>
 </p>
 
+> _*Warning*_: Currently this project has an issue being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/208) which can cause the controller to provide ingress to a service even if it failed to configure an authentication module like Oauth on a particular route due to a configuration or intermittent error. Until this issue is resolved, its not recommended to use this with security sensitive applications.
+
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,8 @@
   </a>
 </p>
 
+> _*Warning*_: Currently this project has an issue being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/208) which can cause the controller to provide ingress to a service even if it failed to configure an authentication module like Oauth on a particular route due to a configuration or intermittent error. Until this issue is resolved, it's not recommended to use this with security sensitive applications.
```
euank, line 31: This makes it sound like it can only happen due to errors ('configuration or intermittent error').

This also happens racily in normal operation of the ingress controller with no errors anywhere. I filed a separate issue for that since I think that's also worth noting here: https://github.com/ngrok/kubernetes-ingress-controller/issues/219

alex-bezek: Updated to include this issue as well

euank (approved):
thank you, I appreciate the level of detail!

This looks accurate from my understanding, and seems worth including a note about.

PR #226, opened 2023-05-15
Some updates to handle different cases for e2e run

bobzilladev:
## What
Changes that made the e2e tests work for me, these may not be universally applicable, so happy to cherry pick anything that is actually useful.

## How

- Typo in developer guide
- Kustomization changes in `e2e-fixtures/hello-world-ingress` to remove `bezek-hello-world-ingress-2` and set via config
- `e2e-fixtures/ingress-class` 404 is HTTP/2 for me
- The secrets keys don't seem to have `NGROK` prefix, unlike the environment variables they set
- `jq` and `yq` weren't installed on this machine, added to nix config, and added some bash to check for them up front. looks like josh is adding nix support though, so this may become irrelevant.
- Changed the test for `GOOGLE_CLIENT_ID` to not error out if not defined, due to the `set -eu -o pipefail`
- A message before the long sleep to make the lack of output more expected

## Breaking Changes
N/A

### DIFF
--- a/e2e-fixtures/hello-world-ingress/EXAMPLE-config-pt2.yaml
+++ b/e2e-fixtures/hello-world-ingress/EXAMPLE-config-pt2.yaml
@@ -3,6 +3,3 @@
 - op: replace
   path: /spec/rules/0/host
   value: <UNIQUE SUBDOMAIN>.ngrok.io
-- op: replace
-  path: /spec/rules/1/host
-  value: <UNIQUE SUBDOMAIN>.ngrok.io
--- a/docs/developer-guide/README.md
+++ b/docs/developer-guide/README.md
@@ -38,7 +38,7 @@ Several examples are provided in the [`e2e-fixtures` folder](https://github.com/
 Then, you need to update the `value` field in that new file.
 
 You can then apply the given example via `kubectl apply -k e2e-fixtures/<example in question>`, i.e.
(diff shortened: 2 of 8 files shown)


e2e-fixtures/hello-world-ingress/EXAMPLE-config-pt2.yaml
```diff
@@ -0,0 +1,5 @@
+# 1. Copy this file to ./config.yaml
+# 2. Set you unique FQDN in the "value" field below
+- op: replace
+  path: /spec/rules/0/host
+  value: <UNIQUE SUBDOMAIN>.ngrok.io
```
alex-bezek, line 5: if this is just to override those bezek specific ones, feel free to rip all that apart and change up the base file. These got out of date near the end since I was the only one using them and all the modulset stuff changed really quickly

alex-bezek:
So i was the only one using these for the most part and near the end of the sprints things changed heavily and these got out of date. They still have the old annoation system we had before we used crds https://github.com/ngrok/kubernetes-ingress-controller/blob/main/e2e-fixtures/route-modules/route-modules.yaml#L6-L11 

We may want to chat with you and the team about if you find these useful or if there are some changes we can make to make things a bit better. I think these tests end up having various state problems and actually were hitting api rate limits quite frequently

Issue #200, opened 2023-03-29
Set the Agent Version 


alex-bezek:
### Description

In the ngrok dashboard you can list your online agents with their version. The ngrok-go agents in the ngrok ingress controller are showing up as `0.0.0` the the dashboard. ngrok-go recently released a way to set a child client to set our own version
https://github.com/ngrok/ngrok-go/blob/main/session.go#L186

### Use Case

The user may have many instances of controllers, agents, and client libraries running in their account and would want to distinguish between the versions.

PR #227, opened 2023-05-16
misc: add direnv and a nix devshell

jrobsonchase:
Add a .envrc and flake.nix to provision development tooling via nix and
direnv.

PR #228, opened 2023-05-16
tunneldriver: plumb the version through ngrok-go

jrobsonchase:
closes: #200 

## What

Surface the client type/version in the ngrok-go session.

## How

Pull in the latest version of ngrok-go and use its `WithChildClient`
option to set the client type/version for display in the ngrok
dashboard.

~~Also some plumbing via go:embed to get the version from the `VERSION`
file.~~


jrobsonchase:
Current dependencies on/for this PR:
* main
  * **PR #227** <a href="https://app.graphite.dev/github/pr/ngrok/kubernetes-ingress-controller/227" target="_blank">Graphite</a> 
    * **PR #228** <a href="https://app.graphite.dev/github/pr/ngrok/kubernetes-ingress-controller/228" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ngrok/kubernetes-ingress-controller/228?utm_source=stack-comment).

nikolay-ngrok (approved):
👍

main.go
```diff
@@ -56,6 +57,9 @@ var (
 	setupLog = ctrl.Log.WithName("setup")
 )
 
+//go:embed VERSION
+var version string
+
```
alex-bezek, line 62: Oh this is really nifty. Do you think we can/should converge this with the other way we get the version via a build flag?
https://github.com/ngrok/kubernetes-ingress-controller/blob/main/internal/version/version.go#L11
https://github.com/ngrok/kubernetes-ingress-controller/blob/main/.github/workflows/docker.yaml#L51

jonstacks: 👍 to consolidating either way. My preference would be on the internal version package so we can get the git sha as well

jrobsonchase: Oops, my repo wasn't up to date! Explains why I didn't find this.

Moved the VERSION file to that package and added a symlink to the repo root, and replaced the ldflag with the embed.

jonstacks (changes requested):
I think moving this will break some of the release automation.

jrobsonchase:
@jonstacks Does any of that require `VERSION` to be a *file*? It's still where it was before - just as a symlink. `go:embed` has a hard requirement on it being a regular file beneath the current directory tree.

jonstacks (approved):

PR #234, opened 2023-05-23
Add e2e config init script

CarlAmko:
## What
Add a script to make e2e initialization less tedious

## How
Added `./scripts/init_e2e_configs.sh`.

## Breaking Changes
No


scripts/init_e2e_configs.sh
```diff
@@ -0,0 +1,30 @@
...
+    filename="${file##*/}"
+    target_filename="${filename#$prefix}"
+    target_file="$directory/$target_filename"
+
+    # Copy the file over, sans prefix
+    cp "$file" "$target_file"
+
+    # Replace text within the copied file
+    sed -i "s/<UNIQUE SUBDOMAIN>/$domain/g" "$target_file"
+    sed -i "s/<UNIQUE SUBDOMAIN 2>/$domain/g" "$target_file"
```
bobzilladev, line 29: I did need two different domains to get the `ingress-class` test to work.

bobzilladev (commented):
Got my e2e working again, and now notice one of my url's is ngrok.app and not ngrok.io, but this only replaces an ngrok.io. Maybe we need full url replacement instead of prefix.

Made this change to make testing e2e itself fasterer: https://github.com/ngrok/kubernetes-ingress-controller/pull/235

alex-bezek:
> Got my e2e working again, and now notice one of my url's is ngrok.app and not ngrok.io, but this only replaces an ngrok.io. Maybe we need full url replacement instead of prefix.

That or defaulting to .app might work as well if we only want to use the global domain

bobzilladev:
I'd be happy with all .app, the future is now!

CarlAmko:
> Got my e2e working again, and now notice one of my url's is ngrok.app and not ngrok.io, but this only replaces an ngrok.io. Maybe we need full url replacement instead of prefix.

Yeah, plus that would be more in line with the latest agent parameter (which removes subdomain). That seems out of scope for this PR though, no? The PR that adjusts it from subdomain to FQDN can modify this script too the new placeholder text.

bobzilladev:
Do you want to s/io/app/ in this PR? Then it works going forward, don't have to fqdn right now.

bobzilladev (approved):
wfm, thanks!

docs/examples/hello-world/manifests.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
   ingressClassName: ngrok
   rules:
-    - host: ${NGROK_SUBDOMAIN}.ngrok.io
+    - host: ${NGROK_SUBDOMAIN}.ngrok.app
```
alex-bezek, line 8: Maybe don't update this, or update the md file as well so they remain consistent https://github.com/ngrok/kubernetes-ingress-controller/blob/main/docs/examples/hello-world/README.md

PR #235, opened 2023-05-23
use raw yq output, split e2e runner from deployment

bobzilladev:
## What

Make it so the e2e tests can be easily run without waiting for a full deployment.
This also adds a `-r` to `yq` so it gets raw output, even on the version 3 from nix (version 4 does the happy thing by default).

## How

Splits out the e2e running into `scripts/postflight.sh` which is then exec'd from `scripts/e2e.sh`.

## Breaking Changes

No

### DIFF
--- a/scripts/e2e.sh
+++ b/scripts/e2e.sh
@@ -54,46 +54,4 @@ done
 echo "--- Waiting for deployment"
 sleep 120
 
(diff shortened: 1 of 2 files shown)


CarlAmko (approved):
Dig it!

PR #237, opened 2023-05-24
Better handle errors encountered while calculating HTTPSEdges

### LINKED ISSUE #233: Missing named objects do not fatal error out at runtime
## Problem

Invalid or missing named objects / resources can lead to invalid configuration at runtime leading to issues like #208 where endpoints can be left in an insecure state.

Example

```yaml
kind: IPPolicy
metadata:
  name: test-policy
spec:
  description: "Deny all policy"
  rules:
  - action: "deny"
    cidr: 0.0.0.0/0
    description: "Deny all"
---
kind: NgrokModuleSet
metadata:
  name: ip-restrictions
modules:
  ipRestriction:
    policies:
    - "test-policy-typo"
```

## Expected behavior

- Check existence of all named resources / objects that are being used
- Error when named resources / objects do not exist

Wonderful-Space:
Resolves #233 

## What
* Errors weren't caught before when resolving modules for HTTPS edges. They now are caught and prevent corresponding deployments from launching.

## How
* Errors are forwarded to tunnel calculators, which prevent the matching tunnel resources from launching.
* Submits error events when reconciling edges.

## Breaking Changes
N/A

## Validation
1. Run `kubectl apply -f docs/examples/bad-module-load/manifest.yml`
2. Run `kubectl get pods`
3. Dump the logs for the ingress pod. It should be named something like: 
```
kubectl logs ngrok-ingress-controller-kubernetes-ingress-controller-<HASH> -f
```
4. Observe logs for errors stemming from bad module specifications:
image
5. Check in dashboard that edges are not created for failed modules:
image

### DIFF
--- a/internal/store/driver.go
+++ b/internal/store/driver.go
@@ -37,11 +37,10 @@ type Driver struct {
 	reentranceFlag        int64
 	bypassReentranceCheck bool
 	customMetadata        string
-	recorder              record.EventRecorder
 }
 
 // NewDriver creates a new driver with a basic logger and cache store setup
-func NewDriver(logger logr.Logger, scheme *runtime.Scheme, controllerName string, recorder record.EventRecorder) *Driver {
+func NewDriver(logger logr.Logger, scheme *runtime.Scheme, controllerName string) *Driver {
 	cacheStores := NewCacheStores(logger)
 	s := New(cacheStores, controllerName, logger)
 	return &Driver{
@@ -49,7 +48,6 @@ func NewDriver(logger logr.Logger, scheme *runtime.Scheme, controllerName string
 		cacheStores: cacheStores,
 		log:         logger,
 		scheme:      scheme,
-		recorder:    recorder,
 	}
 }
 
@@ -394,7 +392,6 @@ func (d *Driver) calculateHTTPSEdges(domains []ingressv1alpha1.Domain) ([]ingres
 			if err != nil {
 				d.log.Error(err, "error getting ngrok moduleset for ingress", "ingress", ingress)
 				module_load_failure = true
-				d.recorder.Event(&edge, corev1.EventTypeWarning, "NgrokModuleLoadFailure", err.Error())
 			}
 
 			// Set edge specific modules
--- a/internal/store/driver.go
+++ b/internal/store/driver.go
@@ -382,7 +382,7 @@ func (d *Driver) calculateHTTPSEdges() []ingressv1alpha1.HTTPSEdge {
 			modSet, err := d.getNgrokModuleSetForIngress(ingress)
 			if err != nil {
 				d.log.Error(err, "error getting ngrok moduleset for ingress", "ingress", ingress)
-				continue
+				panic(err)
 			}
 
 			// Set edge specific modules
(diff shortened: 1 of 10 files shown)


internal/store/driver.go
```diff
@@ -382,7 +382,7 @@ func (d *Driver) calculateHTTPSEdges() []ingressv1alpha1.HTTPSEdge {
 			modSet, err := d.getNgrokModuleSetForIngress(ingress)
 			if err != nil {
 				d.log.Error(err, "error getting ngrok moduleset for ingress", "ingress", ingress)
-				continue
+				panic(err)
```
nikolay-ngrok, line 385: This sounds too invasive to me, especially since it is bringing all tunnels down if the pods are crashlooping?

CarlAmko:
Open question here is whether or not this is the only place that named resources are referenced.

internal/store/driver.go
CarlAmko: @bobzilladev hmm, I don't think so because we wouldn't get to the part where it even [gets to update the tunnels](https://github.com/ngrok/kubernetes-ingress-controller/blob/main/internal/store/driver.go#L276)? If I'm understanding it correctly...

internal/store/driver.go
```diff
@@ -379,10 +389,12 @@ func (d *Driver) calculateHTTPSEdges() []ingressv1alpha1.HTTPSEdge {
 		for _, ingress := range ingresses {
 			namespace := ingress.Namespace
 
+			module_load_failure := false
 			modSet, err := d.getNgrokModuleSetForIngress(ingress)
 			if err != nil {
 				d.log.Error(err, "error getting ngrok moduleset for ingress", "ingress", ingress)
-				continue
+				module_load_failure = true
+				d.recorder.Event(&edge, corev1.EventTypeWarning, "NgrokModuleLoadFailure", err.Error())
```
nikolay-ngrok, line 397: we are just loading the `NgrokModuleSet` objects from the k8s api, this feels like the wrong place to put this. The right place should be somewhere in `httpsedge_controller.go`, where it tries to set the modules for and edge/route.

CarlAmko: You're referring to the recorder event?

nikolay-ngrok: Yeah, on a high level, I'm referring to the fact that its not an issue that this fails, but the reconciliation process that happens over the httpsedges later on is what will error/we need to log

CarlAmko: Got it 👍🏼 I think moving it to the edge reconciler will also let us catch other initialization failures too.

internal/store/driver.go
```diff
@@ -162,8 +162,8 @@ func (d *Driver) Sync(ctx context.Context, c client.Client) error {
 
 	d.log.Info("syncing driver state!!")
 	desiredDomains := d.calculateDomains()
-	desiredEdges := d.calculateHTTPSEdges()
-	desiredTunnels := d.calculateTunnels()
+	desiredEdges, failedEdgeServices := d.calculateHTTPSEdges(desiredDomains)
```
nikolay-ngrok, line 165: I think we should revert this `failedEdgeServices` calculation

PR #241, opened 2023-05-26
validate ip restriction rules, before creating the route

nikolay-ngrok:
Fixes: #233 

## What
When the `iprestriction` points to a policy that doesn't exist, we still create the edge, but leave it widely open.

## How
Pre-validate the names. Ignore names that look like ip policy ids (e.g. prefixed with `ipp_`)

## Breaking Changes
The now invalid configs no longer result in an edge route.

### DIFF
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -200,6 +200,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 			if err := routeModuleUpdater.ipPolicyResolver.validateIPPolicyNames(ctx, edge.Namespace, routeSpec.IPRestriction.IPPolicies); err != nil {
 				if apierrors.IsNotFound(err) {
 					r.Recorder.Eventf(edge, v1.EventTypeWarning, "FailedValidate", "Could not validate ip restriction: %v", err)
+					continue
 				}
 
 				return err
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -30,6 +30,7 @@ import (
 	"reflect"
 
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
@@ -193,6 +194,16 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		backend, err := tunnelGroupReconciler.findOrCreate(ctx, routeSpec.Backend)
 		if err != nil {
 			return err
+		}
+
+		if routeSpec.IPRestriction != nil {
+			if err := routeModuleUpdater.ipPolicyResolver.validateIPPolicyNames(ctx, edge.Namespace, routeSpec.IPRestriction.IPPolicies); err != nil {
+				if apierrors.IsNotFound(err) {
+					r.Recorder.Eventf(edge, v1.EventTypeWarning, "Validate", "Could not validate ip restriction: ", err)
+				}
+
+				return err
+			}
 		}
 
 		match := r.getMatchingRouteFromEdgeStatus(edge, routeSpec)
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -199,7 +199,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		if routeSpec.IPRestriction != nil {
 			if err := routeModuleUpdater.ipPolicyResolver.validateIPPolicyNames(ctx, edge.Namespace, routeSpec.IPRestriction.IPPolicies); err != nil {
 				if apierrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


internal/controllers/httpsedge_controller.go
```diff
@@ -195,6 +196,16 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 			return err
 		}
 
+		if routeSpec.IPRestriction != nil {
+			if err := routeModuleUpdater.ipPolicyResolver.validateIPPolicyNames(ctx, edge.Namespace, routeSpec.IPRestriction.IPPolicies); err != nil {
+				if apierrors.IsNotFound(err) {
+					r.Recorder.Eventf(edge, v1.EventTypeWarning, "Validate", "Could not validate ip restriction: ", err)
+				}
```
nikolay-ngrok, line 203: maybe we should ignore the whole route, and continue with the other ones here 🤔

internal/controllers/utils.go
```diff
@@ -58,6 +74,7 @@ func (r *ipPolicyResolver) resolveIPPolicyNamesorIds(ctx context.Context, namesp
 		if err := r.client.Get(ctx, types.NamespacedName{Name: nameOrId, Namespace: namespace}, policy); err != nil {
 			if client.IgnoreNotFound(err) == nil {
 				m[nameOrId] = true // assume it's an ID
+				continue
```
nikolay-ngrok, line 77: this actually allow us to use `ipp_`, e.g. preexisting policies not managed in k8s.

nikolay-ngrok:
> I'm thinking like here do we want to catch all module resolution errors?

the rest of the modules don't refer to anything by name. My understanding of #233 is that we are trying to use something that should be already defined as a custom resource (ip_policy), but for some reason we are not finding it yet.

internal/controllers/httpsedge_controller.go
alex-bezek: I think that would probably be better. Because all of these steps are done inside this loop over the routes, this could make it have confusing updates where a handful of routes are handled until a bad one errors and skips the rest. 

I think as long as the affected route is not half exposed, trying to keep the rest of the routes up and configure them properly would be the better user experience

internal/controllers/httpsedge_controller.go
```diff
@@ -195,6 +196,16 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 			return err
 		}
 
+		if routeSpec.IPRestriction != nil {
+			if err := routeModuleUpdater.ipPolicyResolver.validateIPPolicyNames(ctx, edge.Namespace, routeSpec.IPRestriction.IPPolicies); err != nil {
+				if apierrors.IsNotFound(err) {
+					r.Recorder.Eventf(edge, v1.EventTypeWarning, "Validate", "Could not validate ip restriction: ", err)
```
alex-bezek, line 202: A bit more information here would be super useful to the user, specifically the IP policy name that can't be found and the namespace. Something like `IPPolicy "ip-policy-typo" not found in namespace "default"` would make it very obvious that they misspelled something or forgot to apply. The meaning of "validate" is just a little vague IMO

internal/controllers/utils.go
```diff
@@ -49,6 +50,21 @@ type ipPolicyResolver struct {
 	client client.Reader
 }
 
+func (r *ipPolicyResolver) validateIPPolicyNames(ctx context.Context, namespace string, namesOrIds []string) error {
+	for _, nameOrId := range namesOrIds {
+		if strings.HasPrefix(nameOrId, "ipp_") && len(nameOrId) == 31 { // TODO better validate
```
alex-bezek, line 55: I was concerned about some edge cases here, but I forgot that k8s doesn't actually allow underscores in names, so this is actually seems pretty solid

```
The IPPolicy "ipp_test" is invalid: metadata.name: Invalid value: "ipp_test": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.'
```

internal/controllers/utils.go
alex-bezek: nice find!

alex-bezek:
This is probably out of scope for this issue, but I wonder if the moduleSet accepting an array of BOTH ip policy names and IDs is actually confusing and instead we should just have `ip_policy_names` and `ip_policy_ids` as separate fields. It would be nonpassive, but something to consider for the future

internal/controllers/httpsedge_controller.go
nikolay-ngrok: Current event log from the controller:
```
2023-05-31T14:44:33Z	DEBUG	events	Could not validate ip restriction: IPPolicy.ingress.k8s.ngrok.com "deny-policy-1" not found	{"type": "Warning", "object": {"kind":"HTTPSEdge","namespace":"ngrok-ingress-controller","name":"ngrok-ingress-npe-full-test-ngrok-io","uid":"039d1019-37b2-402b-9179-0d49470ea23a","apiVersion":"ingress.k8s.ngrok.com/v1alpha1","resourceVersion":"325041"}, "reason": "FailedValidate"}
```

and from events:
```
ngrok-ingress-controller   4m52s (x2 over 5m2s)   Warning   FailedValidate      HTTPSEdge/ngrok-ingress-npe-full-test-ngrok-io                                         Could not validate ip restriction: IPPolicy.ingress.k8s.ngrok.com "deny-policy-1" not found
```

Wonderful-Space (approved):

alex-bezek (approved):

PR #242, opened 2023-05-30
Add graphite cli to Nix env

Wonderful-Space:
## What
Add [Graphite](https://graphite.dev) to Nix env

## How
Added `graphite-cli` to `flake.nix`

## Breaking Changes
No

image

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,6 +18,7 @@
           kubernetes-helm
           jq
           yq
+          graphite-cli
         ];
       };
     }));


jrobsonchase (changes requested):
Mmm, I'm not sure about putting optional workflow tools in the devshell for these public projects. It makes sense for our monorepo where we're standardizing on internal workflows/process, but public projects feel like they should include *just* the things necessary for building and running that specific project.

Wonderful-Space:
> Mmm, I'm not sure about putting optional workflow tools in the devshell for these public projects. It makes sense for our monorepo where we're standardizing on internal workflows/process, but public projects feel like they should include _just_ the things necessary for building and running that specific project.

Yeah, I wasn't entirely sure if we should do it. I didn't think it'd hurt a whole lot, but I suppose keeping the criteria for selecting Nix packages makes sense.

Wonderful-Space:
Something [like this](https://github.com/hercules-ci/flake-parts) sounds like a way we could pull it off. But perhaps not enough bang for buck here. 🤷🏼

Issue #221, opened 2023-05-11
Some update transitions terminally error out


euank:
### What happened

There are a number of updates that do not work and result in the ingress being stuck in a terminal state.

I'll list a few, but I expect there are quite a few more:

**IPolicy duplicate-ish rules**

Applying the following diff:

```diff
 apiVersion: ingress.k8s.ngrok.com/v1alpha1
 kind: IPPolicy
 metadata:
   name: test-policy
 spec:
   rules:
-  - action: allow
+  - action: deny
     cidr: 127.0.0.1/32
```

results in the terminal error:

```
2023-05-11T07:57:48Z	ERROR	Reconciler error	{"controller": "ippolicy", "controllerGroup": "ingress.k8s.ngrok.com", "controllerKind": "IPPolicy", "IPPolicy": {"name":"test-policy","namespace":"default"}, "namespace": "default", "name": "test-policy", "reconcileID": "3d692c18-9bd3-4275-b2d3-ed559c612093", "error": "HTTP 400: IP Policy Rule with CIDR '127.0.0.1/32' already exists [ERR_NGROK_1408]\n\nOperation ID: op_2PdfRGQmUcjcjXkqIRnvGASQQyr"}
```

even though the start and end states are both valid IP Policies.

**OIDC->OAuth**

Creating an edge with a valid OIDC config, and then deleting the OIDC config and replacing it with an OAuth config (in one `kubectl apply`) is a transition from a valid state to a valid state, but terminally errors out with:

```
2023-05-11T08:19:52Z	ERROR	Reconciler error	{"controller": "httpsedge", "controllerGroup": "ingress.k8s.ngrok.com", "controllerKind": "HTTPSEdge", "HTTPSEdge": {"name":"euank-test-ingress-ngrok-app","namespace":"default"}, "namespace": "default", "name": "euank-test-ingress-ngrok-app", "reconcileID": "35cc5666-885a-4215-acc2-0ac70ee6b893", "error": "HTTP 400: The edge specifies conflicting authentication modules. Only one of SAML, OIDC, or OAuth may be enabled. [ERR_NGROK_7079]\n\nOperation ID: op_2Pdi7YrFWhVcycMnDATcHy57itM"}
```

**Changing route path-type**

```diff
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: test
 spec:
   ingressClassName: ngrok
   rules:
     - host: euank-test-ingress.ngrok.app
       http:
[...]

PR #243, opened 2023-05-31
Initialize route backends after module updates

### LINKED ISSUE #208: ingress controller "fails open" if modules have issues
Let's look at a concrete example: we attempt to apply an IP policy that prevents all traffic from reaching the service, but typo the name.

```yaml
kind: IPPolicy
metadata:
  name: test-policy
spec:
  description: "Deny all policy"
  rules:
  - action: "deny"
    cidr: 0.0.0.0/0
    description: "Deny all"
---
kind: NgrokModuleSet
metadata:
  name: ip-restrictions
modules:
  ipRestriction:
    policies:
    - "test-policy-typo"
---
kind: Ingress
metadata:
  name: game-2048
  annotations:
    k8s.ngrok.com/modules: ip-restrictions
... (truncated)
<rest below>
```

<Details>

<Summary>Full example</Summary>

```yaml
kind: IPPolicy
metadata:
  name: test-policy
spec:
  description: "Deny all policy"
  rules:
  - action: "deny"
    cidr: 0.0.0.0/0
    description: "Deny all"
---
kind: NgrokModuleSet
metadata:
  name: ip-restrictions
modules:
  ipRestriction:
    policies:
    - "test-policy-typo"
---
kind: Ingress
metadata:
  name: game-2048
  annotations:
    k8s.ngrok.com/modules: ip-restrictions
spec:
  ingressClassName: ngrok
  rules:
    - host: euank-test-ingress.ngrok.app
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: game-2048
                port:
                  number: 80
---
apiVersion: v1
kind: Service
metadata:
  name: game-2048
spec:
  ports:
    - name: http
      port: 80
      targetPort: 80
  selector:
    app: game-2048

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: game-2048
spec:
  replicas: 1
  selector:
    matchLabels:
      app: game-2048
  template:
    metadata:
      labels:
        app: game-2048
    spec:
      containers:
        - name: backend
          image: alexwhen/docker-2048
          ports:
            - name: http
              containerPort: 80
```

</Details>

If we apply the above manifest (modifying the domain to be one on your account of course), we end up with our service being exposed to the internet with no ip policy at all.

I think that any error configuring modules should fail closed. This is especially true for auth and security related modules (webhook verification, oauth, etc etc), but I think it's true for all modules.

Wonderful-Space:
Resolves #208

## What
Applying edge routes' backend is now deferred until after route modules are successfully applied. If route modules fail to apply for any reason, the route's backend remains unset, thus closing off access to the edge from that route.

This approach was inspired by [Niji's comment](https://github.com/ngrok/kubernetes-ingress-controller/issues/208#issuecomment-1553703178) in the parent issue.

## How
1. Omit backend from new routes / Remove backend from current routes
2. Route modules now try to apply
2a. If route module application was successful, routes are then updated with their specified backend
2b. If unsuccessful, `reconcileRoutes` returns an error and the route remains backend-less
3. Route statuses are updated

## Breaking Changes
Edge routes will no longer be valid if an error occurs while applying route modules.

## Validation

Given this sample `manifest.yml`:
<details><summary>Expand</summary>
<p>

```yml
kind: NgrokModuleSet
metadata:
  name: auth
modules:
  oauth:
    google:
      optionsPassthrough: true
      inactivityTimeout: 10m
      maximumDuration: 24h
      authCheckInterval: 20m
      clientSecret:
        name: auth
        key: client-secret
      emailDomains:
      - ngrok.com
---
kind: Ingress
metadata:
  name: game-2048
  annotations:
    k8s.ngrok.com/modules: auth
spec:
  ingressClassName: ngrok
  rules:
    - host: test-ingress-12345.ngrok.app
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: game-2048
                port:
                  number: 80
---
apiVersion: v1
kind: Service
metadata:
  name: game-2048
spec:
  ports:
    - name: http
      port: 80
      targetPort: 80
  selector:
    app: game-2048

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: game-2048
spec:
  replicas: 1
  selector:
    matchLabels:
      app: game-2048
  template:
    metadata:
      labels:
        app: game-2048
    spec:
      containers:
        - name: backend
          image: alexwhen/docker-2048
          ports:
            - name: http
              containerPort: 80
```

</p>
</details> 

### Creating a new route 
1. Apply the manifest `kubectl apply -f manifest.yml`
2. Check the ingress controller pod logs `kubectl logs ngrok-ingress-controller-kubernetes-ingress-controller-<HASH> -f`
4. Observe for error logs / events suggesting a route module application failure
image
5. Attempting to hit the route results in an ngrok 6015 error / dashboard error
image

image

### Updating a current route
1. Edit the above manifest to remove the `clientSecret` configuration.
2. Re-apply the manifest to create a valid route + backend
image
3. Now undo the edits and attempt to apply a `clientSecret`, like above.
4. This should fail, as there are no secrets loaded into your environment.
5. Observe that backend is still running with the previous (valid) configuration.

## Future Work
* Check for all route prerequisites before attempting any route updates. This would ideally combine with the current `ipPolicies` validation so that we can evaluate other module criteria (such as the correctly set secrets) before executing updates.
* Update this approach to incorporate better wholesale validation before applying updates to minimize downtime. [Niji](https://github.com/ngrok/kubernetes-ingress-controller/pull/243#discussion_r1218499023) captured some theoretical approaches to solving this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@
   </a>
 </p>
 
-> _*Warning*_: Currently this project has an issue being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/208) which can cause the controller to provide ingress to a service even if it failed to configure an authentication module like Oauth on a particular route due to a configuration or intermittent error. Additionally, even when there aren't errors, the controller first sets up ingress and then applies route modules like Oauth immediately after, but there is a brief period where ingress without authentication is provided. This issue is being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/219). Until this issue is resolved, it's not recommended to use this with security sensitive applications.
+> _*Warning*_: The controller first sets up ingress and then applies route modules like Oauth immediately after, but there is a brief period where ingress without authentication is provided. This issue is being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/219). Until this issue is resolved, it's not recommended to use this with security sensitive applications.
 
 # ngrok Ingress Controller for Kubernetes
 
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@
   </a>
 </p>
 
-> _*Warning*_: The controller first sets up ingress and then applies route modules like Oauth immediately after, but there is a brief period where ingress without authentication is provided. This issue is being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/219). Until this issue is resolved, it's not recommended to use this with security sensitive applications.
+> _*Warning*_: Currently this project has an issue being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/208) which can cause the controller to provide ingress to a service even if it failed to configure an authentication module like Oauth on a particular route due to a configuration or intermittent error. Additionally, even when there aren't errors, the controller first sets up ingress and then applies route modules like Oauth immediately after, but there is a brief period where ingress without authentication is provided. This issue is being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/219). Until this issue is resolved, it's not recommended to use this with security sensitive applications.
 
 # ngrok Ingress Controller for Kubernetes
 
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -188,6 +188,8 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		secretResolver:   secretResolver{r.Client},
 	}
 
+	edgeRoutes := r.NgrokClientset.HTTPSEdgeRoutes()
+
 	// TODO: clean this up. This is way too much nesting
 	for i, routeSpec := range edge.Spec.Routes {
 		match := r.getMatchingRouteFromEdgeStatus(edge, routeSpec)
@@ -200,7 +202,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 				Match:     routeSpec.Match,
 				MatchType: routeSpec.MatchType,
 			}
-			route, err = r.NgrokClientset.HTTPSEdgeRoutes().Create(ctx, req)
+			route, err = edgeRoutes.Create(ctx, req)
 		} else {
 			// Otherwise, update the current route to disable its backend
 			req := &ngrok.HTTPSEdgeRouteUpdate{
@@ -209,7 +211,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 				Match:     routeSpec.Match,
 				MatchType: routeSpec.MatchType,
 			}
-			route, err = r.NgrokClientset.HTTPSEdgeRoutes().Update(ctx, req)
+			route, err = edgeRoutes.Update(ctx, req)
 		}
 		if err != nil {
 			return err
@@ -246,7 +248,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 				BackendID: backend.ID,
 			},
 		}
-		route, err = r.NgrokClientset.HTTPSEdgeRoutes().Update(ctx, req)
+		route, err = edgeRoutes.Update(ctx, req)
 		if err != nil {
 			return err
 		}
@@ -270,7 +272,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		}
 		if !found {
 			r.Log.Info("Deleting route", "edgeID", edge.Status.ID, "routeID", remoteRoute.ID)
-			if err := r.NgrokClientset.HTTPSEdgeRoutes().Delete(ctx, &ngrok.EdgeRouteItem{EdgeID: edge.Status.ID, ID: remoteRoute.ID}); err != nil {
+			if err := edgeRoutes.Delete(ctx, &ngrok.EdgeRouteItem{EdgeID: edge.Status.ID, ID: remoteRoute.ID}); err != nil {
 				return err
 			}
 		}
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -209,8 +209,8 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		var route *ngrok.HTTPSEdgeRoute
 		// Now we go ahead and create the route if it doesn't exist, or find the existing route.
 		// It's important to note here that we are intentionally ommiting the `route.Backend` for new routes, and
-		//  removing it from existing routes. This is because we depend on the route modules being successfully applied
-		//  before we apply the new route. This ensures that any route with a backend is considered to be successfully configured.
+		//  removing it from existing routes. The success or failure of applying a route's modules is then strongly 
+		//  linked the state of its backend. Thus, any route with a backend is considered properly configured.
 		//  See https://github.com/ngrok/kubernetes-ingress-controller/issues/208 for additional context.
 		if match == nil {
 			r.Log.Info("Creating new route", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType)
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -252,17 +252,11 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		if err := routeModuleUpdater.updateModulesForRoute(ctx, route, &routeSpec); err != nil {
 			r.Recorder.Event(edge, v1.EventTypeWarning, "RouteModuleUpdateFailed", err.Error())
 
-			codes := make([]int, 0, 100)
-			for i := 400; i <= 500; i++ {
-				codes = append(codes, i)
-			}
-
-			// Ignore 400 error codes as unretryable
-			if ngrok.IsErrorCode(err, codes...) {
+			if errors.IsRetryable(err) {
+				return err
+			} else {
 				return errors.NewErrInvalidConfiguration(err)
 			}
-
-			return err
 		}
 
 		// The route modules were successfully applied, so now we update the route with its specified backend
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -122,8 +122,8 @@ func (r *HTTPSEdgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// Configuration errors can't be resolved, so don't requeue
 	case errors.IsErrInvalidConfiguration(err):
 		return ctrl.Result{}, nil
+	// No error or is a retryable error, so return as normal
 	default:
-		// No error or is a retryable error, so return as normal
 		return ctrl.Result{}, err
 	}
 }
@@ -264,7 +264,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		// TODO: Do an entropy check here to avoid unnecessary updates
 		req := &ngrok.HTTPSEdgeRouteUpdate{
 			EdgeID:    edge.Status.ID,
-			ID:        match.ID,
+			ID:        route.ID,
 			Match:     routeSpec.Match,
 			MatchType: routeSpec.MatchType,
 			Backend: &ngrok.EndpointBackendMutate{
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -118,13 +118,10 @@ func (r *HTTPSEdgeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	err := r.reconcileEdge(ctx, edge)
-	switch {
-	// Configuration errors can't be resolved, so don't requeue
-	case errors.IsErrInvalidConfiguration(err):
-		return ctrl.Result{}, nil
-	// No error or is a retryable error, so return as normal
-	default:
+	if errors.IsErrorReconcilable(err) {
 		return ctrl.Result{}, err
+	} else {
+		return ctrl.Result{}, nil
 	}
 }
 
@@ -251,12 +248,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		r.Log.Info("Applying route modules", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType)
 		if err := routeModuleUpdater.updateModulesForRoute(ctx, route, &routeSpec); err != nil {
 			r.Recorder.Event(edge, v1.EventTypeWarning, "RouteModuleUpdateFailed", err.Error())
-
-			if errors.IsRetryable(err) {
-				return err
-			} else {
-				return errors.NewErrInvalidConfiguration(err)
-			}
+			return err
 		}
 
 		// The route modules were successfully applied, so now we update the route with its specified backend
@@ -740,7 +732,7 @@ func (u *edgeRouteModuleUpdater) setEdgeRouteOAuth(ctx context.Context, route *n
 	}
 
 	if module == nil {
-		return errors.NewErrInvalidConfiguration(fmt.Errorf("no OAuth provider found"))
+		return errors.NewErrInvalidConfiguration(fmt.Errorf("no OAuth provider configured"))
 	}
 
 	if reflect.DeepEqual(module, route.OAuth) {
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -251,6 +251,17 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		r.Log.Info("Applying route modules", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType)
 		if err := routeModuleUpdater.updateModulesForRoute(ctx, route, &routeSpec); err != nil {
 			r.Recorder.Event(edge, v1.EventTypeWarning, "RouteModuleUpdateFailed", err.Error())
+
+			codes := make([]int, 0, 100)
+			for i := 400; i <= 500; i++ {
+				codes = append(codes, i)
+			}
+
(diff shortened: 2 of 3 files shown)


internal/controllers/httpsedge_controller.go
```diff
@@ -210,50 +208,70 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 		match := r.getMatchingRouteFromEdgeStatus(edge, routeSpec)
 		var route *ngrok.HTTPSEdgeRoute
 		if match == nil {
-			r.Log.Info("Creating new route", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType, "backendID", backend.ID)
+			r.Log.Info("Creating new route", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType)
 			// This is a new route, so we need to create it
```
nikolay-ngrok, line 212: lets expand a little bit that we are not setting the backend here because we haven't constructed the route completely?

internal/controllers/httpsedge_controller.go
```diff
@@ -210,50 +208,70 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
...
 				Match:     routeSpec.Match,
 				MatchType: routeSpec.MatchType,
-				Backend: &ngrok.EndpointBackendMutate{
-					BackendID: backend.ID,
-				},
 			}
-			route, err = r.NgrokClientset.HTTPSEdgeRoutes().Create(ctx, req)
+			route, err = edgeRoutes.Create(ctx, req)
 		} else {
-			r.Log.Info("Updating route", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType, "backendID", backend.ID)
```
nikolay-ngrok, line 225: still worthy to have this log line, just saying `Route found, disabling route backend for module apply?`

internal/controllers/httpsedge_controller.go
```diff
@@ -210,50 +208,70 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
...
-			// This is an existing route, so we need to update it
+			// Otherwise, update the current route to disable its backend
```
nikolay-ngrok, line 220: additional explanation of why we are doing it would be nice

internal/controllers/httpsedge_controller.go
```diff
@@ -210,50 +208,70 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
...
+		// Update route status, sans backend
 		routeStatuses[i] = ingressv1alpha1.HTTPSEdgeRouteStatus{
 			ID:        route.ID,
 			URI:       route.URI,
 			Match:     route.Match,
 			MatchType: route.MatchType,
 		}
+
+		// Attempt to apply module updates for a given route
+		r.Log.Info("Evaluating route module updates", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType)
```
nikolay-ngrok, line 242: `Updating route modules`?

nikolay-ngrok (commented):
I think this is working, even tho it has some unexpected errors popping out in the UI

Wonderful-Space:
> it has some unexpected errors popping out in the UI

In the UI? How do you mean? ngrok dashboard?

nikolay-ngrok (approved):

internal/controllers/httpsedge_controller.go
```diff
@@ -209,51 +207,74 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
...
-			// This is an existing route, so we need to update it
+			r.Log.Info("Route found, disabling backend to prepare for module updates", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType)
 			req := &ngrok.HTTPSEdgeRouteUpdate{
 				EdgeID:    edge.Status.ID,
 				ID:        match.ID,
 				Match:     routeSpec.Match,
 				MatchType: routeSpec.MatchType,
-				Backend: &ngrok.EndpointBackendMutate{
-					BackendID: backend.ID,
-				},
```
alex-bezek, line 234: For my clarification, so if you have an Edge with 4 routes and you make any update to the edge, its going to go through and remove the backend for each route temporarily, even if there was no need to update it correct? It would be nice to be able to skip any routes that are already matching what is in the backend API to avoid blipping each endpoint on/off on any change.

I think the data is available via the edge api, but it might be difficult to do all the diffing and would be better done with an atomic API server side.

alex-bezek:
@nijikokun 's comment also talked about creating a static response backend to use as the placeholder while updating

> Static backend that says 'Initializing edge'

image
This message actually makes some sense and does have our styling vs a static response backend is going to be some pretty simple text. I would just verify with Product that what you have without the custom static response backend is an ok user experience

alex-bezek (approved):
Just a couple of questions and maybe future scope items

internal/controllers/httpsedge_controller.go
Wonderful-Space: > if you have an Edge with 4 routes and you make any update to the edge, its going to go through and remove the backend for each route temporarily, even if there was no need to update it correct?

That is correct in the current set of changes.

> It would be nice to be able to skip any routes that are already matching what is in the backend API to avoid blipping each endpoint on/off on any change.

Absolutely agree, I'd love to have something that determines if there is a diff to actually apply. I tried to encapsulate that notion in my notes at the bottom of the description (something like `planModuleUpdates`).

> I think the data is available via the edge api, but it might be difficult to do all the diffing and would be better done with an atomic API server side.

Hmm, okay. Where would you think to put it on the server side then?

Wonderful-Space:
> > the controller still continues to re-sync on failures at its specified interval
> 
> If we get an error that there is no sense in retrying (ie: bad config, real error from ngrok, not just a netwrok flake, etc) you can return a Result object in your reconcile function that specifies not to re-queue it
> 
> * https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile#Reconciler
> * https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile#Result
> 
> You could probably create a specific error type https://github.com/ngrok/kubernetes-ingress-controller/blob/main/internal/errors/errors.go and check for that when returning in the reconciler

Awesome. I'll go ahead and do this. 👍🏼

internal/controllers/httpsedge_controller.go
nijikokun: I'm wondering if we should only do this in this PR for creates based on this statement as the change for updates was an additional portion of my original comment and was an attempt to try and handle the lack of an atomic api.

This means that:

- Creates: No Change
- Update: The way it worked previously (edge is alive even on error)

But, ultimately, Alex is correct that uptime and user-experience where the end-user only experiences an outage from their own mistakes is seemingly better than an enforced one during update.

---

For the future, there could be a few other approaches we could take that don't risk the user-experience without the lack of an atomic api today:

- Create staging-edge with new configuration for validation that things work as intended
  - On success
    - Delete staging-edge
    - Run workflow to create / update existing edges
  - On failure
    - Delete staging-edge
    - Error out

Open questions for this:
- Any limitations around number of objects / billing on additional objects
- If so how to omit from usage

But this would be able to solve not having validation on the client and the create/update issue without atomic api

Will open a new issue for the above, there were alternatives but the downsides were significant:

- Blue-green swapping from staging-edge to production-edge
  - Hostname swapping with downtime
  - Labels swapping
  - Sync issues? Never tried it so not sure how the system would handle live agents when this happens
- Route swapping but routes are limited per account (3-5?) so this would cause issues and is harder to remove that limit

euank: > we are trading user experience for security which doesn't feel great when thinking about usage

The UX feels quite bad, but I think the security issues are worse. If someone's using a security sensitive module, like oauth, ip policies, etc, and we start routing through unauthorized requests, that feels worse than the ingress controller not existing at all.

> forced downtimes can cause interruptions and the shuttering of the service could have greater impacts than a partial one and a partial one would be expected with invalid configuration values.

Forced downtime is still better than a partial outage where the outage is "oauth is no longer being applied, all requests to /admin go through" for example.

It's important to note that completely valid configuration changes will fail to update with how it is written currently, so this doesn't only apply to "invalid configuration values". See https://github.com/ngrok/kubernetes-ingress-controller/issues/221 for a couple examples of these.

------

> Create staging-edge with new configuration for validation that things work as intended
> On success
> * Delete staging-edge
> * Run workflow to create / update existing edges

This doesn't ensure that the transition from the previous edge to the new edge is valid, i.e. for the items mentioned in #221, the new edge would create, but the update still may not be valid.

If we follow this path, I think a simpler pattern is the one you mention below of doing blue/green swaps:

1. Create a new edge with the new configuration
2. Delete the hostport from the old edge
3. Add the hostport to the new edge
4. Delete the old edge

If we're okay having two edges (an "active" and "staging" one), we might as well just shuffle the host port over rather than delete the new one and try to transition.

This solution also fixes the update transition issues since it lets us treat edges as entirely immutable once they're "active". This one seems pretty compelling.

internal/controllers/httpsedge_controller.go
```diff
@@ -189,71 +197,91 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
...
 			return err
 		}
-		routeStatuses[i] = ingressv1alpha1.HTTPSEdgeRouteStatus{
+
+		// With the route properly staged, we now attempt to apply its module updates
+		// TODO: Check if there are no updates to apply here to skip any unnecessary disruption
+		r.Log.Info("Applying route modules", "edgeID", edge.Status.ID, "match", routeSpec.Match, "matchType", routeSpec.MatchType)
+		if err := routeModuleUpdater.updateModulesForRoute(ctx, route, &routeSpec); err != nil {
+			r.Recorder.Event(edge, v1.EventTypeWarning, "RouteModuleUpdateFailed", err.Error())
+			return errors.NewErrInvalidConfiguration(err.Error())
```
euank, line 254: This seems like it turns any ngrok api error into an unretriable "invalid configuration" error.

If I understand correctly, that means a network timeout, or ngrok 5xx, or whatever, would result in us giving up on the update and leaving things in a broken state.

Shouldn't we be retrying network issues, and looking for specific ngrok error codes that indicate invalid configuration?

internal/errors/errors.go
```diff
@@ -135,3 +135,21 @@ func IsErrMissingRequiredSecret(err error) bool {
...
 	return ok
 }
+
+type ErrInvalidConfiguration struct {
+	message string
+}
+
+func NewErrInvalidConfiguration(message string) ErrInvalidConfiguration {
+	return ErrInvalidConfiguration{message: message}
+}
```
euank, line 145: All the callers have an error, so I believe this should take `cause error` as an argument and do usual `errors.Unwrap` go stuff.

internal/controllers/httpsedge_controller.go
```diff
@@ -189,71 +197,91 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
...
 
 		if routeSpec.IPRestriction != nil {
 			if err := routeModuleUpdater.ipPolicyResolver.validateIPPolicyNames(ctx, edge.Namespace, routeSpec.IPRestriction.IPPolicies); err != nil {
 				if apierrors.IsNotFound(err) {
 					r.Recorder.Eventf(edge, v1.EventTypeWarning, "FailedValidate", "Could not validate ip restriction: %v", err)
 					continue
 				}
-
-				return err
+				return errors.NewErrInvalidConfiguration(err.Error())
```
euank, line 211: Same comment regarding this error; we don't know if the returned error from `validateIPPolicyNames` is a 5xx or network error or what, and we want to retry those.

README.md
```diff
@@ -28,7 +28,7 @@
   </a>
 </p>
 
-> _*Warning*_: Currently this project has an issue being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/208) which can cause the controller to provide ingress to a service even if it failed to configure an authentication module like Oauth on a particular route due to a configuration or intermittent error. Additionally, even when there aren't errors, the controller first sets up ingress and then applies route modules like Oauth immediately after, but there is a brief period where ingress without authentication is provided. This issue is being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/219). Until this issue is resolved, it's not recommended to use this with security sensitive applications.
+> _*Warning*_: The controller first sets up ingress and then applies route modules like Oauth immediately after, but there is a brief period where ingress without authentication is provided. This issue is being tracked [here](https://github.com/ngrok/kubernetes-ingress-controller/issues/219). Until this issue is resolved, it's not recommended to use this with security sensitive applications.
```
euank, line 31: The new message doesn't make it clear that there are still terminal cases where the controller will, even with correct authentication modules configured, leave your services exposed to the internet.
Namely, unless I'm reading this PR incorrectly, valid updates can still leave you in terminal insecure states.

I think the new wording, with "brief period", is misleading since it's easily possible to also end up with non-brief periods of insecurity.

internal/controllers/httpsedge_controller.go
```diff
@@ -189,71 +197,91 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
...
-				Backend: &ngrok.EndpointBackendMutate{
-					BackendID: backend.ID,
-				},
 			}
-			route, err = r.NgrokClientset.HTTPSEdgeRoutes().Create(ctx, req)
+			route, err = edgeRoutes.Create(ctx, req)
+
+			// Update status for newly created route
+			routeStatuses[i] = ingressv1alpha1.HTTPSEdgeRouteStatus{
+				ID:        route.ID,
```
euank, line 233: this is a nil pointer exception if `Create` returns `nil, err`, which it's expected to be able to do in various cases.

README.md
nijikokun: This message is to be changed and moved soon, let's keep that out of scope for this PR.

internal/controllers/httpsedge_controller.go
Wonderful-Space: Hmm, fair point. I suppose each individual module call would have to be config error aware.

internal/errors/errors.go
Wonderful-Space: After reworking the individual module apply methods to raise configuration errors, this is not necessarily the case anymore.

internal/controllers/httpsedge_controller.go
Wonderful-Space: It seems like there's a _whole lot_ of [possibilities](https://github.com/ngrok-private/ngrok/blob/0827b5cf976a368dc384b673e560948d900fbfef/js/packages/code-gen/error-code/index.ts) that we could get back, and the errors don't seem to contain that much information (just the error message itself).

Is there a sane way to match against what we could consider a "configuration" error?

internal/errors/errors.go
```diff
@@ -135,3 +136,39 @@ func IsErrMissingRequiredSecret(err error) bool {
...
+
+func (e ErrInvalidConfiguration) Error() string {
+	return fmt.Sprintf("invalid configuration: %s", e.message)
+}
+
+// IsErrInvalidConfiguration: Reflect: returns true if the error is a ErrInvalidConfiguration
+func IsErrInvalidConfiguration(err error) bool {
+	_, ok := err.(ErrInvalidConfiguration)
+	return ok
+}
```
euank, line 156: I don't think we want this function.

Callers should be using:

```go
var invalidConfig ErrInvalidConfiguration
isInvalidConfig := errors.As(err, &invalidConfig)
```

Or, we should implement `func (e ErrInvalidConfiguration) Is(target error) bool`, letting callers do `errors.Is(err, ErrInvalidConfiguration{})`

internal/errors/errors.go
euank: I think my comment got misunderstood, since this is now `return ErrInvalidConfiguration{message: cause.Error()}`.

What I meant by "usual `errors.Unwrap` go stuff" was the following boilerplate:

```go
type ErrInvalidConfiguration struct {
	inner error
}

func NewErrInvalidConfiguration(cause error) error { return ErrInvalidConfiguration{cause} }

func (e ErrInvalidConfiguration) Unwrap() error {
    return e.inner
}
```

Stringifying it loses the inner type. If we want to intentionally lose the inner type, the right way to do that is `errors.Opaque(err)`, which signals intent more specifically, but in this case, I don't think we actually want to lose the inner type.

In addition, the return type should be `error`, not `ErrInvalidConfiguration`, since otherwise type inference will make it annoying if you name it err and try to reuse it, i.e.:

```go
err := NewErrInvalidConfiguration()

// later in the same scope

ret, err := someOtherCall() // compiler error, cannot use err (value of type error) as ErrInvalidConfiguration value in assignment
```

Go's error handling basically requires for you to do runtime type-assertions only (errors.As etc), and at compile-time to squash all types down to just `error`. It's ugly, but with the type system and warts go has, the other options are worse sadly.

internal/errors/errors.go
Wonderful-Space: I assume the reasoning is that the error is readily available after the cast at the call site (rather than discarding it within this kind of function)?

I don't mind adjusting it here, no strong preference from me. I will again highlight that the other errors defined here used this approach, so we'll be going against the grain.

internal/errors/errors.go
Wonderful-Space: Ditto'ing [comment here](https://github.com/ngrok/kubernetes-ingress-controller/pull/243#discussion_r1223044376). Went ahead and changed it as specified.

euank (approved):
Seems like an improvement!

This still doesn't put us in a place where the security issues around modules are solved, but it seems like a good set of changes.

The commits could probably use some rebasing/rewording to make git history read better, but code-wise, this makes sense

internal/errors/errors.go
euank: The goal is to play nice with the `errors.Is`/`errors.As`, since that's the convention the go ecosystem has ended up on, and the best we've got.

It goes with the grain of the go stdlib, but against the grain of this file, which I think didn't realize these broader go conventions existed.

Wonderful-Space:
> Nice, thanks for bearing with me on all my comments!
> 
> The commits could probably use some rebasing/rewording to make git history read better, but code-wise, this makes sense.

Yeah, I'm planning to squash them honestly, just because we changed approaches a few times midflight. Agree!

alex-bezek (approved):

PR #244, opened 2023-06-01
Add namespace to secret

vincetse:
## What
When one generates the manifests with `helm template`, you will notice that the resulting manifest to create the secret credentials is missing the `metadata.namespace` key/value.  This is not a problem when one generates the manifests for use in the `default` namespace.  However, the problem shows itself if the goal is to use a namespace other than `default`.

## How
Adds `namespace` to the `credentials-secret.yaml` template.

## Breaking Changes
No.


jonstacks:
Thanks for the PR, @vincetse. Looks great and nice catch! Can you run `make helm-update-snapshots`? Then it should be good to merge :+1:

vincetse:
Done, @jonstacks.

jonstacks (approved):

PR #256, opened 2023-06-25
feat: Add Support for setting pod/node affinity & anti-affinity

### LINKED ISSUE #254: Add support for setting pod affinity in helm chart
### Description

Allow users to configure pod affinity and anti-affinities through the helm chart `values.yaml`

### Use Case

This allows users to restrict which nodes the ingress controller pods run on or prefer/require that the ingress controller pods don't all run on the same node.

### Related issues

_No response_

jonstacks:
Fixes #254 

## What
Adds support for setting pod & node affinity/anti-affinity via helm chart values.

## How
* Adds `bitnami` common as a dependency for its library of helpers.
* If no `affinity` is set, it adds a default "soft" pod anti-affinity.

## Breaking Changes
No.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified 
 
 .PHONY: deploy
 deploy: docker-build manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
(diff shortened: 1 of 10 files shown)


jonstacks:
Closing this and going to combine it with a few other helm features

PR #260, opened 2023-06-27
Handle IPPolicy CRD state transitions in a safer way

jonstacks:
Contributes to #221 

## What

Since CIDR has to be unique within an IPPolicy: `HTTP 400: IP Policy Rule with CIDR '127.0.0.1/32' already exists [ERR_NGROK_1408]`, we need to calculate the diff and apply the changes in a safe way.

## How

Use the following diffing logic to apply creates, deletes, and updates in the following order:
1. Create all new deny rules that don't exist in the remote with a matching CIDR.
2. Delete any allow rules with matching CIDRs that will be changing to deny rules. Then create the deny rules
3. Delete any deny rules with matching CIDRs that will be changing to allow rules. Then create the allow rules.
4. Create all new allow rules that don't exist in the remote with a matching CIDR.
5. Delete any remaining rules that are not in the spec.
6. Update any rules that exist in the spec and remote but have only different metadata/description.

## Breaking Changes

No.

### DIFF
--- a/internal/controllers/ippolicy_controller.go
+++ b/internal/controllers/ippolicy_controller.go
@@ -179,19 +179,33 @@ func (r *IPPolicyReconciler) createOrUpdateIPPolicyRules(ctx context.Context, po
 	if err != nil {
 		return err
 	}
-	diff := newIPPolicyDiff(remoteRules, policy.Spec.Rules)
-	for _, rule := range diff.needCreate {
-		rule.IPPolicyID = policy.Status.ID
-		_, err := r.IPPolicyRulesClient.Create(ctx, rule)
-		if err != nil {
-			return err
-		}
-	}
+	iter := newIPPolicyDiff(policy.Status.ID, remoteRules, policy.Spec.Rules)
 
-	for _, rule := range diff.needDelete {
-		err := r.IPPolicyRulesClient.Delete(ctx, rule.ID)
-		if err != nil {
-			return err
+	for iter.Next() {
+		for _, d := range iter.NeedsDelete() {
+			r.Log.V(3).Info("Deleting IP Policy Rule", "id", d.ID, "policy.id", policy.Status.ID, "cidr", d.CIDR, "action", d.Action)
+			if err := r.IPPolicyRulesClient.Delete(ctx, d.ID); err != nil {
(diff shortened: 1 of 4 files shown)


internal/controllers/ippolicy_controller.go
```diff
@@ -213,71 +227,181 @@ func (r *IPPolicyReconciler) getRemotePolicyRules(ctx context.Context, policyID
...
+		for cidr, rule := range d.remoteDeny {
+			if !d.existsInSpec(cidr) {
+				d.deletes = append(d.deletes, rule)
+			}
+		}
+		return true
+	case 5: // Update any rules that exist in the spec and remote but have only different metadata/description.
+		for cidr, rule := range d.specAllow {
+			if remoteRule, ok := d.remoteAllow[cidr]; ok {
+				if d.shouldBeUpdated(rule, remoteRule) {
```
bobzilladev, line 338: Maybe `shouldBeUpdated` could be changed to a method that fills in the `updates` itself, so this block doesn't have to be duplicated for allow and deny.

bobzilladev (approved):
A couple minor comments, but overall looks to do what is the desired sequence, which does look like a series safe update transitions.

internal/controllers/ippolicy_controllers_test.go
```diff
@@ -0,0 +1,75 @@
...
+		changedDescriptionRule, // 6. Just changing description
+	}
+
+	diff := newIPPolicyDiff("test", remoteRules, specRules)
+
+	assert.True(t, diff.Next())
+	assert.Empty(t, diff.NeedsDelete())
+	assert.Empty(t, diff.NeedsUpdate())
+	assert.Equal(t, []*ngrok.IPPolicyRuleCreate{
+		{IPPolicyID: "test", CIDR: "10.0.0.0/8", Action: pointer.String(IPPolicyRuleActionDeny)}},
```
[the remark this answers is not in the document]
jonstacks: Fixed in efa7b2e

internal/controllers/ippolicy_controller.go
jonstacks: Fixed in efa7b2e

PR #261, opened 2023-06-28
Improve logging for httpsedge controller

jonstacks:
## What

Improve logging for the `httpsedge` controller.

## How

* Clean up duplicated log values by by making a new logger using `WithValues`
* Pass loggers through the code using contexts
* Change some of the plain `.Info` logs to log at different verbosities. Ex: `log.V(1).Info` to effectively log at the `debug` level.

## Breaking Changes

No

### DIFF
--- a/internal/controllers/httpsedge_controller.go
+++ b/internal/controllers/httpsedge_controller.go
@@ -47,6 +47,14 @@ import (
 	"github.com/ngrok/ngrok-api-go/v5/backends/tunnel_group"
 )
 
+type routeModuleComparision string
+
+const (
+	routeModuleComparisonBothNil        routeModuleComparision = "both nil"
+	routeModuleComparisonBothNilOrEmpty routeModuleComparision = "both nil or empty"
+	routeModuleComparisonDeepEqual      routeModuleComparision = "deep equal"
+)
+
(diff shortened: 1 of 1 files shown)


internal/controllers/httpsedge_controller.go
```diff
@@ -871,10 +940,11 @@ func (u *edgeRouteModuleUpdater) setEdgeRouteWebhookVerification(ctx context.Con
 	}
 
 	if reflect.DeepEqual(&module, route.WebhookVerification) {
-		u.log.Info("Webhook Verification matches desired state, skipping update")
+		log.V(1).Info("Webhook Verification matches desired state, skipping update")
```
deroine, line 943: Nit: worth differentiating the DeepEquals messages from the nil?

deroine (approved):
One minor thing, but otherwise looks ok to me.

internal/controllers/httpsedge_controller.go
jonstacks: Good call. I can add those in where appropriate.

jonstacks: Should be fixed in d5d7d57. Let me know what you think

deroine (approved):

PR #262, opened 2023-06-28
fix: for changing pathType between 'Exact' and 'Prefix'

jonstacks:
Contributes to #221 

## What

Updates the logic for matching routes in the spec to existing routes.

## How

Since ngrok routes are unique by Match(Path in the Ingress Spec), we need to remove the check on MatchType(PathType), so they can be updated.

## Breaking Changes

No


deroine (approved):
Makes sense to me. 👍🏼

PR #276, opened 2023-07-03
fix: Don't remove httpsedges/tunnels that weren't created by the ingr…

### LINKED ISSUE #267: CustomResources created outside of the Ingress Controller are deleted
### What happened

I manually created a `tunnel.ingress.k8s.ngrok.com` resource with custom labels that I wanted running:

```yaml
kind: Tunnel
metadata:
  name: starwars-tunnel
  namespace: default
spec:
  forwardsTo: starwars.default.svc.cluster.local:23
  labels:
    k8s.ngrok.com/namespace: default
    k8s.ngrok.com/service: starwars
    k8s.ngrok.com/port: "23"
```

After a while, the tunnel was no longer present. During the driver's sync, it was calculated as being able to be deleted [here](https://github.com/ngrok/kubernetes-ingress-controller/blob/main/internal/store/driver.go#L285-L288)

The Tunnel should not have been deleted by the Ingress Controller. Since it wasn't created from an Ingress and was manually created, it should have remained intact until someone deleted it.

### How to reproduce

1. Create a `tunnel.ingress.k8s.ngrok.com` manually like the one specified above. 
2. Verify the Tunnel exists: `kubectl get tunnels.ingress.k8s.ngrok.com` . You should see the tunnel you just created.
3. Trigger a sync through one of the following methods and wait a few minutes
  a. Edit the annotations of an ingress
  b. Do a `rollout restart` of the ingress controller deployment.
4. Verify the Tunnel is gone: `kubectl get tunnels.ingress.k8s.ngrok.com`

jonstacks:
Fixes #267 

## What

Fixes Tunnels and HTTPSEdges that weren't created by the ingress controller getting deleted. 

## How

If an Edge or Tunnel are created by the ingress controller to fulfill an Ingress, add an annotation that they are managed by the 
Ingress Controller. When we look to delete dangling tunnels and edges, only do so if they were managed by the controller.

## Testing

Logs:

```
2023-07-03T21:21:10Z	LEVEL(-3)	cache-store-driver	Not deleting tunnel because it is not managed by ingress controller	{"V1Alpha1Tunnel": "default/test-app-3-80"}
2023-07-03T21:21:10Z	LEVEL(-3)	cache-store-driver	Not deleting tunnel because it is not managed by ingress controller	{"V1Alpha1Tunnel": "default/starwars-tunnel"}
```

## Breaking Changes
No.

### DIFF
--- a/internal/store/driver.go
+++ b/internal/store/driver.go
@@ -188,7 +188,7 @@ func (d *Driver) Sync(ctx context.Context, c client.Client) error {
 		for _, currDomain := range currDomains.Items {
 			if desiredDomain.Name == currDomain.Name && desiredDomain.Namespace == currDomain.Namespace {
 				// It matches so lets update it if anything is different
-				// or if it doesn't have a an annotation that its managed by
+				// or if it doesn't have an annotation that it is managed by
 				// the ingress controller.
 				if !reflect.DeepEqual(desiredDomain.Spec, currDomain.Spec) || !hasControllerManagedAnnotation(&currDomain) {
 					currDomain.Spec = desiredDomain.Spec
--- a/internal/store/driver.go
+++ b/internal/store/driver.go
@@ -184,11 +184,15 @@ func (d *Driver) Sync(ctx context.Context, c client.Client) error {
 
 	for _, desiredDomain := range desiredDomains {
 		found := false
+
 		for _, currDomain := range currDomains.Items {
 			if desiredDomain.Name == currDomain.Name && desiredDomain.Namespace == currDomain.Namespace {
 				// It matches so lets update it if anything is different
-				if !reflect.DeepEqual(desiredDomain.Spec, currDomain.Spec) {
+				// or if it doesn't have a an annotation that its managed by
+				// the ingress controller.
+				if !reflect.DeepEqual(desiredDomain.Spec, currDomain.Spec) || !hasControllerManagedAnnotation(&currDomain) {
 					currDomain.Spec = desiredDomain.Spec
+					currDomain.Annotations = desiredDomain.Annotations
 					if err := c.Update(ctx, &currDomain); err != nil {
 						d.log.Error(err, "error updating domain", "domain", desiredDomain)
 						return err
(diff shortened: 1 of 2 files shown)


Abdiramen (approved):

internal/store/driver.go
```diff
@@ -251,8 +264,9 @@ func (d *Driver) Sync(ctx context.Context, c client.Client) error {
 		for _, currTunnel := range currTunnels.Items {
 			if desiredTunnel.Name == currTunnel.Name && desiredTunnel.Namespace == currTunnel.Namespace {
 				// It matches so lets update it if anything is different
-				if !reflect.DeepEqual(desiredTunnel.Spec, currTunnel.Spec) {
+				if !reflect.DeepEqual(desiredTunnel.Spec, currTunnel.Spec) || !hasControllerManagedAnnotation(&currTunnel) {
```
OfTheDelmer, line 267: always curious when I see DeMorgan situations like this if `!(A & B)` vs `!A || !B` feels more readable.

internal/store/driver.go
```diff
@@ -184,11 +184,15 @@ func (d *Driver) Sync(ctx context.Context, c client.Client) error {
 
 	for _, desiredDomain := range desiredDomains {
 		found := false
+
 		for _, currDomain := range currDomains.Items {
 			if desiredDomain.Name == currDomain.Name && desiredDomain.Namespace == currDomain.Namespace {
 				// It matches so lets update it if anything is different
-				if !reflect.DeepEqual(desiredDomain.Spec, currDomain.Spec) {
+				// or if it doesn't have a an annotation that its managed by
```
OfTheDelmer, line 191:
```suggestion
				// or if it doesn't have an annotation that it is managed by
```

nikolay-ngrok (changes requested):
In its current form, this will adopt any manually created resources (tunnels/domains/etc) and will still delete them after we delete the ingress that adopted them. I think we should only mark resources as owned by the ingress controller when it created them. After discussing with @jonstacks , we decided to explore our options a bit more before committing.

internal/store/driver.go
deroine: We historically haven't adopted a style guides around these types of preferences, if we feel strongly enough I would recommend spinning up a conversation or doc outside of this PR. 

I very much appreciate anyone who knows DeMorgan's law and formal logic in general 😄

jonstacks:
@nikolay-ngrok, I'm going to do a write up on the linked issue to capture what we talked about and how we will handle it.

jonstacks:
Closing this in favor of an alternate approach discussed on the related issue.

PR #281, opened 2023-07-10
Create tunnels per namespace/service

nikolay-ngrok:
## What
Its moving from globally unique tunnels (e.g. combo of "service-port") to ones that are only unique in their respective namespace.

## How
Don't merge based on `service-port`, but include the namespace too.

## Breaking Changes
None


CK-Ward:
@nikolay-ngrok Thanks Nikolay — does this address some of what's called out in #267? If so, can we please call that out descriptively so stakeholders can track the progress? 🙏

nikolay-ngrok:
@CK-Ward I don't think its addressing anything specifically, more like an issue I noticed.

jonstacks (approved):

PR #285, opened 2023-07-14
Ingress controller owns tunnels it creates and doesn't delete things it doesn't own.

nikolay-ngrok:
related: #267

## What
Presently, when the ingress controller reconciles state, any tunnels that are not desired (e.g. there is no ingress object that requires one) get deleted. This means that we cannot reliably use the ingress controller to just wire up a manually created tunnel, because it might get deleted. This also means that we cannot use 2 ingress controllers at the same time, since they end up deleting each other's tunnels (unless the set of services they target are the same).

Once this PR merges, tunnels are created with labels that show which ingress controller created them. The ingress controller then only looks at reconciling state for tunnels it labeled, and never touches custom created ones (and the ones created by a different ingress controller), even if they target the same service. 

## How
 * When creating a tunnel CR:
   * Use `GenerateName` of `{service}-{port}-` to get a randomly suffixed name
   * Add labels to the tunnel CRs that include:
```
k8s.ngrok.com/controller-namespace
k8s.ngrok.com/controller-name
k8s.ngrok.com/service
k8s.ngrok.com/port
```
   * Add ownership references to all ingress objects
 * When looking for already existing tunnels (while reconciling), filter by matching labels for `controller-namespace` and `controller-name`.
   * create/update/delete only tunnels in this set

## Breaking Changes
In its current form, this PR will abandon every already existing tunnel and these will need to be deleted manually. So far I haven't found a proper, reliable way of migrating these (that works across controller upgrades, multiple controllers, restricted namespaces and over time). The operators could hit a limit on the number of tunnels, but once the old tunnels get deleted, this will fix itself.

## Followup work
 * use a similar solution for edges (but without ownership references)
 * currently all tunnel CRs are "tunneled" by each ingress controller, but we should probably restrict this to own or unlabeled ones? or maybe we need a labeling to say which goes where?

### DIFF
--- a/helm/ingress-controller/templates/controller-deployment.yaml
+++ b/helm/ingress-controller/templates/controller-deployment.yaml
@@ -81,6 +81,7 @@ spec:
         - --health-probe-bind-address=:8081
         - --metrics-bind-address=:8080
         - --election-id={{ include "kubernetes-ingress-controller.fullname" . }}-leader
+        - --manager-name={{ include "kubernetes-ingress-controller.fullname" . }}-manager
         securityContext:
           allowPrivilegeEscalation: false
         env:
@@ -94,8 +95,6 @@ spec:
             secretKeyRef:
               key: AUTHTOKEN
               name: {{ include "kubernetes-ingress-controller.credentialsSecretName" . }}
-        - name: DEPLOYMENT_NAME
-          value: {{ include "kubernetes-ingress-controller.fullname" . }}-manager
         - name: POD_NAMESPACE
           valueFrom:
             fieldRef:
--- a/helm/ingress-controller/templates/controller-deployment.yaml
+++ b/helm/ingress-controller/templates/controller-deployment.yaml
@@ -94,6 +94,8 @@ spec:
             secretKeyRef:
               key: AUTHTOKEN
               name: {{ include "kubernetes-ingress-controller.credentialsSecretName" . }}
+        - name: DEPLOYMENT_NAME
+          value: {{ include "kubernetes-ingress-controller.fullname" . }}-manager
         - name: POD_NAMESPACE
           valueFrom:
             fieldRef:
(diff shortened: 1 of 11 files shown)


helm/ingress-controller/templates/controller-deployment.yaml
```diff
@@ -94,6 +94,8 @@ spec:
             secretKeyRef:
               key: AUTHTOKEN
               name: {{ include "kubernetes-ingress-controller.credentialsSecretName" . }}
+        - name: DEPLOYMENT_NAME
```
nikolay-ngrok, line 97: should this be a flag instead? 🤔

nikolay-ngrok: simplified to use a flag

nikolay-ngrok:
One way to provide an "automated" way of deleting the old tunnels is to check for presence of a certain config map and ask the users to create it. On startup, if we see it there, we can remove all tunnels, then delete the map. This way, we can aid with removing the unmanaged, but still created by the ingress controller tunnels.

bobzilladev (approved):
All looks logical to me.

Went down a path looking into when UID's change, found [this discussion](https://stackoverflow.com/questions/51068026/when-exactly-do-i-set-an-ownerreferences-controller-field-to-true) talking about using the `controller` field to stop multiple controllers fighting over adopting the same resource. Would only be a problem if running multiple ingress controllers with the same namespace+name, so probably not a situation that needs to be dealt with here, but thought it was interesting, TIL. Maybe if someone one day wants a blue/green deploy with the same name for some reason.

jrobsonchase (approved):
:+1: Makes sense to me!

The old tunnel object cleanup doesn't feel great to me, but I don't have any better suggestions. If we want to provide users with a quick "delete *ALL* the tunnels!" one-liner in the release notes, `kubectl delete tunnel --selector '!k8s.ngrok.com/controller-name'` should work.

Issue #287, opened 2023-07-27
Changes applied to `NgrokModuleSet` do not propagate to managed `Edge`


pmalek:
### What happened

After applying `NgrokModuleSet` with e.g. oauth set to google email address and changing the email address there's no change in the managed `Edge`.

### What you think should happen instead

After applying `NgrokModuleSet` with e.g. oauth set to google email address and changing the email address, managed `Edge` should be changed

### How to reproduce

1. Apply (e.g. via `kubectl apply ...`) below listed manifest (substitute domain and email address with your own)

    ```yaml
    apiVersion: v1
    kind: Service
    metadata:
      name: proxy
      namespace: infra
    spec:
      ports:
      - name: http
        port: 80
        targetPort: 80
      selector:
        app: httpbin
    ---
    apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: httpbin
      namespace: infra
    spec:
      replicas: 1
      selector:
        matchLabels:
          app: httpbin
      template:
        metadata:
          labels:
            app: httpbin
        spec:
          containers:
          - image: kong/httpbin:0.1.0
            name: httpbin
            ports:
            - containerPort: 80
    ---
    apiVersion: networking.k8s.io/v1
    kind: Ingress
    metadata:
      name: proxy
      namespace: infra
      annotations:
        k8s.ngrok.com/modules: oauth
    spec:
      ingressClassName: ngrok
      rules:
      - host: <DOMAIN>.ngrok-free.app
        http:
          paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: proxy
                port:
                  number: 80
    ---
    kind: NgrokModuleSet
    apiVersion: ingress.k8s.ngrok.com/v1alpha1
    metadata:
      name: oauth
      namespace: infra
    modules:
      oauth:
        google:
          emailAddresses:
          - <EMAIL>
    ```

1. Verify that you can access `<DOMAIN>` using `<EMAIL>` address via google authentication.
1. Change `<EMAIL>` to `<EMAIL2>` and apply the changes (e.g. `kubectl apply ...`)
1. Verify that you can still access `<DOMAIN>` using `<EMAIL>` address via google authentication and not via `<EMAIL2>`

nikolay-ngrok:
[...]

PR #290, opened 2023-07-31
Reconciling ngrok-module-set changes

nikolay-ngrok:
closes: #287

## What
Currently, the controller watches for changes in `NgrokModuleSet` CRs (and persists the updates), but it is not actively triggering reconciliation when they change. This means that changes only in modules are not applied to the `HTTPSEdge` CRs when the `NgrokModuleSet` CRs referenced by an `Ingress` get changed.

Once this PR merges, both `Ingress` and `NgrokModuleSet` will trigger reconciliation and will keep dependent objects up to date.

## How
 * Add another controller that owns and watches `NgrokModuleSet`s
 * Trigger partial reconciliation, since only `HTTPSEdges` are affected by modules
 * Make sure we are not doing too many changes, since both `Sync` methods might run concurrently

## Breaking Changes
None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test ./... -coverprofile cover.out -timeout 20s
 
 ##@ Build
 
--- a/internal/controllers/ingress_controller.go
+++ b/internal/controllers/ingress_controller.go
@@ -14,7 +14,6 @@ import (
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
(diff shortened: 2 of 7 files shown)


nikolay-ngrok (commented):
I think this is mostly there. The only part left to be completed is the concurrency story.

Currently, we block and de-bounce calls to `Sync` in the driver. This is done because:
 * we were seeing multiple calls to the controller. It wasn't clear why, since as indicated by [MaxConcurrentReconciles](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controller.go#L40) this shouldn't happen. However, we were not setting this explicitly, and the controller will resolve it value by type, which in this case is `Ingress` and was largely dictated by the cluster config.
 * previously `Sync` would talk to the ngrok API. This is not the case anymore, it will just convert a bunch of `Ingress` objects to a set of `Domain/HTTPSEdge/Tunnel` objects, without ever talking to ngrok, so no rate limits will be actually hit.

However, we now have to concurrent calls happening at the same time - `Sync` and `SyncEdges`, driven by the corresponding controllers, which we need to take into account. Because the above conditions are no longer true, we could be more lenient in how we wanna fix the concurrency. We need to guarantee only one of these is proceeding or they will be racing over different versions of the objects and clutter too much. In my mind, we have the following options:
 * straight up single lock/unlock, e.g. allow only one of them to proceed. The other reconciliation loops will wait for a bit but eventually run and bring the end state to consistency. However, some amount of duplicate work will happen.
 * batch the calls, where we run only first and last one (what is attempted here). In this way any intermediate attempts will be counted "successful" and the last one will just bring the state to consistency.

Long term, I think we should tweak the sync cycle a bit further by queuing changes and sync. What I mean is instead of update and sync cycle, we should be explicit about store evolution, and not let random changes happen while we are syncing.

nikolay-ngrok:
@jrobsonchase this is ready for rereview, the CI is broken from a previous merge (which I'll look at merging before this)

jrobsonchase (approved):
Looks good to me!

I probably read over it three or four times while you were out and kept getting pulled away before managing to push it over the line :sweat_smile:

PR #292, opened 2023-08-18
Check for secret's env vars in 'make deploy'

euank:
I keep running 'make deploy' without having these set, and then have to remember to set them.

It seems nicer to the developer to error out early.


jrobsonchase (approved):
Big "same" :shipit:

PR #293, opened 2023-08-18
Add an ingress uid label to tunnels

### LINKED ISSUE #291: Tunnel mix-up with multiple ngrok Instances
### What happened

We use the ngrok ingress controller to expose a local kind cluster endpoint (Kubernetes in Docker) for connecting with an external service. Every developer has their own local setup with an ngrok ingress controller running. Things work fine with just one ngrok controller, but when there are two, we notice the routing mix-up. My domain's traffic ends up at my coworker's station.

I've followed the setup guide on this GitHub repo and not sure if I missed something or if it's a controller issue.

Can sharing logs or more info help fix this?

Traffic from my ngrok domain should route to my tunnel. 

### How to reproduce

_No response_

jrobsonchase:
Resolves #291

## What

Add a `k8s.ngrok.com/ingress-uid` label to tunnels created by the
ingress controller to uniquely identify them among multiple ingress
controller instances.

## How

Plumb the full ingress object down to the method to create ngrok tunnel
lables, and add the UID from that. The ingress UID should be stable as
long as the object exists in k8s (citation needed?), so we don't have to
worry about unexpected recreate thrashing as a result of it changing.

## Breaking Changes

Shouldn't be any. After the ingress controller is redeployed, a
difference should be detected between the desired and actual tunnel
state, and things should be recreated with the new labels as expected.

### DIFF
--- a/helm/ingress-controller/Chart.lock
+++ b/helm/ingress-controller/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 2.6.0
-digest: sha256:6657a16696e9c132a98eb257951876f491f213b55aac7b4894993d260b2bae09
-generated: "2023-07-13T14:02:39.574375-04:00"
+  version: 2.8.0
+digest: sha256:83dc5879aedee09c5d5d1865ed9e861feeeea6df10e56c41787911b54e141af9
+generated: "2023-08-18T09:47:25.793815648-04:00"
--- a/internal/controllers/base_controller.go
+++ b/internal/controllers/base_controller.go
@@ -73,7 +73,7 @@ func (r *baseController[T]) reconcile(ctx context.Context, req ctrl.Request, cr 
 		}
 	} else {
 		if hasFinalizer(cr) {
-			if r.statusID == nil || r.statusID(cr) != "" {
+			if r.statusID != nil && r.statusID(cr) != "" {
 				sid := r.statusID(cr)
 				r.Recorder.Event(cr, v1.EventTypeNormal, "Deleting", fmt.Sprintf("Deleting %s: %s", r.kubeType, crName))
 				if err := r.delete(ctx, cr); err != nil {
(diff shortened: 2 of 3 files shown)


jrobsonchase:
Current dependencies on/for this PR:
* main
  * **PR #293** <a href="https://app.graphite.dev/github/pr/ngrok/kubernetes-ingress-controller/293" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ngrok/kubernetes-ingress-controller/293?utm_source=stack-comment).

bobzilladev (approved):
Hopefully `r.statusID` wasn't actually ever nil there.

euank (commented):
Just so we're aware, this does have a functional difference that could matter for limits.

To explain what I mean, let's look at the following scenario:
You have one service (`my-app`) listening on port 80, and you have 5 ingresses for it (`ing1.myapp.ngrok.io` through `ing5.myapp.ngrok.io`).
Before, that would result in 1 tunnel with the labels `namespace: default, service: my-app, port: 80`).
After this change, since there's N different ingresses, it's now 5 tunnels, which could get you closer to your session tunnel limit.

An alternative to this would be using the namespace UID, or using the service UID, both of which would be shared in more cases.
I feel like the service UID matches most closely with the existing intent of the labels.

All that said, I think this is probably fine, and like longer term we probably need to rethink this a bit regardless (i.e. the coupling of service and tunnel is a slightly off abstraction, and really we need to be pulling the EdgeID out of the tunnel header we get sent by the server).
I think the service UID probably fits a little better, but I could also be missing something there

jrobsonchase:
The ingress UID maps most cleanly to our current tunnel ownership model. We're also setting the tunnel owner as the ingress, so I don't think it was ever the intended behavior for a tunnel to be usable by multiple ingresses that just so happened to have the same labels.

@bobzilladev Well... it was, and I was getting panics :sweat_smile: Didn't dig much deeper into the underlying cause though.

nikolay-ngrok:
This actually breaks the case where we have multiple ingresses in the same namespaces, defining routes leading to the same service. This is because we are creating multiple https edge routes, but a single tunnel, so the last one that marks the tunnel wins, leaving the rest unroutable.

PR #294, opened 2023-08-21
*: reflect.DeepEqual -> maps/slices.Equal

euank:
We have generics, we can do faster and better equality comparing in a few places.

These should be functionality identical. This is meant to be a rote refactor that the type-system covers, and no additional validation beyond that was done.


nikolay-ngrok (approved):
lgtm

PR #297, opened 2023-08-30
do no check env in CI

nikolay-ngrok:
## What
We introduced a change that requires `NGROK_AUTHTOKEN` and `NGROK_API_KEY` vars to be set on deploy, which is useful when doing manual deploys. However, in CI we use the same, but just to see if we can deploy successfully, without actually using the values for these vars. In this case, the ci build fails, since it checks these vars as part of the `make deploy` run.

## How
Split `deploy` cmd/script to two parts, and use the one that doesn't check in CI.

## Breaking Changes
None

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -140,7 +140,7 @@ jobs:
       run: make docker-build
 
     - name: Deploy controller to local cluster
-      run: make deploy
+      run: make _deploy # do not check env vars
 
     - name: Check if controller is up
       run: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -140,7 +140,7 @@ jobs:
       run: make docker-build
 
     - name: Deploy controller to local cluster
-      run: make _deploy # do not check env vars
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -140,7 +140,7 @@ jobs:
       run: make docker-build
 
     - name: Deploy controller to local cluster
-      run: make deploy
+      run: make _deploy # do not check env vars
```
euank, line 143: One alternative would be swapping this to `make deploy NGROK_API_KEY=fake-ci-check NGROK_AUTHTOKEN=fake-ci-check`

I think that might feel a little better since it lets us not have a ci-specific make target

euank (approved):
This seems reasonable to me.

I also suggested a different change that could work too.

Apologies for missing this in my change!

.github/workflows/ci.yaml
nikolay-ngrok: Ah, I was considering just adding the env vars on the actions job, but this feels a little better. I can do that.

PR #304, opened 2023-09-11
Remove port portion from destination to present FQDN as SNI

nikolay-ngrok:
## What
When forwarding TLS connection, we are setting the `ServerName`/SNI to the target. According to RFC 6066:
```
"HostName" contains the fully qualified DNS hostname of the server,
   as understood by the client.  The hostname is represented as a byte
   string using ASCII encoding without a trailing dot.  This allows the
   support of internationalized domain names through the use of A-labels
   defined in [RFC5890].  DNS hostnames are case-insensitive.  The
   algorithm to compare hostnames is described in [RFC5890], Section
   2.3.2.4.
```
However, our target is in the form `cluster-service-name:port`, which fails validation in some server libraries, as highlighted by https://github.com/libressl/portable/issues/660 . We should instead just set `cluster-service-name` as an SNI.

## How
Remove port portion of the address before setting `ServerName` in the tunnel.

## Breaking Changes
None

### DIFF
--- a/pkg/tunneldriver/driver.go
+++ b/pkg/tunneldriver/driver.go
@@ -225,7 +225,7 @@ func handleConn(ctx context.Context, dest string, protocol string, dialer Dialer
 	if protocol == "HTTPS" {
 		host, _, err := net.SplitHostPort(dest)
 		if err != nil {
-			return err
+			host = dest
 		}
 		next = tls.Client(next, &tls.Config{
 			ServerName:         host,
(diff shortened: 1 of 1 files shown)


pkg/tunneldriver/driver.go
```diff
@@ -223,8 +223,12 @@ func handleConn(ctx context.Context, dest string, protocol string, dialer Dialer
 
 	// Support HTTPS backends
 	if protocol == "HTTPS" {
+		host, _, err := net.SplitHostPort(dest)
+		if err != nil {
+			return err
+		}
```
jrobsonchase, line 229: This will return an error if no port is found: https://go.dev/play/p/vSHWL0VeJxV

Probably better to do:
```suggestion
		host, _, err := net.SplitHostPort(dest)
		if err != nil {
			host = dest
		}
```

jrobsonchase (approved):

PR #339, opened 2024-02-10
Load cluster domain as a config option. Fall back to default if not set

fr6nco:
…vc.cluster.local if not set

## What
Cluster domain is hardcoded. Load the cluster domain from an environment variable. If the environment variable is not set, fall back to the default svc.cluster.local

Currently very simple, just please take a look if this is something you would accept. 
Test were not written, I could not find anything covering the cluster domain, so it should be added. 
Helm chart to be updated, where we could set the cluster domain to a configmap or simply passing it as an extraEnvVar: {}

## How
Cluster domain is read from env variable. 

## Breaking Changes ##
no


salilsub:
Hi @fr6nco . For cases like this we'd want to use a Helm chart override instead of using an environment variable. If you modified this PR/submitted a new PR using this approach, we'd accept that.

fr6nco:
I bit more commints than expected. It was reworked, where the option is passed down as a config option. No longer via an env variable. The option is exposed via the helm chart.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.11.0
+
+- Addded support for cluster domain config option. Defaults to svc.cluster.local. [#339](https://github.com/ngrok/kubernetes-ingress-controller/pull/339)
+
```
CK-Ward, line 11: Hi @fr6nco, thanks for making a PR for this. Will you please remove this change to release a new version? We'll take care of that as maintainers — thanks 🙏

CK-Ward (changes requested):

fr6nco (commented):
Changelog changes removed as requested.

jonstacks (approved):

cmd/main.go
```diff
@@ -122,6 +123,7 @@ func cmd() *cobra.Command {
 	c.Flags().StringVar(&opts.watchNamespace, "watch-namespace", "", "Namespace to watch for Kubernetes resources. Defaults to all namespaces.")
 	c.Flags().StringVar(&opts.managerName, "manager-name", "ngrok-ingress-controller-manager", "Manager name to identify unique ngrok ingress controller instances")
 	c.Flags().BoolVar(&opts.useExperimentalGatewayAPI, "use-experimental-gateway-api", false, "sets up experemental gatewayAPI")
+	c.Flags().StringVar(&opts.clusterDomain, "cluster-domain", "svc.cluster.local", "Cluster domain used in the cluster")
```
alex-bezek, line 126: My only nit for this is this default seems to be declared in a few places and some of the defaults might be overkill. 
The helm values has a default of `clusterDomain: svc.cluster.local`

The template checks if it has a value before passing it, but would only be false if it was overridden to "" via
```
        {{- if .Values.clusterDomain }}
        - --cluster-domain={{ .Values.clusterDomain }}
        {{- end }}
```

But if this isn't passed it then gets the default from this flag. 

Lastly the driver then also has the `const defaultClusterDomain = "svc.cluster.local"` which is then overridden in this main.go via `WithClusterDomain`

It's not the end of the world, and I don't think the system would work if it was set to empty string or something, but it just seemed like a lot of layers to grok.

alex-bezek (approved):

cmd/main.go
jonstacks: Part of this is my fault when merging in main and noticing we have quite a few named arguments to `NewDriver`. I did a small functional options pattern refactor but still feel like having it default in the driver is reasonable. Since we also have it as a command line flag for now, I'm also okay with having it there so you can easily see the default when running `go run ./cmd/main.go --help`

jonstacks:
Thank you for the contribution, @fr6nco!

Issue #340, opened 2024-02-14
 CNAME Target not being set on Ingress Status


ppapp92:
### Kubernetes Version

1.28

### Helm Chart Version

0.12.1

### Ingress Controller Version

0.10.1

### System Info

AWS EKS
ArgoCD
ExternalDNS

### Helm Chart configuration

```
kubernetes-ingress-controller:
  replicaCount: 1
  ingressClass:
    name: ngrok
    create: true
    default: false
  controllerName: "k8s.ngrok.com/ingress-controller"
  credentials:
    apiKey: <API_KEY>
    authtoken: <AUTHTOKEN>
  region: ""
```

### What happened

When an Ngrok class ingress resource is created in K8 the Address(status.loadBalancer) doesn't get updated by the Ngrok Ingress Controller. This means that ExternalDNS does not create appropriate route53 entires to fully automate the deployment of an application.

Similar to https://github.com/ngrok/kubernetes-ingress-controller/issues/161

### What you think should happen instead

Ngrok Ingress Controller should update the ingress resource with the CNAME Target so that ExternalDNS can properly configure DNS for the newly created service/app.

### How to reproduce

Create an ingress resource in K8
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  finalizers:
  - k8s.ngrok.com/finalizer
  generation: 1
  labels:
    app.kubernetes.io/component: server
    app.kubernetes.io/instance: argo-workflows
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: argo-workflows-server
    app.kubernetes.io/part-of: argo-workflows
    helm.sh/chart: argo-workflows-0.40.4
  name: argo-workflows-server
  namespace: argo
spec:
  ingressClassName: ngrok
  rules:
  - host: domain.example.com
    http:
      paths:
      - backend:
          service:
            name: argo-workflows-server
            port:
              number: 2746
        path: /
        pathType: Prefix
```
View the status of the ingress and see that the loadBalancer address is empty
```
status:
  loadBalancer: {}
```

Both the Tunnel and Domain resources get created.

```
apiVersion: ingress.k8s.ngrok.com/v1alpha1
kind: Domain
metadata:
  finalizers:
  - k8s.ngrok.com/finalizer
  generation: 1
  name: example
  namespace: argo
spec:
  description: Created by kubernetes-ingress-controller
  domain: domain.example.com
[...]

Issue #149, opened 2023-02-14
Synchronize ngrok resources in background process


alex-bezek:
### Description

We created this _store_ concept that houses all the gathered resource information, and then a Driver has a sync function which calculates the total differences, syncs those to the ngrok api, and updates the k8s api with statuses. 

We started initially by just plugging this into our current reconcile loop, however we quickly found that the number of reconcile events getting re-triggered would cause sync to be called many times in a row and hit rate limits with the ngrok api. To fix this temporarily, we added a thread safe reentrance lock https://github.com/ngrok/kubernetes-ingress-controller/blob/main/internal/store/driver.go#L123-L134

Instead of doing this, we can just trigger this sync operation on some interval in a background process. This will have the bonus effect of continually reconverging any changes made directly in the ngrok account back to what matches k8s. 

### Use Case

a simpler sync loop to think about and continual drift correction

alex-bezek:
> This will have the bonus effect of continually reconverging any changes made directly in the ngrok account back to what matches k8s.

This is not actually true. While the store sync may run in a loop, it just ensures the CRDs in the k8s api are consistent with the calculated values from the store. Those CRD controllers would actually have to continually reconcile in a loop to achieve any drift detection and reconciliation in the ngrok api.

alex-bezek:
One small issue that this would solve is that right now while we watch our CRDs for changes and push them into the store via https://github.com/ngrok/kubernetes-ingress-controller/blob/main/internal/store/updatestorehandler.go it doesn't trigger a re-sync. syncing in the background would automatically get CRD updates eventually, or we could just call sync in all these functions.

PR #342, opened 2024-02-15
domain controller: run a driver sync after updating

jrobsonchase:
closes: #340

## What

Ensures that the `Status.LoadBalancer.Ingress[].Hostname` field is set after the target CNAME is resolved on the `Domain`.

## How

Add a call to `Driver.updateIngressStatuses` in the `UpdateStoreHandler.Update` method.

## Breaking Changes

None

### DIFF
--- a/internal/store/updatestorehandler.go
+++ b/internal/store/updatestorehandler.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 )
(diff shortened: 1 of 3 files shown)


alex-bezek (approved):

jonstacks (commented):
It's been a while since I looked at this, but I don't think we want to share the driver outside of the specific ingress controller. The ingress controller is already watching Domains among other objects:

https://github.com/ngrok/kubernetes-ingress-controller/blob/69867c300c80c07831d8176688c96db99312843d/internal/controller/ingress/ingress_controller.go#L32-L48

I think the issue is that when it watches these, it is just updating the store with the updates, but that isn't actually triggering an resync on any resources. I think this may actually be a more generic class of problem that we should handle.

It also feels a bit weird to be sharing state like this between controllers IMO.

jrobsonchase:
Is the driver Ingress-specific? I already saw it being shared by the ingress controller and moduleset controller, so I figured why not?

> The ingress controller is already watching Domains among other objects:
> I think the issue is that when it watches these, it is just updating the store with the updates, but that isn't actually triggering an resync on any resources. I think this may actually be a more generic class of problem that we should handle.

Yeah, this was the first path I tried to go down and didn't really get anywhere with. That `Watches` for all of the other resources didn't seem to do anything that I could tell. I expected that the ingress controller would get notifications when any of those resources was updated, but that didn't seem to be the case.

internal/store/updatestorehandler.go
```diff
@@ -44,6 +49,10 @@ func (e *UpdateStoreHandler) Update(ctx context.Context, evt event.UpdateEvent,
 		e.log.Error(err, "error updating object in update", "object", evt.ObjectNew)
 		return
 	}
+	if err := e.driver.updateIngressStatuses(ctx, e.client); err != nil {
```
jonstacks, line 52: @alex-bezek, we have a debouncer on the sync right? rather than passing the client through here, what do you think if we called `driver.Sync` after creates/updates/deletes?

jonstacks (approved):
:+1: I think this will fix the original issue. I think we should still come back and possibly remove the `store.Driver` from the `ModuleSetReconciler` and not sync things from inside it once we figure out a better way to do re-syncs when the store is updated inside the `Ingressreconciler`

internal/store/updatestorehandler.go
alex-bezek: yeah, the debouncer should keep it from overwhelming things and will batch changes. We talked about actually just calling sync on every one of these handlers in here https://github.com/ngrok/kubernetes-ingress-controller/issues/149#issuecomment-1460670311 
I'm not sure if we still have the bug where creating the ingress before the moduleset doesn't re-trigger a sync, but I think we were thinking about how to solve that at the time.

jrobsonchase: I was originally calling `driver.Sync`, but figured it might be better to scope it down to do less work. Happy to change it back if that would be better!

jrobsonchase:
Oops, did that let me merge before checks passed?

PR #344, opened 2024-02-16
Release controller 0.10.2 and helm chart 0.12.2

jrobsonchase:


Abdiramen (approved):
Lets goooooooo!

CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.10.2
+
+### Added
```
jonstacks, line 10: To follow semver, I think we should bump this to `0.11.0` since it introduces new features. 

Also, can we link the PRs that introduced the additions/fixes like we have been doing?

jonstacks (changes requested):

CHANGELOG.md
jrobsonchase: The semver rules only apply to [major version > 0](https://semver.org/#spec-item-7). Before 1.0, [Anything MAY change at any time.](https://semver.org/#spec-item-4) I'm used to ["cargo flavored semver"](https://doc.rust-lang.org/cargo/reference/semver.html#change-categories), which uses the minor version for pre-1.0 breaking changes and the patch version for both additions and fixes. Not sure how helm interprets the semver spec, but strictly speaking, 0.10.2 is just as compatible/incompatible with 0.10.1 as 0.11.0 per the spec.

I'll add PR links :+1:

jonstacks (approved):

CHANGELOG.md
jonstacks: That is just what we tried to follow in the past. Agree with pre-1.0 having no guarantees. Approved so its not blocking and thanks for adding the PR links :+1:

jrobsonchase:
Added a note to `docs/developer-guide/releasing.md` on the pre-1.0 version policy that uses "y" for breaking changes and "z" for features+fixes in "0.y.z" versions.

PR #356, opened 2024-03-27
search for service in httproute namespace

Abdiramen:
## What
Fix back-end namespace bug

## How
Set backendref namespace to httproute namespace

### DIFF
--- a/internal/store/driver.go
+++ b/internal/store/driver.go
@@ -947,7 +947,7 @@ func (d *Driver) calculateHTTPSEdgesFromGateway(edgeMap map[string]ingressv1alph
 
 								refName := string(backendref.Name)
 								//refNamespace := string(*backendref.Namespace)
-								serviceUID, servicePort, err := d.getEdgeBackendRef(backendref.BackendRef, gtw.Namespace)
+								serviceUID, servicePort, err := d.getEdgeBackendRef(backendref.BackendRef, httproute.Namespace)
 								if err != nil {
 									d.log.Error(err, "could not find port for service", "namespace", gtw.Namespace, "service", refName)
 								}


jrobsonchase (approved):

internal/store/driver.go
```diff
@@ -946,8 +946,11 @@ func (d *Driver) calculateHTTPSEdgesFromGateway(edgeMap map[string]ingressv1alph
 								}
 
 								refName := string(backendref.Name)
-								//refNamespace := string(*backendref.Namespace)
-								serviceUID, servicePort, err := d.getEdgeBackendRef(backendref.BackendRef, gtw.Namespace)
+								namespace := httproute.Namespace
+								if backendref.BackendRef.Namespace != nil {
+									namespace = string(*backendref.BackendRef.Namespace)
```
nikolay-ngrok, line 951: > Note that when a namespace different than the local namespace is specified, a ReferenceGrant object is required in the referent namespace to allow that namespace’s owner to accept the reference. See the ReferenceGrant documentation for details.

maybe we shouldn't go there just yet?

Abdiramen: Good call! I'll check on the `HTTPRoute` namespace!

Abdiramen: reverted here: https://github.com/ngrok/kubernetes-ingress-controller/pull/356/commits/3bcb295d84e37b89634cde533efa9485fa15d0db

internal/store/driver.go
```diff
@@ -946,8 +946,7 @@ func (d *Driver) calculateHTTPSEdgesFromGateway(edgeMap map[string]ingressv1alph
 								}
 
 								refName := string(backendref.Name)
-								//refNamespace := string(*backendref.Namespace)
-								serviceUID, servicePort, err := d.getEdgeBackendRef(backendref.BackendRef, gtw.Namespace)
+								serviceUID, servicePort, err := d.getEdgeBackendRef(backendref.BackendRef, httproute.Namespace)
```
nikolay-ngrok, line 949: I'd think it would be the best if we show a different kind of error here. Right now it will render as "not found", but we should probably make different namespaces "not supported" for now.

internal/store/driver.go
```diff
@@ -946,8 +946,7 @@ func (d *Driver) calculateHTTPSEdgesFromGateway(edgeMap map[string]ingressv1alph
...
 								if err != nil {
 									d.log.Error(err, "could not find port for service", "namespace", gtw.Namespace, "service", refName)
```
nikolay-ngrok, line 951: looks like we would continue despite the error, is this expected?

Abdiramen: skipped here: https://github.com/ngrok/kubernetes-ingress-controller/pull/356/commits/477310ccd57ad286a38f0ae7bcbdf2766db25fc5

nikolay-ngrok (approved):
👍

PR #357, opened 2024-03-29
add gateway to client info comment

Abdiramen:
## What
- Add `gateway-api` comment to client info

## How
- Add `gateway-api` comment if `useExperimentalGatewayAPI` is set

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -207,15 +207,12 @@ func runController(ctx context.Context, opts managerOpts) error {
 		os.Exit(1)
 	}
 
-	var comments []tunneldriver.TunnelDriverComments
+	var comments tunneldriver.TunnelDriverComments
 
 	if opts.useExperimentalGatewayAPI {
-		comments = append(
-			comments,
-			tunneldriver.TunnelDriverComments{
-				Gateway: "gateway-api",
-			},
-		)
+		comments = tunneldriver.TunnelDriverComments{
+			Gateway: "gateway-api",
+		}
 	}
 
 	td, err := tunneldriver.New(
@@ -223,7 +220,7 @@ func runController(ctx context.Context, opts managerOpts) error {
 			ServerAddr: opts.serverAddr,
 			Region:     opts.region,
 		},
-		comments,
+		&comments,
 	)
 	if err != nil {
 		return fmt.Errorf("unable to create tunnel driver: %w", err)
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -211,11 +211,11 @@ func runController(ctx context.Context, opts managerOpts) error {
 
 	if opts.useExperimentalGatewayAPI {
 		comments = append(
-      comments,
-      tunneldriver.TunnelDriverComments{
-        Gateway: "gateway-api",
-      },
-    )
+			comments,
+			tunneldriver.TunnelDriverComments{
+				Gateway: "gateway-api",
+			},
+		)
 	}
 
 	td, err := tunneldriver.New(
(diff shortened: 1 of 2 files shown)


pkg/tunneldriver/driver.go
```diff
@@ -53,9 +53,9 @@ type TunnelDriverOpts struct {
 }
 
 // New creates and initializes a new TunnelDriver
-func New(logger logr.Logger, opts TunnelDriverOpts) (*TunnelDriver, error) {
+func New(logger logr.Logger, opts TunnelDriverOpts, comments ...string) (*TunnelDriver, error) {
 	connOpts := []ngrok.ConnectOption{
-		ngrok.WithClientInfo("ngrok-ingress-controller", version.GetVersion()),
+		ngrok.WithClientInfo("ngrok-ingress-controller", version.GetVersion(), comments...),
```
jrobsonchase, line 58: Since we already have precedent for sending a json-serialized struct as the comment, could we do that instead? Should make it easier to include more complex information should the need ever arise.

We could also tag all of the fields with `omitempty`, and then skip the comment entirely if it serializes to just `{}`.

Or instead of a struct, a `map[string]any` would work if we wanted more flexibility and fewer types to update when we want to send more things.

jrobsonchase (commented):
We should probably go ahead and structure the comment data, but otherwise looks good to me!

pkg/tunneldriver/driver.go
Abdiramen: here's a type serialized version: https://github.com/ngrok/kubernetes-ingress-controller/pull/357/commits/311168f82644f9cfb8d55d376e9a273b7f291878

Is this what you meant?

cmd/main.go
```diff
@@ -207,10 +207,24 @@ func runController(ctx context.Context, opts managerOpts) error {
...
 
-	td, err := tunneldriver.New(ctrl.Log.WithName("drivers").WithName("tunnel"), tunneldriver.TunnelDriverOpts{
-		ServerAddr: opts.serverAddr,
-		Region:     opts.region,
-	})
+	var comments []tunneldriver.TunnelDriverComments
+
+	if opts.useExperimentalGatewayAPI {
+		comments = append(
+      comments,
```
OfTheDelmer, line 214: hmm is this checking `gofmt`?

OfTheDelmer (approved):

cmd/main.go
Abdiramen: fmt here: https://github.com/ngrok/kubernetes-ingress-controller/pull/357/commits/c4ccc465a3e9bc4cedf80809cac95b1ac4b55cf1

pkg/tunneldriver/driver.go
```diff
@@ -52,10 +53,29 @@ type TunnelDriverOpts struct {
 	Region     string
 }
 
+type TunnelDriverComments struct {
+	Gateway string `json:"gateway,omitempty"`
+}
+
 // New creates and initializes a new TunnelDriver
-func New(logger logr.Logger, opts TunnelDriverOpts) (*TunnelDriver, error) {
+func New(logger logr.Logger, opts TunnelDriverOpts, tunnelComments []TunnelDriverComments) (*TunnelDriver, error) {
```
jrobsonchase, line 61: I think we only really need one of these. Anything else we add to it will be a field on the struct rather than another entry in this slice.

Abdiramen: un-sliced here: https://github.com/ngrok/kubernetes-ingress-controller/pull/357/commits/e26e0a3a8cb2a78653dd00978939154a76f0fb07

jrobsonchase (approved):

PR #358, opened 2024-04-05
set metadata for edges created by the gateway

Abdiramen:
## What
Distinguish between the ingress controller and gateway api in the edge UI with _metadata_

## How
- Split metadata field to ingress and gateway
- Plumb these values to ingress and gateway segments of code

## Breaking Changes
Nope


bobzilladev (approved):
Looks to detangle the metadata in each call path, ingress to ingress, gateway to gateway.

PR #361, opened 2024-04-16
create policy kind

OfTheDelmer:
## What
Introduce a Policy kind with a spec that wraps `EndpointRule` and has a `Direction` string
## How

Generate the policy kind, but exclude the wiring in main to avoid errors

## Breaking Changes

No, this is a new and independent set of feature work.


Abdiramen (approved):

api/ngrok/v1alpha1/policy_types.go
```diff
@@ -0,0 +1,74 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// PolicySpec defines the desired state of Policy
+type PolicySpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Foo is an example field of Policy. Edit policy_types.go to remove/update
+	EndpointRule ingressv1alpha1.EndpointRule `json:"endpoint_rule,omitempty"`
+	Direction    string                       `json:"direction,omitempty"`
```
OfTheDelmer, line 42: We'll need to iterate on this to remove both `EndpointRule` and `Direction` to just deal with any JSON, e.g. just a `Policy` raw json string?

OfTheDelmer: While we've been calling this a `Policy` CRD call it `TrafficPolicyCRD`

internal/controller/ngrok/suite_test.go
```diff
@@ -0,0 +1,98 @@
...
+		CRDDirectoryPaths:     []string{filepath.Join("..", "..", "..", "config", "crd", "bases")},
+		ErrorIfCRDPathMissing: true,
+
+		// The BinaryAssetsDirectory is only required if you want to run the tests directly
+		// without call the makefile target test. If not informed it will look for the
+		// default path defined in controller-runtime which is /usr/local/kubebuilder/.
+		// Note that you must have the required binaries setup under the bin directory to perform
+		// the tests directly. When we run make test it will be setup and used automatically.
+		BinaryAssetsDirectory: filepath.Join("..", "..", "..", "bin", "k8s",
+			fmt.Sprintf("1.28.3-%s-%s", runtime.GOOS, runtime.GOARCH)),
```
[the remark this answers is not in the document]
Abdiramen: Since this file is generated as a convenience, and this doesn't really test the kind you might as well delete the whole file.

api/ngrok/v1alpha1/ngroktrafficpolicy_types.go
```diff
@@ -0,0 +1,73 @@
...
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	Policy json.RawMessage `json:"policy,omitempty"`
+}
+
+// NgrokTrafficPolicyStatus defines the observed state of NgrokTrafficPolicy
+type NgrokTrafficPolicyStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
```
Abdiramen, line 47: This status should also contain a policy string
```suggestion
		Policy json.RawMessage `json:"policy,omitempty"`
```

bobzilladev: Curious, why is the duplicated storage of the json needed?

Abdiramen: The `spec` describes the desired state while the `status` describes the current state, but I guess since we don't need to do any complex steps to reconcile the difference between the two we could only use the `spec` field and always construct on change.

OfTheDelmer: ahh, well if they're just conventionally mirroring like that I see why we're putting policy here also. Thanks for clarifying 🚀

PR #362, opened 2024-04-18
helm: add apiURL option

jrobsonchase:
## What

Adds a helm value to set the API URL as an alternative to the `NGROK_API_URL` environment variable.

## How

* Boring helm plumbing
* Keep the default from the environment, but override it if the helm value is set

## Breaking Changes

None, since no one should have this helm value set. Existing users can continue with the env var and switch to the helm value at their leisure.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -82,6 +82,7 @@ type managerOpts struct {
 	electionID                string
 	probeAddr                 string
 	serverAddr                string
+	apiURL                    string
 	controllerName            string
 	watchNamespace            string
 	metaData                  string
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.10.4
+
+- Add the `--api-url` option
+  This can be used to set the endpoint for the ngrok API.
+  It can be set through via the helm `apiURL` value.
```
nikolay-ngrok, line 12: you think we can include https://github.com/ngrok/kubernetes-ingress-controller/pull/360 here?

nikolay-ngrok (approved):

CHANGELOG.md
jrobsonchase: Ehhh... maybe let's break out the version bump into its own change and do both there. Didn't realize I would be pulling in other stuff.

jrobsonchase: #363

El-o-heka (approved):
{°¥°}

PR #368, opened 2024-04-26
switch edge kinds to raw json policy

OfTheDelmer:
## What
Switch over to raw json policy in https,tls, and tcp edge kinds

## How
- utilize new base client and implement `Delete` and `Replace` for `RawPolicy` in client set
- Switch reconcile over to new `RawPolicy` escape hatch
- update driver to marshal the `EndpointPolicy` for moduleset into a byte array
- Test-test-test 
   - make deploy local crd changes
   - syncing edges locally and moduleset
   -  verifying the policy changes were picked up in ngrok

### Validation

We were able to sync the following to edges and see a policy

```yaml
---
kind: Tunnel
metadata:
  name: test-tunnel
spec:
  backend:
    protocol: HTTPS
  forwardsTo: kubernetes.default.svc:443
  labels:
    app: httpsedgepolicytest---
kind: NgrokModuleSet
metadata:
  name: policy-module
modules:
  policy:
    inbound:
      - name: FooBarParamNotFound
        expressions:
          - "'bar' in getQueryParam('foo')"
        actions:
          - type: "custom-response"
            config:
              status_code: 404
              content: not found
              headers:
                content-type: text/plain
      - name: BazCookieForLargeRequests
        expressions:
          - "!hasReqCookie('baz')"
          - "req.ContentLength > 5000"
        actions:
          - type: "deny"
    outbound:
      - name: LogUnsuccessfulRequests
        expressions:
          - "res.StatusCode != '200' && res.StatusCode != '204'"
        actions:
          - type: "log"
            config:
              metadata:
                hostport: modulesetpolicytest.ngrok.dev:443
                success: false
---
kind: Ingress
metadata:
  name: example-ingress
  annotations:
    k8s.ngrok.com/modules: policy-module
spec:
  ingressClassName: ngrok
  rules:
    - host: modulesetpolicytest.ngrok.dev
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: game-2048
                port:
                  number: 80
```

```yaml
---
kind: Domain
metadata:
  name: tlsedgetest-ngrok-app
spec:
  domain: tlsedgetest.ngrok.app
---
kind: TLSEdge
metadata:
  name: test-edge
spec:
  hostports:
    - tlspolicytest.ngrok.dev:443
  backend:
    labels:
      app: tlsedgetest
  policy:
    inbound:
      - name: "LimitToKnownIPs"
        expressions:
          - "conn.ClientIP != '8.8.8.8'"
          - "conn.ClientIP != '9.9.9.9'"
        actions:
          - type: deny

---
kind: Tunnel
metadata:
  name: test-tunnel
spec:
  backend:
    protocol: TLS
  forwardsTo: kubernetes.default.svc:443
  labels:
    app: tlsedgetest
```


internal/controller/ingress/tcpedge_controller.go
```diff
@@ -388,26 +388,25 @@ func (r *TCPEdgeReconciler) listTCPEdgesForIPPolicy(ctx context.Context, obj cli
...
-
-	endpointPolicy := policy.ToNgrok()
+	client := r.NgrokClientset.EdgeModules().TCP().RawPolicy()
 
 	// Early return if nothing to be done
-	if endpointPolicy == nil {
+	if policy == nil {
 		if remoteEdge.Policy == nil {
-			r.Log.Info("Module matches desired state, skipping update", "module", "Policy", "comparison", routeModuleComparisonBothNil)
+			r.Log.Info("Module matches desired state, skipping update", "module", "TrafficPolicy", "comparison", routeModuleComparisonBothNil)
```
bobzilladev, line 396: Other two specs don't update this, and we aren't updating the field name for backwards compatibility, so maybe change these 3 `TrafficPolicy` back to `Policy`?

internal/ngrokapi/edge_modules_tls_clientset.go
```diff
@@ -23,6 +74,7 @@ type defaultTLSEdgeModulesClientset struct {
 	mutualTLS      *tls_edge_mutual_tls.Client
 	tlsTermination *tls_edge_tls_termination.Client
 	policy         *tls_edge_policy.Client
+	newPolicy      *rawTLSPolicyClient
```
bobzilladev, line 77: should this and below be `rawPolicy`?

api/ingress/v1alpha1/ngrok_common.go
```diff
@@ -431,6 +431,7 @@ func (amazon *EndpointOAuthAmazon) ToNgrok(clientSecret *string) *ngrok.Endpoint
 	return mod
 }
 
+type EndpointTrafficPolicy json.RawMessage
```
Abdiramen, line 434: We should define this in a new file `api/ngrok/v1alpha1/common.go`. This way as we can onboard the new endpoints content there as-well.

internal/ngrokapi/edge_modules_https.go
```diff
@@ -50,13 +57,58 @@ func (c *defaultHTTPSEdgeModulesClientset) TLSTermination() *https_edge_tls_term
 	return c.tlsTermination
 }
 
+type RawHTTPSModuleSetPolicy json.RawMessage
```
Abdiramen, line 60: Let's import this from `api/ngrok/v1alpha1/common.go`, since we'll eventually change this type it would be nice if we only had to change it in one spot.

OfTheDelmer: This was a client alias that was helpful for `ngrokapi`, and I decided to drop it to avoid confusion and simplify

api/ingress/v1alpha1/ngrok_common.go
OfTheDelmer: We may want to revisit this in the plumbing stage for `ngrok/v1alpha1` kinds, but right now, colocation next to `EnpointPolicy` made sense to us in this diff because we are modifying mostly the `api/ingress` controller to post schemaless json.

Abdiramen: cool, in that case this can wait for the pluming step. The reason why I'm interested in removing this concept from the `api/ingress` is because I wish to separate all ngrok concepts from ingress and gateway. This way we can move towards creating `gateway` and `ingress` changes with minimal `ngrok` concept leaks

Abdiramen (approved):

PR #370, opened 2024-04-29
modules to traffic policy

OfTheDelmer:
## What

Updates the https edge ingress route spec with either the annotated modset policy or the annotated traffic policy policy

## How

- Update the driver and store to parallel the `GetModuleSetV1` path for the traffic policy CRD
- Update the tls, tcp and https edge types to use `json.RawMessage` because the alias of `RawMessage` causes issues with kubebuilder validation internally
- Get the single policy annotation from the annotations on the ingress

## Validation

- Use the following ingress + traffic policy yaml to apply resources locally
- See changes sync to edges on ngrok.com with the corresponding traffic policy

`ingress_trafficpolicy_crd_test.yml`

```yaml
---
kind: NgrokTrafficPolicy
apiVersion: ngrok.k8s.ngrok.com/v1alpha1
metadata:
  name: traffic-policy-2
spec:
    policy:
      inbound:
        - name: "test_in"
          expressions:
            - "req.Method == 'PUT'"
          actions:
            - type: "deny"
---
kind: Ingress
metadata:
  name: example-ingress
  annotations:
    k8s.ngrok.com/traffic-policy: traffic-policy-2
spec:
  ingressClassName: ngrok
  rules:
    - host: trafficpolicyingress-test.ngrok.dev
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: game-2048
                port:
                  number: 80
```

2048.yml

```yaml
apiVersion: v1
kind: Service
metadata:
  name: game-2048
  namespace: ngrok-ingress-controller
spec:
  ports:
    - name: http
      port: 80
      targetPort: 80
  selector:
    app: game-2048
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: game-2048
  namespace: ngrok-ingress-controller
spec:
  replicas: 1
  selector:
    matchLabels:
      app: game-2048
  template:
    metadata:
      labels:
        app: game-2048
    spec:
      containers:
        - name: backend
          image: alexwhen/docker-2048
          ports:
            - name: http
              containerPort: 80
```


internal/store/driver.go
```diff
@@ -844,11 +870,17 @@ func (d *Driver) calculateHTTPSEdgesFromIngress(edgeMap map[string]ingressv1alph
...
 				}
 
-				policyStr, err := json.Marshal(modSet.Modules.Policy)
+				var policyStr []byte
 
-				if err != nil {
-					d.log.Error(err, "cannot convert module-set policy json", "Policy", modSet.Modules.Policy)
-					continue
+				// Handle deciding which policy we want to serialize
+				if trafficPolicy == nil {
```
bobzilladev, line 876: this always felt a little weird, to check trafficpolicy to see if we use modset. it may work ok, but might flow better if we check if trafficPolicy not nil, if so then use trafficPolicy. fallback to check if modules.policy is not nil, then use modules.policy. so everything is nil checked right before use.

bobzilladev (approved):
lgtm still, one comment

internal/store/driver.go
OfTheDelmer: Made this change, ran a make deploy and didn't see my policy on the edge. Same whne I stashed the changes. Working on deleting the `minikube` and starting fresh.

OfTheDelmer: @bobzilladev hot off the press: https://github.com/ngrok/kubernetes-ingress-controller/pull/370/commits/de91cbcc6c8da8f111a3f9c5c20b793e90e34d1b

internal/store/driver.go
```diff
@@ -909,6 +890,34 @@ func (d *Driver) calculateHTTPSEdgesFromIngress(edgeMap map[string]ingressv1alph
...
+
+	if modSet.Modules.Policy != nil && trafficPolicy != nil {
+		return nil, fmt.Errorf("cannot have both a traffic policy and a moduleset policy on ingress: %s", ingress.Name)
+	}
+
+	if trafficPolicy != nil {
+		return trafficPolicy.Spec.Policy, nil
+	}
+
+	if policyJSON, err = json.Marshal(modSet.Modules.Policy); err != nil {
```
bobzilladev, line 913: Seems like `modSet.Modules.Policy` needs a nil check? It's legal to not have a policy set from either modset or policy-crd. Unless this behaves nicely in that scenario (not erroring, returning an empty byte slice).

OfTheDelmer: Yeah, and this aligns with `json.Marshal` call with `modSet.Modules.Policy` not having a `nil` check prior to the traffic policy considerations: https://github.com/ngrok/kubernetes-ingress-controller/pull/370/files#diff-9ece2bb7eeaf621263ba5abe1c62e28734da229e31faf7348e3aa4b260fc73aaL847.

However, you're on to something because in general this func is now getting a pointer mod set and this abstracted helper func should `nil` check `modSet` in general.

internal/store/driver.go
```diff
@@ -910,6 +910,10 @@ func (d *Driver) getPolicyJSON(ingress *netv1.Ingress, modSet *ingressv1alpha1.N
 		return trafficPolicy.Spec.Policy, nil
 	}
 
+	if modSet == nil {
+		return policyJSON, nil
```
OfTheDelmer, line 914: I ran this last change through all the combinations and it's good to go: `ingress + module set`, `ingress + trafffic policy`, `ingress + module set + traffic policy` and just `ingress`. They behaved as expected in logs and synced to ngrok edges

Abdiramen (approved):

Issue #133, opened 2023-02-09
Add Support for Mutual TLS Edge Module


jonstacks:
### Description

Add support for [Mutual TLS](https://ngrok.com/docs/cloud-edge/modules/mutual-tls) Edge module. 

### Use Case

Allow users to configure Mutual TLS for authenticating clients

alex-bezek:
There isn't a pressing need to do this at the moment but we may do it in the future. When we look at doing this, we should consider where we are at with migrating away from the edges concept.

PR #373, opened 2024-05-09
feat: Add support for mutualTLS

jonstacks:
closes: #133 

## What

We already have mTLS support in `TLSEdge`s. However, we don't have mTLS support for `HTTPSEdge`s and, by extension, native `Ingress` resources. This closes out the original ticket by adding support for mutual TLS to the following resources:

* NgrokModuleSets
* HttpsEdges
* Ingress (via modules annotation)

## How

1. Add support for `mutualTLS` to the `Spec` of `HTTPSEdge` and update the test that checks for equality of the returned edge to the spec.
2. Add support for `mutualTLS` to the `NgrokModuleSets`. Longer term, I think it would be cool to integrate with secrets/cert-manager and create the CA on the fly, but they way we are doing it today doesn't preclude that. We can always turn it into an `[]interface{}` later and parse it on the fly as either a string or a type of `ref`. Example:
```yaml
kind: NgrokModuleSet
metadata:
  name: mtls
modules:
  mutualTLS:
    certificateAuthorities:
    - ca_2gDZI7eD3fpxmKGOVqzVuETKedf # This ID can be found on https://dashboard.ngrok.com/security/tls/cert-authorities
```
3. When computing the combined moduleset for an HTTPSEdge, if the result has `mutualTLS`, copy that to the generated edge spec that creates/updates the `HTTPSEdge`.
4. Updated the docs on route modules

## Breaking Changes

There are changes to the CRDs that will need to be applied if the user has an existing install with helm already. This should be noted in the upgrade notes of the next release as I think there are already a few other changes that fall into this category as well.


internal/controller/ingress/httpsedge_controller.go
```diff
@@ -349,6 +353,28 @@ func (r *HTTPSEdgeReconciler) setEdgeTLSTermination(ctx context.Context, edge *n
 	return err
 }
 
+func (r *HTTPSEdgeReconciler) setEdgeMutualTLS(ctx context.Context, edge *ngrok.HTTPSEdge, mtls *ingressv1alpha1.EndpointMutualTLS) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	client := r.NgrokClientset.EdgeModules().HTTPS().MutualTLS()
+	if mtls == nil {
+		if edge.MutualTls == nil {
+			log.V(1).Info("Edge Mutual TLS matches spec")
```
bobzilladev, line 362: Should this `return nil` like `setEdgeTLSTermination` does? Looks like this will move forward with a delete of what it thinks isn't there.

bobzilladev (approved):
a comment about a likely missing `return`, otherwise lgtm

internal/controller/ingress/httpsedge_controller.go
jonstacks: You are right, it should have an early return if both are `nil`. I'll fix it. I don't think I caught it in my testing because if the desired state is no mTLS, I think it may still be valid to delete the mTLS config from the edge. Either way, I'll fix it. Thanks @bobzilladev.

PR #376, opened 2024-05-16
Add GatewayClass to cachestore

bobzilladev:
Resolves:

> 2024-05-16T17:41:30Z	ERROR	cache-store-driver	error updating object in create	{"UpdateStoreHandlerFor": "", "object": {"name": "simple-gatewayclass"}, "error": "unsupported object type: *v1.GatewayClass"}

Also logs when a gateway does not have an https listener (so cannot be used).

### DIFF
--- a/internal/store/cachestores.go
+++ b/internal/store/cachestores.go
@@ -37,8 +37,9 @@ type CacheStores struct {
 	ServiceV1      cache.Store
 
 	// Gateway API Stores
-	Gateway   cache.Store
-	HTTPRoute cache.Store
+	Gateway      cache.Store
+	GatewayClass cache.Store
+	HTTPRoute    cache.Store
 
 	// Ngrok Stores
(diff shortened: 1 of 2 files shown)


nikolay-ngrok (approved):
👍

internal/store/cachestores.go
```diff
@@ -37,8 +37,9 @@ type CacheStores struct {
 	ServiceV1      cache.Store
 
 	// Gateway API Stores
-	Gateway   cache.Store
-	HTTPRoute cache.Store
+	Gateway      cache.Store
+	GatewayClass cache.Store
```
OfTheDelmer, line 41: Nice

OfTheDelmer (approved):
Looks good. Had a question about the `store.go` being updated, but it looks like this is related to some internal cache store changes not called by us.

PR #377, opened 2024-05-17
Add extensionRef support for policy crd inclusion

bobzilladev:
## What

Adds ability for gateway api to use extensionRef to include a Policy CRD.

## How

Adding a filter like this will plumb the policy data through:
```
- type: ExtensionRef
  extensionRef:
     group: ngrok.k8s.ngrok.com
     kind: NgrokTrafficPolicy
     name: my-traffic-policy
```

## Breaking Changes

No


Abdiramen (approved):
lgtm, I like the use of that lambda

PR #379, opened 2024-05-22
Update nix flake, go version, and Makefile dep versions

jonstacks:
## What

Keeping our dependencies up to date.

## How

Updates our go version by updating the nix flake. Also pulls in some changes from the kubebuilder project to the Makefile which updates the versions of env, controller-tools, and adds golang-ci lint(we can turn this on at a later time once we fix some of our linter warnings)

## Breaking Changes

No


bobzilladev (approved):
lgtm, and all runs happily on local machine

PR #383, opened 2024-05-28
chore: Remove custom code for non leader-elected controllers

jonstacks:
## What

[We've PR'd this change upstream](https://github.com/kubernetes-sigs/controller-runtime/pull/2117) and it has made it into a release and we are on a high enough version now. This means we can remove the custom code we had for it.

## How
Remove the custom code we had for this

## Breaking Changes
No

## Validation

Scaled the ingress controller replicas to 2 and validated that 2 agents show up in the ngrok UI as expected.


alex-bezek (approved):
very cool to see it finally go in!

Issue #375, opened 2024-05-15
Add support for Load Balancer CRD with ability to integrate with ExternalDNS


ctindel:
### Description

Right now with ngrok's TCP and TLS Edges there's no way to automate integration with helm charts that use ExternalDNS. 

Can we implement the LoadBalancer CRD which would give us a place to put annotations for ExternalDNS to pick up? This would be helpful when working with helm charts like the strimzi kafka chart.

https://strimzi.io/blog/2019/05/23/accessing-kafka-part-5/

PR #384, opened 2024-05-29
refactor: annotations parsers to handle client.Object instead of just networking.Ingress

jonstacks:
## What

The annotations parser currently expects an ingress of type *networking.Ingress. We can make this more flexible for our use case and allow it to parse annotations from any `client.Object`. We'll want this shortly for being able to extract annotations from services for example in #375 

## How

`*networking.Ingress` implements client.Object so all of our changes are just in the `internal/annotations` package.
 
## Breaking Changes
No


bobzilladev (approved):
s/ing/obj/

PR #386, opened 2024-05-29
feat: Auto-provision domain for TLS Edges

jonstacks:
contributes to #375. Pulling this change out for easier review.

## What

The hostports for a given TLS Edge need to be reserved as domains in ngrok before they can be used. Because of this, it makes sense that the TLSEdge handles reserving the domains if they aren't already reserved in the same way that the TCPEdge controller will take care of reserved addrs and the same way that the ingress controller builds the list of domains to reserve in its store across namespaces.

## How

Parse out the hostports defined in the TLSEdge and automatically find or reserve the necessary domains. Also, populate a new field `cnameTargets` map in the `TLSEdge`'s `Status` that contains the CNAME target from the domains. This will allow other controllers which may create TLSEdges to use the CNAMETargets map. For example, Services can use this to update the `.status.loadBalancer.ingress` fields.

Also handles dangling CNAME records more gracefully in the domain controller by retrying them.

## Breaking Changes

No


internal/controller/ingress/tlsedge_controller.go
```diff
@@ -206,10 +224,20 @@ func (r *TLSEdgeReconciler) updateEdge(ctx context.Context, edge *ingressv1alpha
...
 func (r *TLSEdgeReconciler) delete(ctx context.Context, edge *ingressv1alpha1.TLSEdge) error {
+	// Get all owned domains and delete them.
+	log := ctrl.LoggerFrom(ctx)
+
+	// The edge must be deleted before the domains can be deleted
+	log.Info("Deleting TLSEdge", "edge.ID", edge.Status.ID)
 	err := r.NgrokClientset.TLSEdges().Delete(ctx, edge.Status.ID)
 	if err == nil || ngrok.IsNotFound(err) {
 		edge.Status.ID = ""
+		log.Info("Deleted TLSEdge", "edge.ID", edge.Status.ID)
```
bobzilladev, line 235: Probably want to move this line up one so the message gets the ID before it is cleared out.

internal/controller/ingress/tlsedge_controller.go
```diff
@@ -206,10 +224,20 @@ func (r *TLSEdgeReconciler) updateEdge(ctx context.Context, edge *ingressv1alpha
 }
 
 func (r *TLSEdgeReconciler) delete(ctx context.Context, edge *ingressv1alpha1.TLSEdge) error {
+	// Get all owned domains and delete them.
```
bobzilladev, line 227: The domains will cascade delete with the edge update? Or was more intended here to specifically handle domain deletion?

bobzilladev (approved):
Couple comments, nothing blocking.

internal/controller/ingress/tlsedge_controller.go
jonstacks: Whoops! Thanks for the good set of eyes on that. I'll fix it

internal/controller/ingress/tlsedge_controller.go
jonstacks: I think that is an old comment I had back when I was thinking about deleting domains. I was playing around with adding owner references to domains to track what they were used by, but it didn't really pan out. Reserved domains are kind of tricky especially if it is an ngrok owned subdomain. If we delete it, then you may not be able to get it back. For Ingresses we don't actually delete them since that would allow someone else to snag it. I'll remove the comment there

PR #389, opened 2024-05-30
docs: Add note about external-dns policy and dns issues

jonstacks:
## What

Add some documentation about potential issues when using external-dns


ctindel (approved):
LGTM

PR #390, opened 2024-05-30
fix: TLSEdge not reconciling changes to hostports

jonstacks:
## What

Fixes a bug in the TLSEdge controller where changing the hostports in the spec was not triggering an an API call to update the hostports in the API.

## How

* Updating a bunch of log statements so we get more info on the object that triggered the reconcile.
* Add more logging so we can tell what went wrong
* Fix the bug where we were using the hostports from the status field and not the spec.

## Breaking Changes
No

### DIFF
--- a/internal/controller/ingress/tlsedge_controller.go
+++ b/internal/controller/ingress/tlsedge_controller.go
@@ -157,6 +157,8 @@ func (r *TLSEdgeReconciler) create(ctx context.Context, edge *ingressv1alpha1.TL
 }
 
 func (r *TLSEdgeReconciler) update(ctx context.Context, edge *ingressv1alpha1.TLSEdge) error {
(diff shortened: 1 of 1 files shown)


internal/controller/ingress/tlsedge_controller.go
```diff
@@ -180,7 +183,14 @@ func (r *TLSEdgeReconciler) update(ctx context.Context, edge *ingressv1alpha1.TL
 
 	// If the backend or hostports do not match, update the edge with the desired backend and hostports
 	if resp.Backend.Backend.ID != edge.Status.Backend.ID ||
-		!slices.Equal(resp.Hostports, edge.Status.Hostports) {
+		!slices.Equal(resp.Hostports, edge.Spec.Hostports) {
```
jonstacks, line 186: This is the main fix. The rest are all additions/changes to logging.

bobzilladev (approved):

PR #402, opened 2024-07-30
Add a contributing guide

### LINKED ISSUE #346: Add a "CONTRIBUTING.md"
We should provide documentation/guidance on how to add changes to the ingress controller. This will make things easier for users since they will understand how to make their contributions more likely to be accepted on the first go.

jonstacks:
Closes #346 

## What

Adds a `CONTRIBUTING.md` file explaining how to contribute.

## Breaking Changes

No

### DIFF
--- a/helm/ingress-controller/Chart.lock
+++ b/helm/ingress-controller/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 2.20.3
-digest: sha256:993c3981678605068ef47b75614bcc0bcb70570962337b9e03e3ebc588ca1729
-generated: "2024-07-09T11:39:09.193679089-05:00"
+  version: 2.20.5
+digest: sha256:d887f4d52c98e9524f9eb8be04f987acc240288a9c1cf2653cc1b7c221230ee7
+generated: "2024-07-30T16:56:34.90167-05:00"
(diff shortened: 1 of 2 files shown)


helm/ingress-controller/Chart.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 2.20.3
-digest: sha256:993c3981678605068ef47b75614bcc0bcb70570962337b9e03e3ebc588ca1729
-generated: "2024-07-09T11:39:09.193679089-05:00"
+  version: 2.20.5
```
alex-bezek, line 4: is this related or necessary? It doesn't seem like a reason to release a new chart version, but if this causes the sha to change or something, I could see it.

alex-bezek (approved):

helm/ingress-controller/Chart.lock
jonstacks: No, this is just an artifact of our CI and make command that sets up the helm repo https://github.com/ngrok/kubernetes-ingress-controller/blob/89c4f8074f4ff3666f033805b31612a70ef35ba9/Makefile#L223. I think we initially wanted this to keep our helm chart dependencies up to date but I think there are probably better ways of doing this now that don't pull new deps in for unrelated changes.

I think this change is fine to stay in the main branch until we have new features that need a release

Issue #177, opened 2023-03-13
ngrok API Update isn't triggered when secrets change


alex-bezek:
### What happened

Some of the route-module CRD's now reference k8s secrets by name. If a secret is updated though, it won't pickup the change automatically. Since right now, we only synchronize the data when the sync function is triggered, which happens when ingress objects change, this doesn't happen until either another change happens, or the default multi-hour sync period triggers. 

### What you think should happen instead

When the secret is updated, we should quickly realize and update the resource in the ngrok api. 2 approaches come to mind:
- we could continually sync in a background process https://github.com/ngrok/kubernetes-ingress-controller/issues/149
- we could also watch secrets in the edges controller and do reconciliation loops when those update. 

### How to reproduce

Create a EndpointWebhookVerification endpoint with a secret, confirm it works, update the secret and see the ngrok api isn't updated until something triggers a reconciliation.

alex-bezek:
One problem with watching secrets from the https edges controller is that could watch lots of unnecessary secrets and cause a lot of unnecessary calls to the ngrok api since the edge controller is not batched via the store. An alternative approach would be for the ingress controller to:
- in the driver/store when its building the desired edge objects
- for each moduleset on an edge, find all reference secrets
- take all the secrets `metadata.resourceVersion` fields, combine and hash them, and put an annotation on the httpsedge crd
- so if the secrets don't change, the hash remains the same. If the secrets change, the version id in k8s for the secret will update, resulting in a new hash, a new annotation, and a trigger of the reconciliation loop for the https edge controller

PR #406, opened 2024-08-09
fix(service-controller): Updates not working

### LINKED ISSUE #405: LoadBalancer Services not re-reconciling on changes to modulesets or traffic policy
When changing a moduleset, updates do not trigger re-reconciles of services. This is most likely caused by not fully implementing the watches:

https://github.com/ngrok/kubernetes-ingress-controller/blob/fc9f93ee2c09b275c6b48889a159d60d7c54670d/internal/controller/ingress/service_controller.go#L90

jonstacks:
Fixes #405 

## What

Fixes a bug where changing/adding annotations(ex: `k8s.ngrok.com/modules`) did not update the underlying TCP/TLS Edge spec. Also, the service controller was previously not watching for changes to ngrok modulesets or traffic policy CRs. This also fixes that.

## How
* fix bug that was overwriting the desired state with the existing state
* Add watches & indexers so when a moduleset or traffic policy changes, we trigger a reconcile of services that use the moduleset or traffic policy
* Also pulls in an update to the bitnami/common library chart. We probably want to disable to auto update mechanism in favor of just building helm deps initially, but that is out of scope for this.

## Breaking Changes
no

### DIFF
--- a/internal/controller/ingress/service_controller.go
+++ b/internal/controller/ingress/service_controller.go
@@ -462,7 +462,7 @@ func (r *baseSubresourceReconciler[T, PT]) Reconcile(ctx context.Context, c clie
 
 		log.Info(fmt.Sprintf("Updating %T", e), "desired", d, "existing", e)
 		// Fetch the existing resource as it may have been updated
-		if err := c.Get(ctx, client.ObjectKeyFromObject(e), d); err != nil {
+		if err := c.Get(ctx, client.ObjectKeyFromObject(e), e); err != nil {
 			return err
 		}
 		r.mergeExisting(*d, e)
--- a/internal/controller/ingress/service_controller.go
+++ b/internal/controller/ingress/service_controller.go
@@ -57,6 +57,7 @@ import (
 const (
 	OwnerReferencePath     = "metadata.ownerReferences.uid"
 	ModuleSetPath          = "metadata.annotations.k8s.ngrok.com/module-set"
+	TrafficPolicyPath      = "metadata.annotations.k8s.ngrok.com/traffic-policy"
 	NgrokLoadBalancerClass = "ngrok"
 )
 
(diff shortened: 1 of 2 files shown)


internal/controller/ingress/service_controller.go
```diff
@@ -86,9 +92,19 @@ func (r *ServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 			},
-		})
-		// TODO: Add watches for modulesets and traffic policies so we get updates
-
+		}).
+		// Watch modulesets for changes
+		Watches(
+			&ingressv1alpha1.NgrokModuleSet{},
+			handler.EnqueueRequestsFromMapFunc(r.findServicesForModuleSet),
+		).
```
alex-bezek, line 100: Since the moduleset also can reference a secret, I assume we'd also want to trigger a reconcile when a secret is updated? This is the problem with the https edge https://github.com/ngrok/kubernetes-ingress-controller/issues/177 which I assume is applicable here as well? If so, should we also watch the secret here, or should we just solve this problem across multiple controllers at a later time?

internal/controller/ingress/service_controller.go
```diff
@@ -108,6 +124,34 @@ func (r *ServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		}
 	}
 
+	// Index the services by the module set(s) they reference
+	err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Service{}, ModuleSetPath, func(obj client.Object) []string {
```
alex-bezek, line 128: this is new to me, is this just a more efficient way than to get all the objects or something?

alex-bezek (approved):

internal/controller/ingress/service_controller.go
jonstacks: This indexes the services by the module sets the use. This allows us to efficiently answer the question: "which services need to be reconciled when a module set changes" and then trigger the reconciles for those services. For modulesets, of which you can have multiple in the annotation, this comment in the FieldIndexer is important:

https://github.com/kubernetes-sigs/controller-runtime/blob/6f1c3d819175d2c2f72865c6e544b0f78dc8c280/pkg/client/interfaces.go#L196-L202

internal/controller/ingress/service_controller.go
jonstacks: The secret watching needs to happen in the HTTPSEdge,TLSEdge, and TCPEdge controllers. We'll fix that in a future PR

PR #412, opened 2024-08-16
feat: Support for wildcard domains

### LINKED ISSUE #148: Wildcard domains
### Description

The k8s ingress spec specifies that it does support wildcard hostnames https://kubernetes.io/docs/concepts/services-networking/ingress/#hostname-wildcards
For other ingress controllers which are simply proxy servers, they can set up rules to handle wildcard hosts and leave the dns up to the user. With ngrok, a matcher rule doesn't do anything if an edge isn't setup with an actual host with dns routing to it, so we'll need to setup an edge.

Currently if you try to use a wildcard, you get this error

```
2023-02-14T16:57:35Z	ERROR	controllers.ingress	Failed to sync ingress to store	{"error": "Domain.ingress.k8s.ngrok.com \"*-bezek-hello-world-ingress-ngrok-io\" is invalid: metadata.name: Invalid value: \"*-bezek-hello-world-ingress-ngrok-io\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')"}
```

### Use Case

Allow the utilization of wildcard domains

### Related issues

_No response_

jonstacks:
Closes #148 

## What

Allows for using wildcard domains via the ingress controller.

### `Domain` CR naming

The algorithm for creating domains was to aggregate all of the `rules.Host`s from ingresses and the `listener.hostnames` from Gateways. This would give us a list of ngrok `Domain` CRs  to create/update. We currently replace `.` with `-` in the domain name to create the `Domain` CRs. This works fine expect that `*` are not valid in kubernetes resource names. So we now replace `*` with `wildcard` when generating the `Domain` CR Name.  This makes the name now a valid k8s name.

This could cause future problems if you want to reserve the domain `wildcard.example.com` as the name wouldn't be able to be distinguished from the one generated by `*.example.com` as they would both generate a `Domain` CR named `wildcard-example-com`. We should probably look at generating these names like we do for `HTTPSEdge` CRs going forward to prevent this future problem, but this is not handled in this PR. 

### `HTTPS` Edge Labels

Previously we would add a label to the `HTTPSEdge` CR with the domain. This causes problems when using wildcards since `*` is not a valid character for labels. I initially debated about hashing the domain and using that as the label instead. However, we can extract the `hostports` from the existing edge. For `HTTPSEdges` generated by the driver, they will only ever have a single domain per `HTTPSEdge` and will always be listening on port 443. We can use this to compare if the edge in question is the same one as the one we are generating from the Ingress/Gateway configuration. This allows us to get rid of using the domain label or having to implement hashing for it.

## Breaking Changes

No, with the caveat that if you could hit the above edge case temporarily if you are using both `wildcard.example.com` and `*.example.com`

### DIFF
--- a/internal/store/driver.go
+++ b/internal/store/driver.go
@@ -31,10 +31,14 @@ import (
 
 const clusterDomain = "svc.cluster.local" // TODO: We can technically figure this out by looking at things like our resolv.conf or we can just take this as a helm option
 
+var domainNameForResourceNameReplacer = strings.NewReplacer(
+	".", "-", // replace dots with dashes
+	"*", "wildcard", // replace wildcard with the literal "wildcard"
+)
+
 const (
 	labelControllerNamespace = "k8s.ngrok.com/controller-namespace"
 	labelControllerName      = "k8s.ngrok.com/controller-name"
-	labelDomain              = "k8s.ngrok.com/domain"
 	labelNamespace           = "k8s.ngrok.com/namespace"
 	labelServiceUID          = "k8s.ngrok.com/service-uid"
 	labelService             = "k8s.ngrok.com/service"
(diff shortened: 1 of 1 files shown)


alex-bezek (approved):
For the potential edge case of the conflicting `*` vs `wildcard` do we plan to address that right away, or is it something we should document somewhere for our users and support just incase someone hits it?

jonstacks:
> For the potential edge case of the conflicting `*` vs `wildcard` do we plan to address that right away, or is it something we should document somewhere for our users and support just incase someone hits it?

I think it is lower on the priority list. I'll write up an issue for it and link it so it appears in our list of issues.

Issue #56, opened 2022-12-05
Support Multiple Rules per ingress


alex-bezek:
Currently we only support 1 rule per ingress as a shortcut during our initial development phase. We should support multiple rules

jonstacks:
While we are doing this, I'm going to look at supporting multiple rules per ingress as well as multiple rules for the same host defined in multiple ingresses.

jonstacks:
We are going to wait until we implement a store. This will be much easier after that

alex-bezek:
This functionality was handled as a part of the store https://github.com/ngrok/kubernetes-ingress-controller/issues/93

PR #413, opened 2024-08-20
fix(store): Multiple ingress rules per ingress not working

jonstacks:
Was originally supposed to be completed in https://github.com/ngrok/kubernetes-ingress-controller/issues/56, but might have been a slight regression

## What

Adds support for having multiple rules per ingress. Example:
```yaml
kind: Ingress
metadata:
  name: test-ingress
  namespace: default
  annotations:
    k8s.ngrok.com/modules: "compression,tls,only-trusted-ips,circuit-breaker"
spec:
  ingressClassName: ngrok
  rules:
  - host: my-test-app-4.ngrok.io
    http:
      paths:
      - path: /
        pathType: Prefix
        backend:
          service:
            name: test-app-1
            port:
              name: http
  - host: my-test-app-5.ngrok.io
    http:
      paths:
      - path: /app-2
        pathType: Prefix
        backend:
          service:
            name: test-app-2
            port:
              name: http
      - path: /app-1
        pathType: Prefix
        backend:
          service:
            name: test-app-1
            port:
              name: http
```

## How

* We had a check that marked ingresses with multiple rules as an error. Remove that check
* Add a test that checks that when we list ngrok ingresses in the store that we get back ones with more than 1 ingress rule.

## Breaking Changes
No

### DIFF
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -417,9 +417,6 @@ func (s Store) shouldHandleIngressCheckClass(ing *netv1.Ingress) (bool, error) {
 // shouldHandleIngressIsValid checks if the ingress should be handled by the controller based on the ingress spec
 func (s Store) shouldHandleIngressIsValid(ing *netv1.Ingress) (bool, error) {
 	errs := errors.NewErrInvalidIngressSpec()
-	if len(ing.Spec.Rules) > 1 {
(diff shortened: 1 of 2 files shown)


alex-bezek (approved):
Nice. Can't believe we did the code fixes to allow this and forgot to remove the validator 😅

PR #418, opened 2024-08-28
feat(ci): Add a matrix style test for different versions of k8s & helm

jonstacks:
## What

While trying to reproduce #416, I made a GitHub Actions matrix job that allows us to install to do an e2e style deployment of our helm charts. We aren't providing an actual auth token or API key currently, so it will just fail to connect, but we can use this to validate that chart versions install properly along with the CRDs. If we do find a case that can re-create #416, this will also help serve as a regression test in the future.

## How

Created a GitHub action that allows for `workflow_dispatch`, so we can test any of our helm chart versions. I chose the latest 4 versions of helm as well as k8s to test against using a kind cluster

## Breaking Changes

No, this is an internal tool for testing.

### DIFF
--- a/.github/workflows/install-relased-helm.yaml
+++ b/.github/workflows/install-relased-helm.yaml
@@ -1,0 +1,66 @@
+name: Helm install Matrix
+on:
+  workflow_dispatch:
+    inputs:
+      chart_version:
+        description: 'The version of the chart to install'
+        required: true
+        default: 'v0.14.1'
+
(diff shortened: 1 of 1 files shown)


jonstacks:
Here is one of the test runs for reference https://github.com/jonstacks/ngrok-ingress-controller/actions/runs/10605121421

nikolay-ngrok (approved):
👍

PR #424, opened 2024-09-10
Add feature flag support for bindings

hjkatz:
## What

Adds feature flags support (primarily for bindings)

## How

* Support flag: `--enable-feature-bindings`

## Breaking Changes

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -94,6 +94,10 @@ type managerOpts struct {
 	zapOpts                   *zap.Options
 	clusterDomain             string
 
+	// feature flags
+	enableFeatureIngress  bool
+	enableFeatureBindings bool
+
 	// env vars
 	namespace   string
 	ngrokAPIKey string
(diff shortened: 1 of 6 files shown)


jonstacks (approved):

cmd/main.go
```diff
@@ -125,6 +129,13 @@ func cmd() *cobra.Command {
 	c.Flags().BoolVar(&opts.useExperimentalGatewayAPI, "use-experimental-gateway-api", false, "sets up experemental gatewayAPI")
 	c.Flags().StringVar(&opts.clusterDomain, "cluster-domain", "svc.cluster.local", "Cluster domain used in the cluster")
 	c.Flags().StringVar(&opts.rootCAs, "root-cas", "trusted", "trusted (default) or host: use the trusted ngrok agent CA or the host CA")
+
+	// feature flags
+	// default always enabled for now
+	// c.Flags().BoolVar(&opts.enableFeatureIngress, "enable-feature-ingress", true, "Enables the Ingress controller")
+	opts.enableFeatureIngress = true
```
alex-bezek, line 136: What do you see for the future of this? This goes back to what I was mentioning about considering what are feature flags we want to put features behind during development/alpha and then will remove later on once stable and default on. VS long lived configuration options to opt in and out of. 

Are you thinking one day we will want to let people toggle off the ability for Ingress objects to work?

helm/ingress-controller/values.yaml
```diff
@@ -203,3 +203,7 @@ log:
 ## ref: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
 ##
 lifecycle: {}
+
+## @param bindings.enabled Whether to enable the Endpoint Bindings feature
+bindings:
+  enabled: true
```
alex-bezek, line 209: do you want this enabled by default during initial development phases?

cmd/main.go
hjkatz: Yes, we will want to support orthagonal features.

For now these can be on the opts struct, but I imagine a `pkg/featureflags/` and a struct/interface returned that the rest of the code can inspect to see what is enabled and what they should do.

helm/ingress-controller/values.yaml
hjkatz: Nope, I'll go change that. Thank you.

cmd/main.go
alex-bezek: > we will want to support orthagonal features.

Can you explain what you mean here more and how that applies? I get we'll be adding additional new features that will be either feature flagged during roll out, or there will be features orthogonal to the core feature set that people will want to be able to toggle on/off. I just haven't seen Ingress controllers disable ingress. 

Are you thinking there will be a product request to install the operator and only let it create bindings but not create ingress and that needs to be handled at the controller level vs something like not making `ngrok` ingress classes?

hjkatz: > Are you thinking there will be a product request to install the operator and only let it create bindings but not create ingress and that needs to be handled at the controller level vs something like not making ngrok ingress classes?

Yes, exactly this.

The ngrok-operator will be able to enable/disable various features which atm in the code are just 1 feature: the ingress controller.

So that feature is staying enabled by default and we can backport it into a feature flag at a later time.

PR #433, opened 2024-09-16
Update README.md to use ngrok Kubernetes Operator instead of ingress controller. 

stmcallister:
Caught a few more instances of `ingress controller` in README and changed them to use `ngrok Kubernetes Operator`.


jonstacks:
Thanks @stmcallister!

PR #446, opened 2024-10-11
refactor: Split the agent and api controllers

jonstacks:
## What

Splits the agent and API controllers into their own managers and k8s deployments. This separates the data path from the API configuration path so that one or the other can easily be turned on or off and scaled independently. This also helps with providing an upgrade path from the ngrok Ingress Controller to the new ngrok operator.

## How

* Move the helm make commands into the Helm Chart
* The main manager is now the API manager.
* Remove the tunnel controller and driver from the api manager
* Create a new agent manager which has only the tunnel controller
* Wire it all up in helm with some basic options. This is the MVP part, we can go through and add additional options for the agent deployment like a separate PDB, different image tag, etc as we go 

## Breaking Changes

There should not be. This should be a passive change.


docs/deployment-guide/migrating.md
```diff
@@ -0,0 +1,63 @@
...
+   kubectl get crds | grep "ngrok" | awk '{print $1}' | xargs -I '{}' kubectl annotate crd '{}' 'meta.helm.sh/release-name=ngrok-operator' --overwrite
+   ```
+   ```shell
+   kubectl get crds | grep "ngrok" | awk '{print $1}' | xargs -I '{}' kubectl annotate crd '{}' 'meta.helm.sh/release-namespace=ngrok-operator' --overwrite
+   ```
+   ```shell
+   kubectl get ingressclass | grep "ngrok" | awk '{print $1}' | xargs -I '{}' kubectl annotate ingressclass '{}' 'meta.helm.sh/release-name=ngrok-operator' --overwrite
+   ```
+   ```shell
+   kubectl get ingressclass | grep "ngrok" | awk '{print $1}' | xargs -I '{}' kubectl annotate ingressclass '{}' 'meta.helm.sh/release-namespace=ngrok-operator' --overwrite
```
alex-bezek, line 33: I'm not sure how many people use gateway, but we might need similar instructions for the gateway class cluster level resource

alex-bezek (approved):

docs/deployment-guide/migrating.md
jonstacks: Good call, I'll update the instructions to match

docs/deployment-guide/migrating.md
```diff
@@ -0,0 +1,63 @@
+# Migration Guide
+
+This guide provides instructions for migrating from the ngrok Ingress Controller chart to the ngrok Operator chart.
```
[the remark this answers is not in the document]
alex-bezek: Thanks!

cmd/agent/main.go
```diff
@@ -0,0 +1,209 @@
...
+	c.Flags().AddGoFlagSet(goFlagSet)
+
+	return c
+}
+
+func runController(ctx context.Context, opts managerOpts) error {
+	ctrl.SetLogger(zap.New(zap.UseFlagOptions(opts.zapOpts)))
+
+	buildInfo := version.Get()
+	setupLog.Info("starting manager", "version", buildInfo.Version, "commit", buildInfo.GitCommit)
```
hjkatz, line 130: Nit: May be worth going through this file searching for "manager" or similar to change to "agent" or similar

docs/deployment-guide/migrating.md
```diff
@@ -0,0 +1,87 @@
...
+
+   If you have other workloads running in the namespace, you will need to manually delete the old resources.
+
+7. Scale the manager replica count to 1:
+
+    ```shell
+    kubectl -n ngrok-operator scale deployment ngrok-operator-manager --replicas=1
+    ```
+
+    Don't forget to update the `replicaCount` in any future helm commands to match.
```
hjkatz, line 87: <i class='graphite__hidden'>[Re: lines +83 to +88]<br/><br/></i>

Q. Can we double check that this process will not create a kubectl-last-applied annotation that may mess up future deployments if `.spec.replicaCount` is not passed to the `kind: Deployment` template?

Happy to just chat through this offline as a thought exercise to confirm.<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-operator/446?utm_source=unchanged-line-comment">Graphite</a>.</span>

docs/developer-guide/architecture.md
```diff
@@ -30,15 +43,44 @@ The following controllers for the most part manage a single resource and reflect
...
 
 TODO: Update more about the various pieces of the ingress controller portion such as the store, the driver, how annotations work, etc.
 
 <img src="../assets/images/Under-Construction-Sign.png" alt="Under Construction" width="350" />
+
+## Architecture
+
+### Overview
+
+Previously, in the ngrok Ingress Controller, the Tunnel Controller(non leader-elected) ran under the same manager as the rest of the API controllers(leader-elected). This did not allow users to independetly scale the data forwarding and API management features. Going forward, we will be splitting the Tunnel Controller into its own manager so that it can be scaled independently of the rest of the controllers. This will allow users to scale the data forwarding and API management features independently.
```
hjkatz, line 63: Nit: I like the historical context, but I feel that the history should come after the present.

Perhaps this paragraph can be 2 paragraphs:
1. The current architectural intent with 2 deployments
2. The historical reasoning

helm/ngrok-operator/templates/agent/service-account.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.agent.serviceAccount.create -}}
```
hjkatz, line 1: Q. Do we need to do any backwards compatibility checking for different `.Values` paths for the fields moved under `.Values.agent.*`?

hjkatz (approved):

helm/ngrok-operator/templates/agent/service-account.yaml
jonstacks: The `agent` section is a net new addition to the `values.yaml`. There are only overrides right now for `replicas`, `priorityClassName`, and `serviceAccount` currently. This will likely expand over time as we want to override more settings for the agent. These were chosen to start with since:
* `.Values.agent.serviceAccount.*` - We want to use a more tightly scoped cluster role & binding for this deployment, so we need a new service account. We'll create one by default for the new agent deployment.
* `.Values.agent.priorityClassName` - It seems reasonable that end users might want to give forwarding of traffic to pods a higher priority class than the API reconciliation. This defaults to not being set, but can be changed.
* `.Values.agent.replicas` - This is needed for the migration and also allows users to scale the data forwarding path independent of the api manager.

What kind of backwards compatibility checks are you thinking?

cmd/agent/main.go
jonstacks: Done

docs/deployment-guide/migrating.md
hjkatz: Yeah I was worried about that while reading, but I see now that the replicas are always passed so no problemo.

docs/developer-guide/architecture.md
jonstacks: I like that format. This will change again once we include the bindings forwarder. I think I'm going to leave it for now since it will get changed soon and reword it for current state once we have the bindings piece

PR #453, opened 2024-10-17
Implement AggregateBindingEndpoints for interacting with the ngrok api

hjkatz:
## What

`AggregateBindingEndpoints` turns multiple `binding_endpoints` into a map of `hostport -> EndpointBinding`.

This is the abstraction to handle the Many-to-One relationship from Endpoint -> Service (as represented by `kind: EndpointBinding` in our operator).

## How
*Describe the solution*

## Breaking Changes

No

### DIFF
--- a/api/bindings/v1alpha1/bindingconfiguration_types.go
+++ b/api/bindings/v1alpha1/bindingconfiguration_types.go
@@ -25,7 +25,7 @@ SOFTWARE.
 package v1alpha1
 
 import (
-	"github.com/ngrok/ngrok-api-go/v5"
+	v6 "github.com/ngrok/ngrok-api-go/v6"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -104,7 +104,7 @@ type BindingConfigurationStatus struct {
 // BindingEndpoint is a reference to an Endpoint object in the ngrok API that is attached to the kubernetes operator binding
 type BindingEndpoint struct {
 	// Ref is the ngrok API reference to the Endpoint object (id, uri)
-	ngrok.Ref `json:",inline"`
+	v6.Ref `json:",inline"`
 
 	// +kubebuilder:validation:Required
 	// +kubebuilder:default="unknown"
--- a/api/bindings/v1alpha1/endpointbinding_types.go
+++ b/api/bindings/v1alpha1/endpointbinding_types.go
@@ -33,14 +33,14 @@ import (
 
 // EndpointBindingSpec defines the desired state of EndpointBinding
 type EndpointBindingSpec struct {
-	// FQRI "Fully Qualified Resource Identifier" is the unique identifier
+	// EndpointURI is the unique identifier
 	// representing the EndpointBinding + its Endpoints
 	// Format: <scheme>://<service>.<namespace>:<port>
 	//
 	// +kubebuilder:validation:Required
 	// See: https://regex101.com/r/9QkXWl/1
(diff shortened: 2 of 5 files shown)


hjkatz:
* **#453** <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-operator/453?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

internal/ngrokapi/bindingendpoint_aggregator.go
```diff
@@ -0,0 +1,128 @@
...
+	bindingsv1alpha1 "github.com/ngrok/ngrok-operator/api/bindings/v1alpha1"
+)
+
+var (
+	defaultScheme = "https"
+	defaultPort   = map[string]int32{
+		"http":  80,
+		"https": 443,
+		"tls":   443,
+		"tcp":   80, // TODO: Unclear exactly what/if any default port should be provided
```
natasha-jarus, line 18: no default port here, tcp endpoints will always specify a port

api/bindings/v1alpha1/endpointbinding_types.go
```diff
@@ -33,6 +33,15 @@ import (
...
 // EndpointBindingSpec defines the desired state of EndpointBinding
 type EndpointBindingSpec struct {
+	// FQRI "Fully Qualified Resource Identifier" is the unique identifier
+	// representing the EndpointBinding + its Endpoints
+	// Format: <scheme>://<service>.<namespace>:<port>
+	//
+	// +kubebuilder:validation:Required
+	// See: https://regex101.com/r/9QkXWl/1
+	// +kubebuilder:validation:Pattern=`^((?P<scheme>(tcp|http|https|tls)?)://)?(?P<service>[a-z][a-zA-Z0-9-]{0,62})\.(?P<namespace>[a-z][a-zA-Z0-9-]{0,62})(:(?P<port>\d+))?$`
+	FQRI string `json:"fqri"`
```
hjkatz, line 43: Nit: I struggled to come up with a name for this field, but I really do want to store this data because the whole parsing hostport thing should only really happen once.

Thoughts on a name? FQRI is what I came up with in the shower haha

api/bindings/v1alpha1/bindingconfiguration_types.go
```diff
@@ -104,7 +104,7 @@ type BindingConfigurationStatus struct {
 // BindingEndpoint is a reference to an Endpoint object in the ngrok API that is attached to the kubernetes operator binding
 type BindingEndpoint struct {
 	// Ref is the ngrok API reference to the Endpoint object (id, uri)
-	ngrok.Ref `json:",inline"`
+	v6.Ref `json:",inline"`
```
jonstacks, line 107: I think this is okay, but generally we've strayed from embedding ngrok-api-go types since it uses snake_case for json serialization([Example](https://github.com/ngrok/ngrok-api-go/blob/2ebad4353c11ffe54afa2a310fa6d6853c6e95dc/datatypes.go#L68)), where as the kubernetes convention is to use camelCase.

api/bindings/v1alpha1/endpointbinding_types.go
jonstacks: This feels very much like [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.2). Off the cuff, I would go with URI or something similar. I'll think on it though as well.

api/bindings/v1alpha1/bindingconfiguration_types.go
jonstacks: Talked synchronously about this and we're okay with leaving it for now since both of the fields in here are the same either snake_case or camelCase (`id` and `uri`). Only other thing is this probably belongs under `BindingEndpointStatus`.

jonstacks (approved):

hjkatz:
### Merge activity

* **Oct 18, 1:51 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/ngrok/ngrok-operator/453).

PR #455, opened 2024-10-18
feat(ci): Add gofmt to lints

jonstacks:
## What

Makes sure code is `gofmt`'d properly.

## How

Enable the linter in golangci-lint configuration. First commit will fail CI as a test, then we'll push it after its been formatted and it should succeed.

## Breaking Changes
No.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,6 @@
 area/ci:
 - .github/workflows/**/*
+- .golangci.yml
 
 area/controller:
 - '**/*.go'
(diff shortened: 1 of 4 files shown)


jonstacks:
Oh yeah, we aren't going to be easily able to test this with CI since we do some fancy checks for which checks to skip based on which code has changed. Since we aren't changing any of the go code, it gets skipped. If you run `make lint` locally from the first commit it fails though.

hjkatz (approved):

PR #456, opened 2024-10-18
update traffic policy for phase-based naming

TheConcierge:
## What
This change updates our `TrafficPolicy` offerings to support the new phase-based names as well as the new `TrafficPolicy` API. This changelist aims to support the new options without breaking current configurations.

### update `ngrok-api-go` dependency
`6.0.0` is out! I've updated it across the entire controller. Maybe this is a bad idea but :shrug: 

### Edge Reconcile Functions
- updated to use new API
- support both `Policy` and `TrafficPolicy` in the CRD spec
  - this allows us to read newly created and older policy
- removed custom client implementation
  - the shape of the API no longer matches the shape we store our Edge CRDs in, so the custom client is no longer convenience
- if using legacy format (inbound/outbound), emit K8s event warning them to update
  - we still send legacy formats to the backend, where they are still supported
- if policy JSON has "enabled" set, extract it and set on the API call (and send k8s event!)
  - this is only "documented" in the module, which is being deprecated. eventually we can remove this processing

### Gateway API
- Filters were updated to use the new phases
- extension ref was update to use the new phases, which get merged with the policy generated from the gateway API
- legacy policy configured in the extension ref is converted to new phase
  - tls/tcp edges are currently not supported on the Gateway API, so this is a safe translation for now
- merging extension ref policy with Filter policy now uses more generic types instead of strict types from the Module config
  - this allows user supplied config to utilize new phases/formats without having to necessarily update the operator
  - decoupling from the deprecated Module types will make it easier to fully remove in the future 
 
### Ingress TrafficPolicy CRD
- save policy to Edge CRD as `TrafficPolicy` instead of `Policy`

### Module
- mark `Policy` module as deprecated
- save policy to Edge CRD as `TrafficPolicy` instead of `Policy`

### <Proto>Edge CRD
- adds `TrafficPolicy` option
- marks `Policy` as deprecated

## Breaking Changes
This PR should not have any breaking changes and instead opts to punt breaking changes to later down the road.

### `NgrokTrafficPolicy` CRD
- creating new resources after the controller upgrade works
- existing policy CRDs using `inbound`/`outbound` continue to work during/after upgrade

### `NgrokModuleSet` CRD
Even though this is deprecated, we need to make sure older configurations continue to work on upgrade

- existing policy CRDs using `inbound`/`outbound` continue to work during/after upgrade

### Gateway API
- after upgrade, existing policy is mapped to new phases, properly merged with "hard-coded" filters
- changing configured policy correctly gets reflected on edge

### `e2e` scripts

Run without failure!

Well...

I did get this failure. However, this appears to possibly be something with the setup/test case vs something not functioning. It does 404?
```--- 
Testing './e2e-fixtures/ingress-class/config-different.yaml' with Edge 'k8s-ops-test-2.ngrok.app'
Testing 'https://k8s-ops-test-2.ngrok.app/' expecting 'HTTP/2 404': FAILED!
  Expected:"HTTP/2 404" received:"HTTP/1.1 404 Not Found" with:

        HTTP/1.1 404 Not Found
        Connection: close
        Content-Type: text/html
        Ngrok-Error-Code: ERR_NGROK_3200
        Referrer-Policy: no-referrer
        Date: Tue, 22 Oct 2024 19:09:50 GMT
```

### DIFF
--- a/api/ingress/v1alpha1/ngrok_common.go
+++ b/api/ingress/v1alpha1/ngrok_common.go
@@ -3,7 +3,7 @@ package v1alpha1
 import (
 	"encoding/json"
 
-	"github.com/ngrok/ngrok-api-go/v5"
+	"github.com/ngrok/ngrok-api-go/v6"
 	"k8s.io/apimachinery/pkg/api/resource"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -487,62 +487,4 @@ type EndpointAction struct {
 	// +kubebuilder:pruning:PreserveUnknownFields
 	// +kubebuilder:validation:Type=object
 	Config json.RawMessage `json:"config,omitempty"`
-}
-
-func (policy *EndpointPolicy) ToNgrok() *ngrok.EndpointPolicy {
-	if policy == nil {
-		return nil
-	}
-
-	var inbound []ngrok.EndpointRule
-	for _, rule := range policy.Inbound {
-		p := rule
-		inbound = append(inbound, *p.ToNgrok())
-	}
-	var outbound []ngrok.EndpointRule
-	for _, rule := range policy.Outbound {
-		p := rule
-		mod := p.ToNgrok()
-		if mod != nil {
-			outbound = append(outbound, *mod)
-		}
-	}
-
-	return &ngrok.EndpointPolicy{
-		Enabled:  policy.Enabled,
-		Inbound:  inbound,
-		Outbound: outbound,
-	}
-}
-
-func (rule *EndpointRule) ToNgrok() *ngrok.EndpointRule {
-	if rule == nil {
-		return nil
-	}
-
-	var actions []ngrok.EndpointAction
-	for _, action := range rule.Actions {
-		a := action
-		mod := a.ToNgrok()
-		if mod != nil {
-			actions = append(actions, *mod)
-		}
-	}
-
-	return &ngrok.EndpointRule{
-		Expressions: rule.Expressions,
-		Actions:     actions,
-		Name:        rule.Name,
-	}
-}
-
-func (action *EndpointAction) ToNgrok() *ngrok.EndpointAction {
-	if action == nil {
-		return nil
-	}
-
-	return &ngrok.EndpointAction{
-		Type:   action.Type,
-		Config: action.Config,
-	}
 }
--- a/internal/controller/ingress/httpsedge_controller.go
+++ b/internal/controller/ingress/httpsedge_controller.go
@@ -44,12 +44,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/go-logr/logr"
-	"github.com/ngrok/ngrok-api-go/v5"
-	"github.com/ngrok/ngrok-api-go/v5/backends/tunnel_group"
+	"github.com/ngrok/ngrok-api-go/v6"
+	"github.com/ngrok/ngrok-api-go/v6/backends/tunnel_group"
 	ingressv1alpha1 "github.com/ngrok/ngrok-operator/api/ingress/v1alpha1"
 	"github.com/ngrok/ngrok-operator/internal/controller"
 	ierr "github.com/ngrok/ngrok-operator/internal/errors"
+	"github.com/ngrok/ngrok-operator/internal/events"
 	"github.com/ngrok/ngrok-operator/internal/ngrokapi"
+	"github.com/ngrok/ngrok-operator/internal/util"
 )
 
 type routeModuleComparision string
@@ -197,6 +199,7 @@ func (r *HTTPSEdgeReconciler) reconcileRoutes(ctx context.Context, edge *ingress
 	}
 
 	routeModuleUpdater := &edgeRouteModuleUpdater{
+		recorder:         r.Recorder,
 		edge:             edge,
 		clientset:        r.NgrokClientset.EdgeModules().HTTPS().Routes(),
 		ipPolicyResolver: controller.IpPolicyResolver{Client: r.Client},
@@ -537,6 +540,8 @@ func (r *tunnelGroupBackendReconciler) findOrCreate(ctx context.Context, backend
 }
 
 type edgeRouteModuleUpdater struct {
+	recorder record.EventRecorder
+
 	edge *ingressv1alpha1.HTTPSEdge
 
 	clientset ngrokapi.HTTPSEdgeRouteModulesClientset
@@ -556,7 +561,7 @@ func (u *edgeRouteModuleUpdater) updateModulesForRoute(ctx context.Context, rout
 		u.setEdgeRouteOIDC,
 		u.setEdgeRouteSAML,
 		u.setEdgeRouteWebhookVerification,
-		u.setEdgeRoutePolicy,
+		u.setEdgeRouteTrafficPolicy,
 	}
 
 	for _, f := range funcs {
@@ -1065,14 +1070,16 @@ func (r *HTTPSEdgeReconciler) takeOfflineWithoutAuth(ctx context.Context, route 
 	return nil
 }
 
-func (u *edgeRouteModuleUpdater) setEdgeRoutePolicy(ctx context.Context, route *ngrok.HTTPSEdgeRoute, routeSpec *ingressv1alpha1.HTTPSEdgeRouteSpec) error {
+func (u *edgeRouteModuleUpdater) setEdgeRouteTrafficPolicy(ctx context.Context, route *ngrok.HTTPSEdgeRoute, routeSpec *ingressv1alpha1.HTTPSEdgeRouteSpec) error {
 	log := ctrl.LoggerFrom(ctx)
-	policy := routeSpec.Policy
-	client := u.clientset.RawPolicy()
+
+	trafficPolicy := routeSpec.Policy
+
+	client := u.clientset.TrafficPolicy()
 
 	// Early return if nothing to be done
-	if policy == nil {
-		if route.Policy == nil {
+	if trafficPolicy == nil {
+		if route.TrafficPolicy == nil {
 			u.logMatches(log, "Policy", routeModuleComparisonBothNil)
 			return nil
 		}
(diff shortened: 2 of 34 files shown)


api/ingress/v1alpha1/ngrok_common.go
```diff
@@ -463,6 +463,14 @@ func (amazon *EndpointOAuthAmazon) ToNgrok(clientSecret *string) *ngrok.Endpoint
 	return mod
 }
 
+// EndpointTrafficPolicy is not exposed and is mostly used for assisting in processing Gateway API filters.
+type EndpointTrafficPolicy struct {
+	// OnHttpRequest traffic rule
+	OnHttpRequest []EndpointRule `json:"on_http_request,omitempty"`
+	// OnHttpResponse traffic rule
+	OnHttpResponse []EndpointRule `json:"on_http_response,omitempty"`
```
hjkatz, line 471: Nit: In k8s we prefer json fields be `camelCased` and the ngrok-go-sdk prefers them `lower_snake_cased`

@jonstacks can say more to this

internal/controller/ingress/httpsedge_controller.go
```diff
@@ -1079,11 +1092,30 @@ func (u *edgeRouteModuleUpdater) setEdgeRoutePolicy(ctx context.Context, route *
 		return client.Delete(ctx, edgeRouteItem(route))
 	}
 
-	log.Info("Updating Policy module")
-	_, err := client.Replace(ctx, &ngrokapi.EdgeRoutePolicyRawReplace{
+	module := ngrok.EndpointTrafficPolicy{}
+
+	if util.IsLegacyPolicy(trafficPolicy) {
+		u.recorder.Eventf(u.edge, v1.EventTypeWarning, "UpdateWarning", "Traffic Policy is using legacy directions: ['inbound', 'outbound']. Update to new phases: ['on_tcp_connect', 'on_http_request', 'on_http_response']")
```
hjkatz, line 1098: Nit: The 3rd field of k8s events is a grouping key that can match events together and update their messages. I recommend using `Updated` for both success and failure events related to updating. Similar for `Created`, etc...

internal/controller/ingress/httpsedge_controller.go
```diff
@@ -1079,11 +1092,30 @@ func (u *edgeRouteModuleUpdater) setEdgeRoutePolicy(ctx context.Context, route *
...
+	}
+	module.Enabled = enabledFieldVal
+
+	if enabledFieldVal != nil {
+		u.recorder.Eventf(u.edge, v1.EventTypeWarning, "UpdateWarning", "Traffic Policy has 'enabled' set. This is a legacy option that will stop being supported soon.")
+	}
+
+	module.Value = string(trafficPolicy)
+
+	log.Info("Updating Traffic Policy module")
```
hjkatz, line 1114: Nit: Can we add a k8s event here for success on the object too?

internal/controller/ingress/tcpedge_controller.go
```diff
@@ -447,25 +449,53 @@ func (r *TCPEdgeReconciler) listTCPEdgesForIPPolicy(ctx context.Context, obj cli
 }
 
 func (r *TCPEdgeReconciler) updatePolicyModule(ctx context.Context, edge *ingressv1alpha1.TCPEdge, remoteEdge *ngrok.TCPEdge) error {
-	policy := edge.Spec.Policy
-	client := r.NgrokClientset.EdgeModules().TCP().RawPolicy()
+	if edge.Spec.Policy != nil && edge.Spec.TrafficPolicy != nil {
+		return fmt.Errorf("configuration contains both Policy and Traffic Policy")
+	}
+
+	client := r.NgrokClientset.EdgeModules().TCP().TrafficPolicy()
```
hjkatz, line 456: Q. TCP here but policies only support HTTP. What's going on here?

internal/store/driver.go
```diff
@@ -1259,8 +1255,36 @@ type AddHeadersConfig struct {
 	Headers map[string]string `json:"headers"`
 }
 
-func (d *Driver) handleExtensionRef(extensionRef *gatewayv1.LocalObjectReference, namespace string, inboundRules *EndpointRules,
-	outboundRules *EndpointRules) error {
+// extractPolicy parses the policy message into a format such that it can be combined with policy from other filters.
+// If the legacy "inbound/outbound" format is detected, inbound remaps to `on_http_request`, outbound remaps to
+// `on_http_response`. This is safe so long as HTTP Edges are the only ones supported on the gateway API.
+func extractPolicy(jsonMessage json.RawMessage) (ingressv1alpha1.EndpointTrafficPolicy, error) {
```
hjkatz, line 1261: Can we add unit tests for this function?

internal/util/traffic_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsLegacyPolicy(t *testing.T) {
+	testCases := []struct {
+		name             string
+		msg              json.RawMessage
+		expectedIsLegacy bool
+	}{
```
hjkatz, line 16: Nit: Can we add a `nil` and empty `[]byte()` cases?

internal/util/traffic_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+		},
+		{
+			name:             "message is not valid json",
+			msg:              []byte(`ngrok is All-in-one API gateway, Kubernetes Ingress, DDoS protection, firewall, and global load balancing as a service.`),
+			expectedIsLegacy: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
hjkatz, line 45: Nit: It's useful to mark the `t.Run()` cases as `t.Parallel()` (and the top-level test too if you want)

internal/util/traffic_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	testCases := []struct {
+		name                  string
+		msg                   json.RawMessage
+		expectedReturnedMsg   json.RawMessage
+		expectedSetEnabledVal *bool
+		expectedErr           error
+	}{
+		{
+			name:                  "message has enabled in top level field (true)",
```
hjkatz, line 64: Nit: Same here for nil, empty, and invalid json cases.

internal/controller/ingress/tcpedge_controller.go
TheConcierge: policies support TCP and TLS

internal/controller/ingress/httpsedge_controller.go
TheConcierge: yep, i'll change this!

internal/controller/ingress/tcpedge_controller.go
TheConcierge: i know the gateway api only supports HTTP atm, but from what i understand, you can still use policy on TLS and TCP edges using the modules and/or the `NgrokTrafficPolicy` CRD

api/ingress/v1alpha1/ngrok_common.go
TheConcierge: this is going to get written to the `<protocol>Edge` CRDs, then directly extracted and shoved into the ngrok API.   so we need it in the format we are eventually going to send it in. 

we could add additional code in the reconciler to convert between camel and snake case, but i feel like i'd rather not do that. especially because, with the `NgrokTrafficPolicy` CRD, the phases will be camel case anyway. So this will make the `NgrokTrafficPolicy` CRD look the same between the different configuration methods 🤷

TheConcierge: > we could add additional code in the reconciler to convert between camel and snake case

we have some special top-level key processing already in the reconciler, but that stuff is theoretically temporary and would be _really_ nice to rip out eventually

TheConcierge: i know that's a lot of stuff, please let me know if you want me to elaborate on anything here

internal/controller/ingress/tcpedge_controller.go
TheConcierge: there's also an `on_tcp_connect` phase :)

TheConcierge: we just don't have that phase explicitly written down anywhere within the operator. the only place that _needs_ the phases is the gateway API, since we convert the Filters into policy. Right now, Filters are only valid for HTTP edges, so those are the only phases we are concerned with atm

TheConcierge: for TCP/TLS edges using modules or the `NgrokTrafficPolicy`, we take their raw policy and just shove it into the api call. So for those edges, the user can use the `on_tcp_connect` phase

TheConcierge: fwiw it would be nice if we could completely delete awareness of phases from the operator, but I'm not sure if that's fully possible considering we need to translate the Filters. 

We will probably need to revisit this once we support TLS/TCP edges in the gateway

TheConcierge: acutally...
what do you think about representing the TrafficPolicy within the gateway api stuff as a `json.RawMessage` as well?

The upside is that we don't have to update the gateway api stuff every time there is a new phase. 
The downside is it's quite a bit messier, implementation wise. When resolving all the filters, we'd likely have to represent the policy document as `map[string]any` and dynamically merge the top-level keys.

TheConcierge: Ok, I actually went ahead and made this change. It decouples us from the deprecated Module definitions and (hopefully) lessens our future maintenance burden.

internal/controller/ingress/httpsedge_controller.go
TheConcierge: Talked to Stacks about this. We are going to use `PolicyDeprecation` as the event reason here. That way, we can have users filter on that in their events when we actually need them to start fixing this stuff. It also aligns with how k8s does their event reasons.

internal/controller/ngrok/ngroktrafficpolicy_controller.go
```diff
@@ -63,7 +66,25 @@ type NgrokTrafficPolicyReconciler struct {
...
+
+	if parsedTrafficPolicy.IsLegacyPolicy() {
+		r.Recorder.Eventf(policy, v1.EventTypeWarning, events.PolicyDeprecation, "Traffic Policy is using legacy directions: ['inbound', 'outbound']. Update to new phases: ['on_tcp_connect', 'on_http_request', 'on_http_response']")
+	}
+
+	if parsedTrafficPolicy.Enabled() != nil {
+		r.Recorder.Eventf(policy, v1.EventTypeWarning, events.PolicyDeprecation, "Traffic Policy has 'enabled' set. This is a legacy option that will stop being supported soon.")
+	}
+
+	err = r.Driver.SyncEdges(ctx, r.Client)
```
TheConcierge, line 87: we may want to consider moving the recorder stuff to after the driver syncs. at the very least, not error out if the parsing, etc stuff doesn't work for the purposes of events. If this event stuff fails, recorder functions will also probably certainly fail. however, we may want to not take a risk there.

internal/ngrokapi/edge_modules_https.go
```diff
@@ -57,57 +50,13 @@ func (c *defaultHTTPSEdgeModulesClientset) TLSTermination() *https_edge_tls_term
...
-}
-
 type HTTPSEdgeRouteModulesClientset interface {
 	Backend() *https_edge_route_backend.Client
 	CircuitBreaker() *https_edge_route_circuit_breaker.Client
 	Compression() *https_edge_route_compression.Client
 	IPRestriction() *https_edge_route_ip_restriction.Client
 	OAuth() *https_edge_route_oauth.Client
-	Policy() *https_edge_route_policy.Client
-	RawPolicy() *rawHTTPSPolicyClient
```
TheConcierge, line 110: `RawPolicy` stuff was ripped out because we can't just shove the configured values directly into the API client anymore. It makes it cleaner to just use the provided client than to try to retrofit this.

internal/controller/ngrok/ngroktrafficpolicy_controller.go
```diff
@@ -63,7 +66,25 @@ type NgrokTrafficPolicyReconciler struct {
...
 
-	err := r.Driver.SyncEdges(ctx, r.Client)
+	policy := &ngrokv1alpha1.NgrokTrafficPolicy{}
+	if err := r.Get(ctx, req.NamespacedName, policy); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	parsedTrafficPolicy, err := util.NewTrafficPolicyFromJson(policy.Spec.Policy)
+	if err != nil {
+		return ctrl.Result{}, err
```
jonstacks, line 76: We might also want to record an event here if we can't parse the policy for whatever reason. Probably also in the other places we unmarshal the policy from json as well.

internal/ngrokapi/clientset.go
```diff
@@ -45,7 +44,7 @@ type DefaultClientset struct {
 
 // NewClientSet creates a new ClientSet from an ngrok client config.
 func NewClientSet(config *ngrok.ClientConfig) *DefaultClientset {
-	v6Config := &v6.ClientConfig{
+	v6Config := &ngrok.ClientConfig{
```
jonstacks, line 47: **Non-blocking**: I think this `v6config` can be removed and we just pass `config` to the `kubernetesOperatorClient`. We had `v6config` in here while we were supporting both `v5` and `v6`, but now with the awesome work you did to remove v5, we can remove it.

jonstacks (approved):

jonstacks (approved):

PR #457, opened 2024-10-21
Register operator

jonstacks:
## What

Registers the the Kubernetes Operator with the ngrok API. If bindings are enabled, it generates the TLS secret with a private key, creates a CSR as part of registration, and saves the returned certificate in the secret.

## How

* Consolidate the binding configuration and TLS secret reconcilers into the KubernetesOperator reconciler.

## Breaking Changes
Yes, some of the existing CRDs are being replaced with this change. We haven't released these yet, so unless someone is running on the latest `main`, they won't be impacted.


internal/controller/bindings/endpointbinding_poller.go
```diff
@@ -154,7 +154,7 @@ func (r *EndpointBindingPoller) createBinding(ctx context.Context, hashedName st
 	return nil
 }
 
-func (r *EndpointBindingPoller) updateBinding(ctx context.Context, binding *bindingsv1alpha1.EndpointBinding, apiEndpoint *EndpointBinding, urlBits *URLBits) error {
+func (r *EndpointBindingPoller) updateBinding(ctx context.Context, binding *bindingsv1alpha1.EndpointBinding, _ *EndpointBinding, urlBits *URLBits) error {
```
hjkatz, line 157: Heads up we might conflict on this change

internal/controller/ngrok/kubernetesoperator_controller.go
```diff
@@ -0,0 +1,453 @@
...
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/reconcile
+func (r *KubernetesOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	return r.controller.Reconcile(ctx, req, new(ngrokv1alpha1.KubernetesOperator))
+}
+
+func (r *KubernetesOperatorReconciler) create(ctx context.Context, ko *ngrokv1alpha1.KubernetesOperator) error {
```
hjkatz, line 124: I see that the API registration is happening in the reconcile loop rather then on startup. I think this changes the expected behaviour of the ngrok-op where previously we would not want the container to ready up and we do not want bindingconfig to be reconciled unless the ngrok-op can claim/register with the API first.

hjkatz (approved):

internal/controller/bindings/endpointbinding_poller.go
jonstacks: I can drop it out that commit easily if needed.

jonstacks:
I need to fix this before its ready. I thought it was but I discovered that we can't create the KubernetesOperator CR in the helm chart like I thought we could. I'll be moving this back to a similar pattern like we had above where it is created as part of the api-manger's main.

PR #465, opened 2024-10-24
Bindings forwarder

jonstacks:
## What

This adds in a new bindings forwarder deployment. The bindings forwarder watches for `bindingsv1alpha1.EndpointBindings` CRs and listens on the desired port so that incoming connections destined for the service have something to connect to. We still need to wire this up to connect to ngrok and forward the bytes there so that the endpoint is fully bound. For now, we simply use an example connection handler that prints information about the connection and echos back what the user inputs.

## How

* Added the initial bindings forwarder program, `bindings-forwarder-manager`. This sets up another manager and makes a new non leader elected controller with the `ForwarderReconciler`. The `ForwarderReconcile` listens starts and stop listening on the configured ports as needed.
* Added an example ConnectionHandler. We'll change this later, but for now it'll help us test
* Added in a new Deployment, Role, Rolebinding, and (optional) service account to the helm chart

## Breaking Changes
No, this is all hidden behind the helm values `bindings.enabled` which defaults to `false`.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -472,12 +472,14 @@ func enableGatewayFeatureSet(_ context.Context, _ managerOpts, mgr ctrl.Manager,
 func enableBindingsFeatureSet(_ context.Context, opts managerOpts, mgr ctrl.Manager, _ *store.Driver, _ ngrokapi.Clientset) error {
 	// EndpointBindings
 	if err := (&bindingscontroller.EndpointBindingReconciler{
-		Client:             mgr.GetClient(),
-		Scheme:             mgr.GetScheme(),
-		Log:                ctrl.Log.WithName("controllers").WithName("EndpointBinding"),
-		Recorder:           mgr.GetEventRecorderFor("bindings-controller"),
-		ClusterDomain:      opts.clusterDomain,
-		PodForwarderLabels: []string{}, // TODO(hkatz) Implement me
+		Client:        mgr.GetClient(),
+		Scheme:        mgr.GetScheme(),
+		Log:           ctrl.Log.WithName("controllers").WithName("EndpointBinding"),
+		Recorder:      mgr.GetEventRecorderFor("bindings-controller"),
+		ClusterDomain: opts.clusterDomain,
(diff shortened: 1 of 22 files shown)


cmd/api/main.go
```diff
@@ -472,12 +472,14 @@ func enableGatewayFeatureSet(_ context.Context, _ managerOpts, mgr ctrl.Manager,
...
-		ClusterDomain:      opts.clusterDomain,
-		PodForwarderLabels: []string{}, // TODO(hkatz) Implement me
+		Client:        mgr.GetClient(),
+		Scheme:        mgr.GetScheme(),
+		Log:           ctrl.Log.WithName("controllers").WithName("EndpointBinding"),
+		Recorder:      mgr.GetEventRecorderFor("bindings-controller"),
+		ClusterDomain: opts.clusterDomain,
+		UpstreamServiceLabelSelector: map[string]string{
+			"app.kubernetes.io/component": "bindings-forwarder",
+		}, // TODO(hkatz) Implement me
```
hjkatz, line 482: Nit: Remove todo?

cmd/bindings-forwarder/main.go
```diff
@@ -0,0 +1,161 @@
...
+	mgr, err := ctrl.NewManager(k8sConfig, options)
+	if err != nil {
+		return fmt.Errorf("unable to start bindings-forwarder-manager: %w", err)
+	}
+
+	bd := bindingsdriver.New()
+
+	if err = (&bindingscontroller.ForwarderReconciler{
+		Client:         mgr.GetClient(),
+		Log:            ctrl.Log.WithName("controllers").WithName("tunnel"),
```
hjkatz, line 138: Nit: Is this a tunnel controller or some other kind?

hjkatz (approved):

cmd/bindings-forwarder/main.go
jonstacks: Looks like a copypasta mistake. I'll update it

cmd/api/main.go
jonstacks: Will fix

PR #467, opened 2024-10-24
Add status to KubernetesOperator

hjkatz:
## What

Update status fields

Add basic registration status handling to KubernetesOperator

## How
*Describe the solution*

Use similar pattern to #464

## Breaking Changes
No.

### DIFF
--- a/api/ngrok/v1alpha1/kubernetesoperator_types.go
+++ b/api/ngrok/v1alpha1/kubernetesoperator_types.go
@@ -137,7 +137,6 @@ type KubernetesOperatorSpec struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="ID",type=string,JSONPath=`.status.id`,description="Kubernetes Operator ID"
-// +kubebuilder:printcolumn:name="Name",type="string",JSONPath=".metadata.name"
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.registrationStatus"
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`,description="Age"
 
--- a/api/ngrok/v1alpha1/kubernetesoperator_types.go
+++ b/api/ngrok/v1alpha1/kubernetesoperator_types.go
@@ -80,7 +80,29 @@ type KubernetesOperatorStatus struct {
 
 	// URI is the URI for this Kubernetes Operator
 	URI string `json:"uri,omitempty"`
+
+	// RegistrationStatus is the status of the registration of this Kubernetes Operator with the ngrok API
+	// +kubebuilder:validation:Required
+	// +kube:validation:Enum=registered;error;unknown
+	// +kubebuilder:default="unknown"
+	RegistrationStatus string `json:"registrationStatus,omitempty"`
+
+	// RegistrationErrorCode is the returned ngrok error code
+	// +kube:validation:Optional
+	// +kubebuilder:validation:Pattern=`^NGROK_ERR_\d+$`
+	RegistrationErrorCode string `json:"registrationErrorCode,omitempty"`
+
+	// RegistrationErrorMessage is a free-form error message if the status is error
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:MaxLength=4096
+	RegistrationErrorMessage string `json:"errorMessage,omitempty"`
 }
+
+const (
+	KubernetesOperatorRegistrationStatusSuccess = "registered"
+	KubernetesOperatorRegistrationStatusError   = "error"
+	KubernetesOperatorRegistrationStatusUnknown = "unknown"
+)
 
 const (
 	KubernetesOperatorFeatureIngress  = "ingress"
@@ -115,6 +137,8 @@ type KubernetesOperatorSpec struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="ID",type=string,JSONPath=`.status.id`,description="Kubernetes Operator ID"
+// +kubebuilder:printcolumn:name="Name",type="string",JSONPath=".metadata.name"
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.registrationStatus"
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`,description="Age"
 
 // KubernetesOperator is the Schema for the ngrok kubernetesoperators API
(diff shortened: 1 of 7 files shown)


hjkatz:
* **#467** <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-operator/467?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

api/ngrok/v1alpha1/kubernetesoperator_types.go
```diff
@@ -115,6 +137,8 @@ type KubernetesOperatorSpec struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="ID",type=string,JSONPath=`.status.id`,description="Kubernetes Operator ID"
+// +kubebuilder:printcolumn:name="Name",type="string",JSONPath=".metadata.name"
```
jonstacks, line 140: We can remove `Name` here. It is already in the output by default and adding it causes it to be printed in the output twice.

internal/controller/ngrok/kubernetesoperator_controller.go
```diff
@@ -51,6 +51,7 @@ import (
 	ngrokv1alpha1 "github.com/ngrok/ngrok-operator/api/ngrok/v1alpha1"
 	"github.com/ngrok/ngrok-operator/internal/controller"
 	"github.com/ngrok/ngrok-operator/internal/ngrokapi"
+	ngrokgo "golang.ngrok.com/ngrok"
```
jonstacks, line 54:
```suggestion
        "github.com/ngrok/ngrok-api-go/v6"
```
The library you are pulling in is for the [ngrok go sdk](https://github.com/ngrok/ngrok-go). We want the [ngrok-api-go](https://github.com/ngrok/ngrok-api-go) lib here for checking API errors.

internal/controller/ngrok/kubernetesoperator_controller.go
```diff
@@ -160,11 +167,11 @@ func (r *KubernetesOperatorReconciler) create(ctx context.Context, ko *ngrokv1al
...
-		return err
-	}
-
-	// Update the status with the KubernetesOperator ID
-	if err := r.updateStatus(ctx, ko, k8sOp); err != nil {
+		if errors.As(err, &ngrokErr) {
+			// nothing to do?
+		} else {
+			ngrokErr = nil // some other error
+		}
```
jonstacks, line 174: This doesn't look like it is adding any additional error checking. Our current pattern has just been to return ngrok errors like this and let the base controller figure out what to do. [Here is the common logic for that](https://github.com/ngrok/ngrok-operator/blob/main/internal/controller/base_controller.go#L132-L155). If we have specific error codes that we need to treat differently in this reconciler,  we also have a pattern for reconciler specific behavior([Example](https://github.com/ngrok/ngrok-operator/blob/main/internal/controller/ingress/httpsedge_controller.go#L87-L94) in the `HTTPSEdgeReconciler`).

internal/controller/ngrok/kubernetesoperator_controller.go
```diff
@@ -300,7 +339,8 @@ func (r *KubernetesOperatorReconciler) findExisting(ctx context.Context, ko *ngr
 		iterLogger.V(3).Info("found matching KubernetesOperator")
 		return item, nil
 	}
-	return nil, nil
+
+	return nil, iter.Err()
```
jonstacks, line 343: Good catch on the `iter.Err()` needing to be returned.

internal/controller/ngrok/kubernetesoperator_controller.go
```diff
@@ -116,17 +117,23 @@ func (r *KubernetesOperatorReconciler) Reconcile(ctx context.Context, req ctrl.R
...
 	if err != nil {
 		return err
 	}
 
+	// defer status update
+	defer func() {
+		// best effort
+		_ = r.updateStatus(ctx, ko, k8sOp, ngrokErr)
+	}()
+
```
jonstacks, line 133: I've thought about this a bit more and can we hold off on using this new defer pattern for updating the statuses?

api/ngrok/v1alpha1/kubernetesoperator_types.go
```diff
@@ -80,8 +80,30 @@ type KubernetesOperatorStatus struct {
 
 	// URI is the URI for this Kubernetes Operator
 	URI string `json:"uri,omitempty"`
+
+	// RegistrationStatus is the status of the registration of this Kubernetes Operator with the ngrok API
+	// +kubebuilder:validation:Required
+	// +kube:validation:Enum=registered;error;unknown
```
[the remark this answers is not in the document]
hjkatz: I think that makes sense. I always tend to create an unknown state for any status/state enums as a rule of thumb.

But, I haven't found a use for unknown status in either this code or the endpointbinding status code. So I think changing this makes sense 👍

internal/controller/ngrok/kubernetesoperator_controller.go
hjkatz: So I asked about this here and I'm confused about which errors we care about. 

https://ngrok.slack.com/archives/C057633V0E7/p1729800605168539?thread_ts=1729800021.427839&cid=C057633V0E7

I think we should chat through this in a huddle so I can grok our ngrok errors :P

internal/controller/ngrok/kubernetesoperator_controller.go
hjkatz: Yes, and I want to do that with both status PRs merged so I can have 1 PR to overhaul the logic.

internal/controller/ngrok/kubernetesoperator_controller.go
hjkatz: 1. Let's chat about this offline b/c I need some understanding dropped on me wrt ngrok errors
2. It's not adding additional checking - it's doing a cast to `ngrok.Error` so the defer has the proper filled type

api/ngrok/v1alpha1/kubernetesoperator_types.go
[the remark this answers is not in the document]
hjkatz: I decided to go with `pending`

internal/controller/ngrok/kubernetesoperator_controller.go
hjkatz: Chatted offline - I will implement ngrok error handling in a follow-up PR

internal/controller/base_controller.go
```diff
@@ -136,6 +153,22 @@ func (self *BaseController[T]) handleErr(op BaseControllerOp, obj T, err error)
 	return CtrlResultForErr(err)
 }
 
+// ReconcileStatus is a helper function to reconcile the status of an object and requeue on update errors
+func (self *BaseController[T]) ReconcileStatus(ctx context.Context, obj T) (ctrl.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	patch := client.MergeFrom(obj)
+	if err := self.Kube.Status().Patch(ctx, obj, patch); err != nil {
```
jonstacks, line 161: I think this will always end up creating an empty patch. The patch needs to be created from the object status in its current state and them modified before calling `.Status().Patch`.

https://sdk.operatorframework.io/docs/building-operators/golang/references/client/#patch

PR #472, opened 2024-10-28
Generate and commit schema.json file

alex-bezek:
## What

We already use the readme generator tool, lets have it also generate a schema.json file and commit it. 

## How

We are already generating the file and committing the readme.md file change, just also change where we write this file to and commit it after.

## Breaking Changes

None


helm/ngrok-operator/values.yaml
```diff
@@ -120,13 +120,13 @@ lifecycle: {}
...
-## @param podDisruptionBudget.maxUnavailable Maximum number/percentage of pods that may be made unavailable
+## @param podDisruptionBudget.maxUnavailable [string] Maximum number/percentage of pods that may be made unavailable
+## @param podDisruptionBudget.minAvailable [string] Minimum number/percentage of pods that should remain scheduled
 ##
 podDisruptionBudget:
   create: false
-  minAvailable: ""
-  maxUnavailable: 1
+  maxUnavailable: "1"
+  # minAvailable:
```
alex-bezek, line 129: This is technically non-passive, but i also wonder if this is what we want for our defaults. This seems like we copied what we have internally where we favor safety and only allow 1 down at a time, but I could see the default being to just ensure you have at least 1 up at a time. Looking at other charts, they are even less perscriptive more similar to how charts don't default replicaCount or resources to specific numbers

jonstacks (approved):

hjkatz (approved):

helm/ngrok-operator/values.yaml
alex-bezek: hmm, i guess I just meant since it was changing from a number to a string. But I guess we didn't have a schema before so its "passive" :)

PR #474, opened 2024-10-29
Add clusterId and checks for KubernetesOperators registration

hjkatz:
## What

During k8sop registration we try to match the k8sop in ngrok API with the configured KubernetesOperator in ngrok-operator.

Currently we're matching on:
* operator name
* operator namespace
* operator namespace UID
* operator binding name (if exists)

But, the operator name and namespace may be the same (likely) in different clusters, however we do not have a field to check for cluster name/id.

So this PR tentatively adds that feature (but doesn't implement the ngrok API side of things just yet).

## How
*Describe the solution*

## Breaking Changes

Yes, `.Values.clusterId` is now a required field.


hjkatz:
* **#474** <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-operator/474?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

helm/ngrok-operator/templates/controller-deployment.yaml
```diff
@@ -62,6 +62,7 @@ spec:
         - /api-manager
         args:
         - --release-name={{ .Release.Name }}
+        - --cluster-id={{ .Values.clusterId | required "Missing required .Values.clusterId!"}}
```
alex-bezek, line 65: I think we need a way to have this not be supplied for an initial install. The requirements we talked about in slack are detained a bit better by this old issue where we resolved the issue that we required an authtoken and API key on start https://github.com/ngrok/ngrok-operator/issues/317

hjkatz: I'm reading from that issue that the desire is the ngrok-operator can be installed into a non-functional/non-reconciling state and we give an error event on the `kind: KubernetesOperator` itself.

Is that what you mean?

alex-bezek: yep, i think something like that would fufill the requirements. Similarly, if you one click install, the operator is just going to log and event that it needs the credentias I believe. The idea is a helm instal without custom options should "just work" to at least helm install, but it may not fully function.

PR #476, opened 2024-10-29
Bindings forwarder implementation

jonstacks:
## What

This adds a binding forwarder implementation, so that when a client connects to the given port, we open a connection to the bindings ingress endpoint, passing along the header, and then joining the net.Conns together.

## How

* In order to send the header information to mux, we pull in pg_agent so we have the definition for ConnRequest and ConnResponse. In the future, this will most likely be included as part of ngrok-go IMO.
* Added some additional printer columns to `KubernetesOperator`. This now gives additional columns when doing `-o wide`. Ex: `kubectl get kubernetesoperators.ngrok.k8s.ngrok.com -A -o wide`
* Swap out the example handler for one that dials the ingress endpoint using the cert from the KubernetesOperator, exchanges the binding header, and then joins the connections together.

## Breaking Changes
No


api/ngrok/v1alpha1/kubernetesoperator_types.go
```diff
@@ -138,6 +142,9 @@ type KubernetesOperatorSpec struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="ID",type=string,JSONPath=`.status.id`,description="Kubernetes Operator ID"
 // +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.registrationStatus"
+// +kubebuilder:printcolumn:name="Enabled Features",type="string",JSONPath=".status.enabledFeatures"
+// +kubebuilder:printcolumn:name="Binding Name", type="string", JSONPath=".spec.binding.name",priority=2
+// +kubebuilder:printcolumn:name="Binding Ingress Endpoint", type="string", JSONPath=".spec.binding.ingressEndpoint",priority=2
```
hjkatz, line 147: TIL: `priority=2` very cool

cmd/api/main.go
```diff
@@ -149,6 +150,7 @@ func cmd() *cobra.Command {
 	c.Flags().BoolVar(&opts.enableFeatureBindings, "enable-feature-bindings", false, "Enables the Endpoint Bindings controller")
 	c.Flags().StringSliceVar(&opts.bindings.allowedURLs, "bindings-allowed-urls", []string{"*"}, "Allowed URLs for Endpoint Bindings")
 	c.Flags().StringVar(&opts.bindings.name, "bindings-name", "default", "Name of the Endpoint Binding Configuration")
+	c.Flags().StringVar(&opts.bindings.ingressEndpoint, "bindings-ingress-endpoint", "", "The endpoint the bindings forwarder connects to")
```
hjkatz, line 153: So I think this field is actually returned server side from the k8sop registration request, see: https://github.com/ngrok-private/ngrok/blob/main/proto/api/kubernetes_operator.proto#L214

internal/controller/bindings/forwarder_controller.go
```diff
@@ -112,11 +120,112 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
 		"port", epb.Spec.Port,
 	)
 
+	// Get the KubernetesOperator
+
+	operators := ngrokv1alpha1.KubernetesOperatorList{}
+	if err := r.Client.List(ctx, &operators); err != nil {
+		log.Error(err, "failed to list KubernetesOperators")
+		return err
+	}
```
hjkatz, line 129: Q. Why list the operators here rather than used the defined `managerOpts.releaseName` to find the associated `kind: KubernetesOperator`?

internal/controller/bindings/forwarder_controller.go
```diff
@@ -112,11 +120,112 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
...
+		},
+	}
+
+	endpointURI, err := url.Parse(epb.Spec.EndpointURI)
+	if err != nil {
+		return err
+	}
+
+	host := endpointURI.Hostname()
+	port, err := strconv.Atoi(endpointURI.Port())
```
hjkatz, line 180: OOC: Why `Atoi` and then later `Itoa` for the port?

internal/controller/bindings/forwarder_controller.go
```diff
@@ -112,11 +120,112 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
...
+		log = log.WithValues("endpoint.id", resp.EndpointID, "proto", resp.Proto)
+		if err != nil {
+			log.Error(err, "failed to upgrade connection")
+			return err
+		}
+
+		if resp.ErrorCode != "" || resp.ErrorMessage != "" {
+			err := fmt.Errorf("%s: %s", resp.ErrorCode, resp.ErrorMessage)
+			log.Error(err, "failed to upgrade connection", "errorCode", resp.ErrorCode, "errorMessage", resp.ErrorMessage)
+			return err
```
hjkatz, line 218: OOC: Does returning errors from the forwarder lead to CrashLoopBackoff or a NotReady state via its health checks?

internal/controller/bindings/forwarder_controller.go
```diff
@@ -112,11 +120,112 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
...
+		}
+
+		log.Info("Bound connection")
+		return joinConnections(log, conn, ngrokConn)
+	}
+
 	log.Info("Listening on port")
-	port := int32(epb.Spec.Port)
 
 	// TODO: wire this up to ngrok. For now, we'll just use the example connection handler
```
hjkatz, line 227: Nit: TODO can be removed, yes?

internal/controller/bindings/forwarder_controller.go
```diff
@@ -132,40 +241,27 @@ func (r *ForwarderReconciler) statusID(epb *bindingsv1alpha1.BoundEndpoint) stri
...
 			}
-		}
-		return nil
-	}
+		}()
+
+		_, err := io.Copy(conn2, conn1)
+		return err
+	})
+	return g.Wait()
```
hjkatz, line 266: Q. If the connection is interrupted or closed unexpectedly, what is the behaviour of the pod forwarder to re-dial?

internal/controller/ngrok/kubernetesoperator_controller.go
```diff
@@ -207,6 +208,7 @@ func (r *KubernetesOperatorReconciler) updateStatus(ctx context.Context, ko *ngr
 		ko.Status.RegistrationStatus = ngrokv1alpha1.KubernetesOperatorRegistrationStatusError
 		ko.Status.RegistrationErrorCode = errorCode
 		ko.Status.RegistrationErrorMessage = errorMessage
+		ko.Status.EnabledFeatures = strings.Join(ngrokKo.EnabledFeatures, ",")
```
hjkatz, line 211: OOC: Why include the enabled features on status as well as spec?

See: https://github.com/ngrok/ngrok-operator/blob/main/api/ngrok/v1alpha1/kubernetesoperator_types.go#L122

hjkatz (approved):

internal/controller/bindings/forwarder_controller.go
jonstacks: The forwarder only dials on the initial connection. If the upstream connection to mux is closed, it will disconnect the in-cluster client and they will have to re-dial.

hjkatz: Who has to re-dial? This would be the forwarder needing to re-open the tcp connection, no?

internal/controller/ngrok/kubernetesoperator_controller.go
jonstacks: This is so the additionalPrinterColumn is properly formatted. I ran into https://github.com/kubernetes-sigs/controller-tools/issues/652 and this seems to be the accepted solution on how to accomplish this. I think this also has another benefit in that it will show the true status of the enabled features as seen by ngrok.

internal/controller/bindings/forwarder_controller.go
jonstacks: Not currently. The challenge I see is that if a single port forwarder is returning errors, but the rest are fine, what do we do? Since our services all use the same selectors for targeting the forwarder pods, marking them not ready if just one forwarder has an issue would lead to knocking all port forwarders offline

cmd/api/main.go
jonstacks: Talked offline: It is, but we don't have defaulting for it yet in the API and there may be the case that we need to whitelabel and use a custom domain, so we'll leave it for now. Once the API returns this and supports custom overrides, we'll drop this. Probably also the agent `serverAddr` at the same time as well.

internal/controller/bindings/forwarder_controller.go
jonstacks: I had started working on this before managerOpts.releaseName was included in the API manager. We don't have it plumbed through yet for the bindings-forwarder. I will make that change though.

internal/controller/bindings/forwarder_controller.go
jonstacks: Just what I had for local variables. I'll switch it to capture an intermediate string var for the port

internal/controller/bindings/forwarder_controller.go
jonstacks: The client of bindings-forwarder would need to try to re-connect to in-cluster service. If the connection to the ingress endpoint is interrupted or closed, the connection to the in-cluster client is no longer usable.

internal/controller/bindings/forwarder_controller.go
jonstacks: Fixed

internal/controller/bindings/forwarder_controller.go
jonstacks: Fixed. Removed the Itoa below

internal/controller/bindings/forwarder_controller.go
jonstacks: Fixed

internal/controller/bindings/forwarder_controller.go
hjkatz: I would think that having 1 Pod marked as NotReady would tell k8s Service to update the EndpointSlice which would then have our client services not talk to the Pods that are NotReady.

I think that this is what we want, no?

internal/controller/bindings/forwarder_controller.go
hjkatz: Chatted through this in Slack

internal/controller/bindings/forwarder_controller.go
hjkatz: We chatted in Slack about this. This is a hard problem to solve because we don't want 1 BoundEndpoint's port to take the whole Pod offline because that would affect other ports in that Pod too.

Instead we talked about ways we could check for statuses of individual Pod IPs via headless Services, but came to the conclusion that because we're relying on k8s' service network routing that we may not be able to solve for this exact case using the proxy forwarder implementation.

So we will come back to this question post launch.

hjkatz (approved):

PR #481, opened 2024-10-31
feat: BoundEndpointPoller polls from the API

jonstacks:
## What

This replaces the mock endpoints we had for testing with actually polling the API for bound endpoints.

## How

* Pass the ngrokclientset into the poller
* Replace the mock endpoints we had with the results returned from the API

## Breaking Changes
No

### DIFF
--- a/internal/controller/bindings/boundendpoint_poller.go
+++ b/internal/controller/bindings/boundendpoint_poller.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
+	"github.com/ngrok/ngrok-api-go/v6"
 	v6 "github.com/ngrok/ngrok-api-go/v6"
 	bindingsv1alpha1 "github.com/ngrok/ngrok-operator/api/bindings/v1alpha1"
 	ngrokv1alpha1 "github.com/ngrok/ngrok-operator/api/ngrok/v1alpha1"
@@ -43,6 +44,9 @@ type BoundEndpointPoller struct {
 
 	// KubernetesOperatorConfigName is the expected name of the KubernetesOperator that we should poll
 	KubernetesOperatorConfigName string
+
+	// NgrokClientset is the ngrok API clientset
+	NgrokClientset ngrokapi.Clientset
 
 	// PollingInterval is how often to poll the ngrok API for reconciling the BindingEndpoints
 	PollingInterval time.Duration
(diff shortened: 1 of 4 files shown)


internal/controller/bindings/boundendpoint_poller.go
```diff
@@ -163,20 +167,26 @@ func (r *BoundEndpointPoller) reconcileBoundEndpointsFromAPI(ctx context.Context
...
+	var apiBindingEndpoints []v6.Endpoint
+	iter := r.NgrokClientset.KubernetesOperators().GetBoundEndpoints(r.koId, &ngrok.Paging{})
+	for iter.Next(ctx) {
+		item := iter.Item()
+		if item != nil {
+			apiBindingEndpoints = append(apiBindingEndpoints, *item)
+		}
+	}
+
+	err := iter.Err()
```
hjkatz, line 184: Nice plug and play!

I am curious on what you think we should do on errors here. If they are client errors then we won't get a retry from the base controller, but if it's other errors then we should get retries with backoff, right?

I just want to make sure that if the API is down/broken for 1 polling interval that we don't necesarrily wait the whole polling interval to try again. Wdyt?

hjkatz (approved):

natasha-jarus (approved):

internal/controller/bindings/boundendpoint_poller.go
alex-bezek: Should we make this logic re-usable for controllers not extending the base controller?
https://github.com/ngrok/ngrok-operator/blob/4768485b14e6b31b1be0898a11e1cea4a29cc011/internal/controller/base_controller.go#L156-L168

hjkatz: Isn't that what this code is doing? We're not checking the specific error code here, rather relying on the base controller to check and issue a retry.

alex-bezek: I thought that was your point about how _if there are client errors then we won't get a retry from the base controller_. This controller doesn't extend the base controller so it doesn't get retries and the RequeueAfter

jonstacks: Since this runnable doesn't use reconciles, I think `CtrlResultForError()` won't help us as Alex mentioned since we won't be able to do anything with `reconcile.Result`. If there happens to be an API issue, it will retry on the next polling interval, which I think is fine for handling this for now. If we want to implement exp backoff for this like our API Reconcilers get by virtue of being Reconcilers, I think that is fine

hjkatz: I'll file a ticket to brainstorm this behaviour for now. 👍

PR #488, opened 2024-11-06
fix: domain stuck when ID is not found

jonstacks:
## What

Fixes an issue where we would not retry to reconcile domains when receiving a 404 from the ngrok API. This could happen when the resource was deleted either in the dashboard or via the API outside of the ngrok operator.

## How

* If a domain can't be found in the ngrok API, clear the status and return the original not found error
* Also, update our base controller's error handling to retry on not found errors. This was probably affecting all resources that were clearing their status but would not be re-reconciled because we were swallowing them in `CtrlResultForErr`.

## Breaking Changes

No

### DIFF
--- a/internal/controller/ingress/domain_controller.go
+++ b/internal/controller/ingress/domain_controller.go
@@ -137,6 +137,12 @@ func (r *DomainReconciler) create(ctx context.Context, domain *ingressv1alpha1.D
 func (r *DomainReconciler) update(ctx context.Context, domain *ingressv1alpha1.Domain) error {
 	resp, err := r.DomainsClient.Get(ctx, domain.Status.ID)
 	if err != nil {
+		// If the domain is gone, clear the status and trigger a re-reconcile
+		if ngrok.IsNotFound(err) {
+			domain.Status = ingressv1alpha1.DomainStatus{}
+			return r.controller.ReconcileStatus(ctx, domain, err)
+		}
+
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


internal/controller/ingress/domain_controller.go
```diff
@@ -137,6 +137,12 @@ func (r *DomainReconciler) create(ctx context.Context, domain *ingressv1alpha1.D
 func (r *DomainReconciler) update(ctx context.Context, domain *ingressv1alpha1.Domain) error {
 	resp, err := r.DomainsClient.Get(ctx, domain.Status.ID)
 	if err != nil {
+		// If the domain is gone, clear the status and trigger a re-reconcile
+		if ngrok.IsNotFound(err) {
+			domain.Status = ingressv1alpha1.DomainStatus{}
+			return r.controller.ReconcileStatus(ctx, domain, err)
```
hjkatz, line 143: Q. Do we want to do the regular exponential backoff here or should we set our own backoff time such as `3 * time.Minute`?

hjkatz (approved):

internal/controller/ingress/domain_controller.go
jonstacks: I think the regular exp backoff should be fine. I checked and it caps it at 1000 seconds. All the reconcilers should be clearing the status ID on 404s and then it should go through the create flow. If there is a persistent problem updating the Status on the CR coming from k8s, I'm not sure if there is much we can do.

PR #489, opened 2024-11-06
Use unique context for endpoint poller reconcile actions

hjkatz:
## What

Use a unique context so we can control the cancel rather than the ctrl manager.

## How
*Describe the solution*

## Breaking Changes

### DIFF
--- a/internal/controller/bindings/boundendpoint_poller.go
+++ b/internal/controller/bindings/boundendpoint_poller.go
@@ -216,22 +216,23 @@ func (r *BoundEndpointPoller) reconcileBoundEndpointsFromAPI(ctx context.Context
 	toCreate, toUpdate, toDelete := r.filterBoundEndpointActions(ctx, existingBoundEndpoints, desiredBoundEndpoints)
 
 	// create context + errgroup for managing/closing the future goroutine in the reconcile actions loops
-	errGroup, ctx := errgroup.WithContext(ctx)
-	ctx, cancel := context.WithCancel(ctx)
+	errGroup, reconcileActionCtx := errgroup.WithContext(context.Background())
+	reconcileActionCtx, cancel := context.WithCancel(reconcileActionCtx)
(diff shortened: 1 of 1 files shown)


internal/controller/bindings/boundendpoint_poller.go
```diff
@@ -216,22 +216,23 @@ func (r *BoundEndpointPoller) reconcileBoundEndpointsFromAPI(ctx context.Context
...
+		return r.updateBinding(reconcileActionCtx, binding)
 	})
 
-	r.reconcileBoundEndpointAction(ctx, errGroup, toDelete, "delete", func(ctx context.Context, binding bindingsv1alpha1.BoundEndpoint) error {
-		return r.deleteBinding(ctx, binding)
+	r.reconcileBoundEndpointAction(reconcileActionCtx, errGroup, toDelete, "delete", func(reconcileActionCtx context.Context, binding bindingsv1alpha1.BoundEndpoint) error {
+		return r.deleteBinding(reconcileActionCtx, binding)
 	})
 
 	return nil
```
jonstacks, line 238: **Question** Do we need an `errGroup.Wait()` here or are we just using the errgroup to spawn all of them at the same time and relying on the context being canceled? If the latter, it might be clearer if we drop the errgroup and just use goroutines for  concurrent processing.

jonstacks (approved):

hjkatz:
### Merge activity

* **Nov 7, 5:44 AM EST**: [Graphite](https://app.graphite.dev/github/pr/ngrok/ngrok-operator/489) rebased this pull request as part of a merge.

internal/controller/bindings/boundendpoint_poller.go
hjkatz: No I think you're right - I'm mixing up using errGroup and managing the ctx cancel myself. I'll clean this up.

PR #490, opened 2024-11-07
Add Deployment.StartedAt

hjkatz:
## What

Add a StartedAt field to the KubernetesOperator CRD.

This helps us register the k8sop with the API on every restart
and allows users to see the last time the operator was started and registered.

## How
*Describe the solution*

## Breaking Changes


api/ngrok/v1alpha1/kubernetesoperator_types.go
```diff
@@ -48,6 +48,9 @@ type KubernetesOperatorDeployment struct {
 	Namespace string `json:"namespace,omitempty"`
 	// The version of the operator that is currently running
 	Version string `json:"version,omitempty"`
+
+	// StartedAt is the time at which the operator was last started
+	StartedAt *metav1.Time `json:"startedAt,omitempty"`
```
jonstacks, line 53: What do you think about just relying on `kubectl get pods` to see when it was last started?

I'm hesitant to add this to the spec for 2 reasons:
1. The spec defines the desired state of the system. Typically it doesn't define transitions like this and instead these are put into status conditions.
2. These map pretty closely what the ngrok API allows us to configure. The KubernetesOperator is somewhat of an exception with the `tlsSecretName`, but it is still part of the desired state for where the tls.key and tls.crt should live.

hjkatz:
Closing because this is fixed in #493

PR #493, opened 2024-11-07
fix: Make sure we update the status

jonstacks:
## What

While fixing the panics that we were getting, I introduced a bug in https://github.com/ngrok/ngrok-operator/pull/483 that would only update the status if bindings was enabled. We should update the status either way.

## How

Don't return nil, make sure we update the status. Also, we clearing the Status.ID on all errors. We probably don't want to do this in the case of network errors, so only explicitly clear the ID and URI on not found errors.

## Breaking Changes
No


hjkatz (approved):
Thank you for finding this!

PR #503, opened 2024-11-12
Support one-click-demo mode

hjkatz:
## What

We would like to support a 1-click installation for marketplace environments.
However, these environments do not allow users to provide required config before 1-click installing.
And, the installation is only successful if the Pod goes Ready (helm install success).
But, our ngrok-op requires user configuration to work, namely API key and authtoken.

So we provide this new `oneClickDemoMode: true/false` value that can change the behaviour of the installed ngrok-op.
- When true, ngrok-op will become Ready and log messages about missing required configuration.
- When false, ngrok-op will run in its normal operator mode (and report errors for missing fields as NotReady, etc...)

## How
*Describe the solution*

Factor out startup functions

Add oneClickDemoMode to helm values.yaml

## Breaking Changes
No.

### DIFF
--- a/helm/ngrok-operator/values.yaml
+++ b/helm/ngrok-operator/values.yaml
@@ -6,11 +6,13 @@
 ## @param description ngrok-operator description that will appear in the ngrok dashboard
 ## @param commonLabels Labels to add to all deployed objects
 ## @param commonAnnotations Annotations to add to all deployed objects
+## @param oneClickDemoMode If true, then the operator will startup without required fields or API registration, become Ready, but not actually be running
 nameOverride: ""
 fullnameOverride: ""
 description: "The official ngrok Kubernetes Operator."
 commonLabels: {}
 commonAnnotations: {}
+oneClickDemoMode: false
 
 ##
 ## @section Image configuration
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -29,6 +29,7 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
 
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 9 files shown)


hjkatz:
* **#503** <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-operator/503?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

helm/ngrok-operator/values.yaml
```diff
@@ -6,11 +6,13 @@
 ## @param description ngrok-operator description that will appear in the ngrok dashboard
 ## @param commonLabels Labels to add to all deployed objects
 ## @param commonAnnotations Annotations to add to all deployed objects
+## @param oneClickDemoMode If true, then the operator will startup without required fields or API registration, become Ready, but not actually be running
 nameOverride: ""
 fullnameOverride: ""
 description: "The official ngrok Kubernetes Operator."
 commonLabels: {}
 commonAnnotations: {}
+oneClickDemoMode: false
```
alex-bezek, line 15: doesn't this need to be set to true in order for it to work without having to set any required fields? Or do these on click app marketplaces allow specifying a helm value like this by default and its just we can't require user specific values to be set like API crednetials?

hjkatz: >  Or do these on click app marketplaces allow specifying a helm value like this by default and its just we can't require user specific values to be set like API crednetials?

The latter: My understanding is we can provide a `values.yaml` but not a pop-up for required fields. So we'll set `onClickDemoMode: true` in those marketplace values.

alex-bezek: ahh ok, thats good to know, thanks

alex-bezek: if we can provide default values i'm kinda surprised we didn't just default these installations' replica count to 0 and instruct you to scale up after provider credentials vs a no-op mode, but I don't think I have all the historic context here

alex-bezek (approved):

hjkatz:
### Merge activity

* **Nov 13, 11:01 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/ngrok/ngrok-operator/503).

PR #525, opened 2024-11-26
add agentendpoint crd

Alice-Lilith:
Initial draft of the agent endpoint CRD

Example validations on the traffic policy config:

```sh
k apply -f -<<EOF
apiVersion: ngrok.k8s.ngrok.com/v1alpha1
kind: AgentEndpoint
metadata:
  name: valid-inline-policy
spec:
  url: https://example.ngrok.app
  trafficPolicy:
    type: inline
    inline:
      action: "allow"
      rules: []
  description: "Valid inline traffic policy"
  metadata: '{"team":"devops"}'
  bindings:
    - public
    - k8s/namespace
EOF
agentendpoint.ngrok.k8s.ngrok.com/valid-inline-policy created
```

```sh
 k apply -f -<<EOF
apiVersion: ngrok.k8s.ngrok.com/v1alpha1
kind: AgentEndpoint
metadata:
  name: valid-inline-policy
spec:
  url: https://example.ngrok.app
  trafficPolicy:
    type: targetRef
    inline:
      action: "allow"
      rules: []
  description: "Valid inline traffic policy"
  metadata: '{"team":"devops"}'
  bindings:
    - public
    - k8s/namespace
EOF
The AgentEndpoint "valid-inline-policy" is invalid: spec.trafficPolicy: Invalid value: "object": When type is targetRef, targetRef must be set, and inline must not be set.
```


api/ngrok/v1alpha1/agentendpoint_types.go
```diff
@@ -0,0 +1,169 @@
...
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Items=pattern=`^(public|internal|k8s/.*)$`
+	Bindings []string `json:"bindings,omitempty"`
+}
+
+type TrafficPolicyCfgType string
+
+const (
+	TrafficPolicyCfgType_K8sRef TrafficPolicyCfgType = "targetRef"
```
Alice-Lilith, line 96: We currently support k8s object refs and inline for the traffic policy, but we might want to support other ways to supply one in the future such as configuring one in the ngrok cloud and having one or more clusters running the operator reference that one and have the definition pulled from the cloud.

api/ngrok/v1alpha1/agentendpoint_types.go
```diff
@@ -0,0 +1,169 @@
...
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Type=object
+	Inline json.RawMessage `json:"inline,omitempty"`
+
+	// Reference to a TrafficPolicy resource to attach to the Agent Endpoint
+	//
+	// +kubebuilder:validation:Optional
+	Reference *K8sObjectRef `json:"targetRef,omitempty"`
```
Alice-Lilith, line 130: Currently, this will take only a `name`, but lets us add in support for `namespace` in the future in a non-breaking way.

api/ngrok/v1alpha1/agentendpoint_types.go
```diff
@@ -0,0 +1,169 @@
...
+	Type TrafficPolicyCfgType `json:"type"`
+
+	// Inline definition of a TrafficPolicy to attach to the agent Endpoint
+	// The raw json encoded policy that was applied to the ngrok API
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +kubebuilder:validation:Type=object
+	Inline json.RawMessage `json:"inline,omitempty"`
```
Alice-Lilith, line 125: IMO Switching from raw json message to structural types would be a good idea, but for now I want to keep consistency with the other resources. We can discuss whether we want to change this prior to GA in a follow-up.

api/ngrok/v1alpha1/cloudendpoint_types.go
```diff
@@ -30,28 +30,45 @@ import (
...
 	// +kubebuilder:validation:Optional
 	TrafficPolicy *NgrokTrafficPolicySpec `json:"trafficPolicy,omitempty"`
 
-	// Description is a human-readable description of this cloud endpoint
+	// Human-readable description of this cloud endpoint
+	//
 	// +kubebuilder:default:=`Created by the ngrok-operator`
 	Description string `json:"description,omitempty"`
 
-	// Metadata is a string of arbitrary data associated with the object in the ngrok API/Dashboard
```
jonstacks, line 49: Just a heads up, you might see these doc string formats with the struct field as the first word, mostly because we would just copy them from the ngrok-api-go client docstrings and then just make the properties camelcase instead of snake case.

api/ngrok/v1alpha1/cloudendpoint_types.go
```diff
@@ -63,13 +80,16 @@ type CloudEndpointStatus struct {
 	ID string `json:"id,omitempty"`
 }
 
+// CloudEndpoint is the Schema for the cloudendpoints API
+//
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:resource:categories="networking";"ngrok"
+// +kubebuilder:resource:shortName=cep
```
jonstacks, line 88: Very nice on the shortName. I think internally we use the shorthand `clep` for cloud endpoints. Might want to get products thoughts on what this should be and make it consistent across our products.

api/ngrok/v1alpha1/agentendpoint_types.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+// AgentEndpoint is the Schema for the agentendpoints API
+//
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:categories="networking";"ngrok"
+// +kubebuilder:resource:shortName=aep
+// +kubebuilder:printcolumn:name="URL",type="string",JSONPath=".status.assignedURL"
+// +kubebuilder:printcolumn:name="Traffic Policy",type="string",JSONPath=".status.trafficPolicy"
```
jonstacks, line 40: We might want to set a priority on this if `.status.trafficPolicy` is the full traffic policy string so it only shows up with `-o wide`

Alice-Lilith: It would only display `"inline"`, `"none"`, or `"<name>"` to indicate where the traffic policy comes from, not including the whole raw string if it is inlined. Just to help give users debugging/editing a quick heads-up on where the policy comes from without needing to fully inspect/edit the AgentEndpoint.

Alice-Lilith: Removed for now since we aren't going to support the status yet

api/ngrok/v1alpha1/agentendpoint_types.go
```diff
@@ -0,0 +1,249 @@
...
+	// +kubebuilder:validation:Required
+	URL string `json:"url"`
+
+	// Specifies the protocol to use when connecting to the upstream. Currently only http1 and http2 are supported
+	// with prior knowledge (defaulting to http1). alpn negotiation is not currently supported.
+	//
+	// +kubebuilder:validation:Enum=http1;http2
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default=http1
+	Protocol UpstreamProtocol `json:"protocol"`
```
Alice-Lilith, line 118: My intention is that for now we only accept `http1`/`http2`, but in the future we can accept this as an alpn config like: `h2[, http/1.1]`

Alice-Lilith: Might have an additional helper field like `protocolIsAlpn` or something along those lines that is non-breaking.

jonstacks (approved):

api/ngrok/v1alpha1/agentendpoint_types.go
```diff
@@ -0,0 +1,249 @@
...
+	//         For https port will be443.
+	//         For http port will be 80.
+	//         Host will be localhost.
+	// Port (shorthand) - 8080
+	//     When using port you are defining the port on the local host that will receive traffic.
+	//         Scheme will default to http.
+	//         Host will default to localhost.
+	//
+	// +kubebuilder:validation:Required
+	URL string `json:"url"`
```
hjkatz, line 110: btw I have a function that parses out the public_url into a 4-tuple that you may find a use for in this work: https://github.com/ngrok/ngrok-operator/blob/main/internal/ngrokapi/bindingendpoint_aggregator.go#L92

PR #536, opened 2024-12-02
feat: Wire up AgentEndpoint CRD to ngrok-go

jonstacks:
## What

We now have an `AgentEndpoint` CRD, let's wire it up so that it actually creates those agent endpoints.

## How

* Upgrade the ngrok-go SDK b/c we need new features
* In the same way that we watch for `Tunnel` CRs in the tunnel driver, we can also watch for `AgentEndpoint` CRs and create the agent endpoints using the ngrok go SDK.
* Also allow for restarting the an agent pod by wiring it up to the pod's liveness probe. Clicking restart in the dashboard on a ngrok-go SDK will cause the pod to fail its liveness probes and restart after about 1 minute.

## Breaking Changes
No


jonstacks:
Included in https://github.com/ngrok/ngrok-operator/pull/538

PR #549, opened 2024-12-10
Fix http endpoint scheme

jonstacks:
## What

This fixes the agent endpoint getting created as a HTTPS endpoint even though the scheme in the url is HTTP.
 
## How
* Add some extra logging
* If the URL scheme is `http`, add the scheme opt so that the endpoint is created as a http endpoint instead of https. We may want to consider fixing this in the ngrok SDK going forward.

## Breaking Changes
No

### DIFF
--- a/pkg/tunneldriver/driver.go
+++ b/pkg/tunneldriver/driver.go
@@ -452,6 +452,12 @@ func (td *TunnelDriver) CreateAgentEndpoint(ctx context.Context, name string, sp
 			upstreamProto = string(*spec.Upstream.Protocol)
 		}
 		opts = append(opts, config.WithAppProtocol(upstreamProto))
+
+		// TODO: This should probably be inferred from the scheme in the URL
+		if ingressURL.Scheme == "http" {
+			opts = append(opts, config.WithScheme(config.SchemeHTTP))
+		}
+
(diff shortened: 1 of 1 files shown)


pkg/tunneldriver/driver.go
```diff
@@ -452,6 +452,12 @@ func (td *TunnelDriver) CreateAgentEndpoint(ctx context.Context, name string, sp
 			upstreamProto = string(*spec.Upstream.Protocol)
 		}
 		opts = append(opts, config.WithAppProtocol(upstreamProto))
+
+		// TODO: This should probably be inferred from the scheme in the URL
```
Alice-Lilith, line 456: Did you mean to leave this comment in? If the URL was already parsed using the helper func, then it should already have a scheme, even if the user used one of the shorthand formats.

Alice-Lilith (approved):

pkg/tunneldriver/driver.go
jonstacks: This was meant to comment on the ngrok-go SDK not respecting the scheme in the URL that we give it to remind me to fix it upstream. Open to removing it or modifying it for clarity.

Alice-Lilith: No worries, was just a little confused

PR #553, opened 2024-12-13
refactor: store driver

Alice-Lilith:
While working on the endpoint support, I felt like `internal/store` was a very overloaded package. 
This refactor while large, does not rewrite or change any of the logic. It does two main things:
1. Splits up the `driver.go` (which was very large) into supporting files for operations relating to edges/tunnels/etc to make it easier to read.
2. Moves the driver logic out of `internal/store` since I felt that while it consumes `internal/store`, it didn't really belong in that package since it is mostly just the logic for for the API manager service. I moved it into `pkg/managerdriver` to align with the tunnel driver and bindings driver. Unless there is a reason we have these drivers in `pkg`, they feel more like internal implementation and would be more suited under `internal/drivers/...`. I renamed the `UpdateStoreHandler` to be more in-line with it's new package and function.


hjkatz:
Note: No need to add the `safe to test` label anymore - e2e tests are always run as part of the mergequeue.

hjkatz (approved):
Personally I think of these drivers as belonging to the `ingress` and `gateway` feature sets (probably just "ingress" since both features are that same traffic direction), but `managerdriver` is just as good a name as any really.

PR #554, opened 2024-12-16
Rewrite Host header for HTTP protocols via net.Conn interceptor

hjkatz:
## What

When an endpoint is started with `--url foo.example` then the ngrok edge
expects all `Host: foo.example` headers to match _exactly_.

But this isn't always the case for kubernetes where `foo.example.svc.cluster.local` (via a cluster domai)
is resolvable to the bound endpoint forwarder service.

Therefore we need to also support updating the Host header somehow, here we choose to use an interceptor.

## How
*Describe the solution*

Intercept `net.Conn` reads via a `bufio.NewReader()` in chunks of 1024 bytes
- intercept until a replacement is found
- match/replace via provided function
- read remaining directly from the buffer via `io.Copy()`

## Breaking Changes
No

### DIFF
--- a/internal/controller/bindings/forwarder_controller.go
+++ b/internal/controller/bindings/forwarder_controller.go
@@ -25,6 +25,7 @@ SOFTWARE.
 package bindings
 
 import (
+	"bufio"
 	"context"
 	"crypto/tls"
 	"fmt"
@@ -32,6 +33,7 @@ import (
 	"net"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


hjkatz:
* **#554** <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-operator/554?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-operator/554?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

euank:
Closing this, it ended up getting solved service-side, where the host header is already parsed out regardless

PR #559, opened 2024-12-18
feat(helm): Add support for nodeSelectors, tolerations, and topologyS…

### LINKED ISSUE #557: [Helm] Support for nodeSelectors, podTopologyConstraints, and tolerations
Add support to our helm chart for nodeSelectors, podTopologyConstraints, and tolerations

jonstacks:
Fixes #557 

## What

Add support for `nodeSelectors`, `tolerations`, and `topologySpreadConstraints`.

## How

* Add these to the values.yaml and allow them to be configured for the bindings-forwarder, agent, and api manager. The top level values are for the api manager. We have plans to nest this in the future so that there can be global config that applies to all of these deployments, but for now, stick with this pattern.
* Also, move up the `podAnnotations` and `podLabels` config to the common section, since they are common configuration and not specific to the API/main manager.

## Breaking Changes
No


Alice-Lilith (approved):

Johndot1:
Ey

Johndot1 (approved):

PR #562, opened 2024-12-20
Add conversion support from ingress to endpoints

Alice-Lilith:
Adds support for conversion from `Ingress` resources when the `k8s.ngrok.com/use-endpoints: "true"` annotation is added. Without this annotation, `Ingress` resources will continue to create edges.

For each hostname on applicable `Ingress` resources, a `CloudEndpoint` managed resource will be created that routes to one or more managed `AgentEndpoint` resources that will be created for each unique upstream.


jonstacks (approved):
I haven't tested migrating from edges -> endpoints, but I deployed this to my home cluster and all my edges are still up :+1:

PR #564, opened 2025-01-08
feat: Add trafficpolicy package and conversion util

jonstacks:
## What

Adds a new `trafficpolicy` package for creating traffic policy. It also adds new `util` function to convert a moduleset to a traffic policy.

## How

Add utilities for creating adding rules and actions to a traffic policy. Create helpers to create actions for each type of action. Write the translation logic that converts modulesets to traffic policy with tests for each.

## Breaking Changes

No. All net new utilities

### DIFF
--- a/internal/trafficpolicy/policy.go
+++ b/internal/trafficpolicy/policy.go
@@ -5,8 +5,8 @@ import "time"
 // ActionType is a type of action that can be taken. Ref: https://ngrok.com/docs/traffic-policy/actions/
 type ActionType string
 
-// Expression is a string that represents a
-type Expresson string
+// Expression is a string that represents a traffic policy expression.
+type Expression string
 
 const (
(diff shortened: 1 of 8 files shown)


internal/trafficpolicy/policy.go
```diff
@@ -0,0 +1,199 @@
+package trafficpolicy
+
+import "time"
+
+// ActionType is a type of action that can be taken. Ref: https://ngrok.com/docs/traffic-policy/actions/
+type ActionType string
+
+// Expression is a string that represents a
```
TheConcierge, line 8: nit :)

Alice-Lilith (approved):
🔥 🚀

internal/trafficpolicy/policy.go
jonstacks: Thanks! Fixed in 3a13f54

PR #567, opened 2025-01-09
Remove binding name

masonj5n:
## What
The API for kubernetes bindings is being updated and no longer includes the `binding_name` field. Instead, kubernetes endpoints will be scoped to the account instead of a specific operator.

## How
Removes references to `binding_name`.

## Breaking Changes
The ngrok API for kubernetes bindings itself is breaking with this change so this is just following that.
As far as the operator itself:
- The `KubernetesOperatorBinding` resource had its `Name` field removed.
- `values.yaml` has been updated to no longer include `bindings.name`


Alice-Lilith:
🙏 looking forward to not getting my logs spammed with 

```
2025-01-10T00:31:52Z	DEBUG	events	Failed to update KubernetesOperator/ngrok-operator: HTTP 422: The request parameter 'name' is unknown and not expected. The supported fields are: allowed_urls, csr, ingress_endpoint. Please check your API client implementation and review the API documentation: https://ngrok.com/docs/api#api-kubernetes-operators-update. [ERR_NGROK_214]
```

api/ngrok/v1alpha1/kubernetesoperator_types.go
```diff
@@ -51,11 +51,6 @@ type KubernetesOperatorDeployment struct {
 }
 
 type KubernetesOperatorBinding struct {
-	// Name is the name of the k8s-binding for the account to bind to this configuration and the ngrok API
-	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Pattern=`^k8s[/][a-zA-Z0-9-]{1,63}$`
-	Name string `json:"name,omitempty"`
-
```
jonstacks, line 58: I wonder if we shouldn't leave this field in here but mark it as deprecated and change the validation to optional. The only thing I'm worried about if we remove it right now is if someone deploys this change and then rolls back to a previous version, what happens?

If there aren't a lot of users using bindings right now, I think its a lower risk change.

masonj5n: Since we're breaking the API on the hosted side bindings won't work with this field anyway so I don't think it matters _too_ much. Whenever we cut a release with these and a couple other upcoming changes that'll be the first release that works with bindings essentially

jonstacks (approved):

PR #574, opened 2025-01-16
add newly created agent endpoints to the map

Alice-Lilith:
We can't delete agent endpoints when the `AgentEndpoint` custom resources are deleted because we never add them to the map that tracks them. Adding the entry to the map makes this work as intended.

### DIFF
--- a/pkg/tunneldriver/driver.go
+++ b/pkg/tunneldriver/driver.go
@@ -480,6 +480,7 @@ func (td *TunnelDriver) CreateAgentEndpoint(ctx context.Context, name string, sp
 	if err != nil {
 		return err
 	}
+	td.agentEndpoints.Add(name, tun)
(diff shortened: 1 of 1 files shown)


pkg/tunneldriver/driver.go
```diff
@@ -480,6 +480,8 @@ func (td *TunnelDriver) CreateAgentEndpoint(ctx context.Context, name string, sp
 	if err != nil {
 		return err
 	}
+	td.agentEndpoints.Add(name, tun)
+	td.tunnels[name] = tun
```
jonstacks, line 484: I don't think we need to add this to the tunnels map. The agent endpoints should be tracked and removed separately from tunnels.

jonstacks (approved):
LGTM

PR #575, opened 2025-01-16
fix race condition in cloud endpoint controller

Alice-Lilith:
Fixes a race condition in the cloud endpoint controller. Currently we rely on the status.ID of a could endpoint to inform the controller about whether the cloud endpoint already exists and needs to be updated, or if this is a new custom resource and we need to update the cloud endpoint definition on the server side.

Relying on the status.ID within the controller leads to race conditions when a user (or in this case, the translation from ingresses to endpoints) updates/changes a `CloudEndpoint` resource after it was first created, but before the controller has assigned it a status.ID and updated it. When this happens, the incoming changed but not new `CloudEndpoint` still has no status.ID (which will eventually get properly set) and this causes the controller to create multiple different definitions for the cloud endpoint on the ngrok server side and then not being able to properly remove them once the `CloudEndpoint` resource is deleted. 

Instead of doing that, this makes the controller keep an in-memory map of the known `CloudEndpoint` resources so that we can simply check the in-memory map to see if one of them has been created and assigned an ID or not, rather than waiting for the status to be updated and the resource to be re-reconciled again after that.

### DIFF
--- a/internal/controller/ngrok/cloudendpoint_controller.go
+++ b/internal/controller/ngrok/cloudendpoint_controller.go
@@ -34,6 +34,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


Alice-Lilith:
I'm going to close this for now since I don't think anyone other than me has experienced this so far. We can re-open this if that is not the case.

PR #588, opened 2025-02-11
feat: gateway api -> endpoints

Alice-Lilith:
Adds support for conversion from Gateway API into endpoints. Best reviewed one commit at a time since there are a ton of updates to the tests and test data (which is a bulk of the changed content)

I'll add support for reference grants in a follow-up after this. We already don't support them for the Gateway API Edges conversion and this PR is already quite large even minus the test data.

I updated our Gateway API version to pull in some of the newer features which also required an update to controller runtime that meant I had to go update some changed function signatures in a few other unrelated controllers to satisfy the new package requirements.


jonstacks (approved):

Alice-Lilith:
quick rebase before merge to fix conflicts

PR #592, opened 2025-02-20
auto enable gateway api support

Alice-Lilith:
This PR aims to auto enable support for Gateway API when possible:
- Gateway API support flag changed from disabled by default to enabled by default
- When flag is omitted, default is enabled instead of disabled
- When enabled, we check if the Gateway API CRDs are installed first
  - If they are, we enable the Gateway API feature set
  - If they are not, we log that the Gateway API CRDs were not detected and the feature set will be disabled

I didn't want to remove the flag itself and rely solely on the CRD check for two reasons:
- Users may have the Gateway API CRDs installed in the cluster but not want to use them with the ngrok operator, so they should be able to disable that feature set and not waste resources running controllers for something they don't intend to use
- Users upgrading from one version to the next by just changing the image tag and not doing a `helm upgrade` would get a crash since they would be trying to set a flag that no longer exists

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -111,7 +111,7 @@ func cmd() *cobra.Command {
 
 	// feature flags
 	c.Flags().BoolVar(&opts.enableFeatureIngress, "enable-feature-ingress", true, "Enables the Ingress controller")
-	c.Flags().BoolVar(&opts.enableFeatureGateway, "enable-feature-gateway", false, "Enables the Gateway controller")
+	c.Flags().BoolVar(&opts.enableFeatureGateway, "enable-feature-gateway", true, "When true, enables support for Gateway API if the CRDs are detected. When false, Gateway API support will not be enabled")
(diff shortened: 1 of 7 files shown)


cmd/agent/main.go
```diff
@@ -145,56 +145,55 @@ func runController(ctx context.Context, opts managerOpts) error {
 	}
 
 	// shared features between Ingress and Gateway (tunnels)
-	if opts.enableFeatureIngress || opts.enableFeatureGateway {
```
Alice-Lilith, line 148: For the agent pods, I removed this check because currently users would not be able to use `AgentEndpoints` if they opt-out of Ingress and Gateway API support and that seems wrong IMO. Using the `AgentEndpoint` resources (or any others) directly instead of through our translation layer should be a perfectly valid use-case.

jonstacks (approved):

Alice-Lilith:
quick fix for the e2e tests before merge. Nothing interesting, just shifted the existing logic earlier.

PR #599, opened 2025-03-06
fix: add support for https agent endpoints from ingress/gwapi

Alice-Lilith:
We were previously not doing anything to support https upstreams for `AgentEndpoint` resources created from `Ingress`/`Gateway`/`HTTPRoute` resources. 

We now support for endpoints, the previously used `k8s.ngrok.com/app-protocols` from the Edges support, and additionally, when not present, we try to infer whether the upstream is HTTP/HTTPS based on the port's name or number. 

Additionally, previously when `backendTLS` was used on a `Gateway`, we would set client TLS certificates for ALL upstreams used for that `Gateway`. I've modified it to only include the client certificate refs on generated `AgentEndpoint` resources if the upstream URL scheme is `https://` although if the upstream URL scheme is `http://` then client certificate refs will be ignored because we aren't doing TLS.

As a stylistic change that has no impact on behaviour, I changed the generated `-tls-<hash>` part of AgentEndpoint names that use client certificates to `-mtls-<hash>` since it only gets generated when we are using client certificates.

TL;DR 

before this: `AgentEndpoints` generated from `Ingress`/`Gateway`/`HTTPRoute` resources always used an `http://` scheme which was wrong


pkg/managerdriver/utils.go
```diff
@@ -422,3 +397,80 @@ func doHostGlobsMatch(hostname1 string, hostname2 string) (bool, error) {
...
+			"service", fmt.Sprintf("%s.%s", service.Name, service.Namespace),
+		)
+		return "HTTP", nil
+	case int32(443):
+		log.Info("inferred protocol HTTPS for service port based on port number",
+			"port name", portName,
+			"port number", portNumber,
+			"service", fmt.Sprintf("%s.%s", service.Name, service.Namespace),
+		)
+		return "HTTPS", nil
```
jonstacks, line 471: Can we hold off on this new behavior? I'm not sure relying on the name or port number will do the right thing more times than it may do the wrong thing. In a service, someone might have named the port "HTTP", have it listening on port 443, and the scheme they actually want is `https`. 

I really like the idea, but in practice I think it might cause more problems. Maybe we punt on this for 1.0?

jonstacks (changes requested):

jonstacks (approved):

PR #602, opened 2025-03-07
fix(helm-chart): Resource requests for agent & bindings forwarder

### LINKED ISSUE #598: Allow configuring resource request/limits for agent independent of manager
### Description

Running this in prod, we are seeing a big difference in resource usage between the manager pod and the agent pods. Each agent pod is using roughly `952.15 MiB` memory vs the manager uses `65 MiB`. This makes sense based on the workloads each is handling. 

The helm chart does have a resources value https://github.com/ngrok/ngrok-operator/blob/31d8241695fa4c33b067cff6672a8b903eec8c0a/helm/ngrok-operator/values.yaml#L140-L161 but these are used for both the agent and manager. It would be nice to allow it to be configured here https://github.com/ngrok/ngrok-operator/blob/31d8241695fa4c33b067cff6672a8b903eec8c0a/helm/ngrok-operator/values.yaml#L258

### Use Case

Fine tuning resource requests

### Related issues

_No response_

jonstacks:
Fixes #598 

## What
Resource requests for the agent pods and bindings forwarder pods should be independent of the resource requests for the main api-manager.

## How

Add ability to set resource requests for the agent & bindings forwarder pods separate of the api-manager.

## Breaking Changes

Not necessarily breaking, but setting `resources.requests` and `resources.limits` will no longer propagate to the agent and bindings forwarder pods by default. If you don't set resource requests for each of these pods, they will now be unspecified.


helm/ngrok-operator/templates/agent/deployment.yaml
```diff
@@ -129,7 +129,7 @@ spec:
           initialDelaySeconds: 5
           periodSeconds: 10
         resources:
-        {{- toYaml .Values.resources | nindent 10 }}
+        {{- toYaml $agent.resources | nindent 10 }}
```
alex-bezek, line 132: nit: you might consider using the original value as a fallback. Just thinking this is passive for the helm values since they both default to empty. But in our case, we set this value so its being applied to both the manager and agent. This change would remove any resource requests from our agent unless we explicitly add the agent resources.

alex-bezek (approved):
Minor comment about passivity for your consideration, but looks good otherwise. Thanks!

helm/ngrok-operator/templates/agent/deployment.yaml
jonstacks: Yeah, I called that out in breaking changes. I don't think we should default to the root, but to maintain passivity I guess we can. This is something I want to fix by 1.0. We should have separate configuration for each of the 3 workloads, and then globals that apply to all.

jonstacks: Talked about this more synchronously, if we make this passive, there is no way to only set resource requests on the api-manager pods, since falling back to those values for the agent and bindings-fowarder if they are empty would overwrite that.

I think we are just going to note this in the changelog as mentioned on the PR.

PR #607, opened 2025-03-12
Consume IngressEndpoint from status

masonj5n:
## What
We're moving ingress endpoints to be something that's always provided by the ngrok api instead of something configurable by the user. This change starts using the ingress endpoint that's set on the KubernetesOperator status instead of the one on the spec.

We'll (eventually) remove it from the spec and chart.

## How
Get the ingress url off the status instead of spec.

## Breaking Changes
If you have a very old operator that hasn't ever gotten the ingress endpoint from the API it may stop working, but any update from the controller should fill it in.

### DIFF
--- a/internal/controller/bindings/forwarder_controller.go
+++ b/internal/controller/bindings/forwarder_controller.go
@@ -181,6 +181,11 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
 		return err
 	}
 
+	ingressEndpoint, err := getIngressEndpointWithFallback(op.Status.BindingsIngressEndpoint, log)
+	if err != nil {
+		log.Error(err, "failed to determine bindings ingress endpoint")
+	}
+
 	cnxnHandler := func(conn net.Conn) error {
 		defer conn.Close()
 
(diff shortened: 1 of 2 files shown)


internal/controller/bindings/forwarder_controller.go
```diff
@@ -196,7 +197,12 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
 
 		log.Info("Handling connnection")
 
-		ngrokConn, err := tlsDialer.Dial("tcp", *op.Spec.Binding.IngressEndpoint)
+		ingressEndpoint, err := getIngressEndpointWithFallback(op.Status.BindingsIngressEndpoint, log)
```
jonstacks, line 200: Should we pull this logic up out of the cnxnHandler? I think this parsing can be once outside of handling each connection.

masonj5n: Ah yep good catch, I definitely had tunnel vision there. I was a little worried about making sure connections get updated when that field on the k8sop status does but I don't think moving it up out of the connection handler actually changes that.

jonstacks (approved):

PR #610, opened 2025-03-18
add deployment name and version to _update calls

Megalonia:
## What
add deployment into update params
## How
*Describe the solution*
Add deployment to updataParms during _update calls
## Breaking Changes


internal/controller/ngrok/kubernetesoperator_controller.go
```diff
@@ -256,6 +256,10 @@ func (r *KubernetesOperatorReconciler) _update(ctx context.Context, ko *ngrokv1a
 		Metadata:        ptr.To(r.tryMergeMetadata(ctx, ko)),
 		EnabledFeatures: calculateFeaturesEnabled(ko),
 		Region:          ptr.To(ko.Spec.Region),
+		Deployment: &ngrok.KubernetesOperatorDeploymentUpdate{
+			Name:    &ko.Spec.Deployment.Name,
```
Megalonia, line 260: one concern i am having is if a user passes nothing on update will this field be ""? if a user explicitly sets to "" id say "i love it", but if not and it is ""... id say "i hate it"

jonstacks (approved):

PR #616, opened 2025-03-27
fix: traffic-policy annotation not working when using mapping-strategy of edges

jonstacks:
## What

When using the `edges` mapping-strategy, the `traffic-policy` annotation on ingresses was not getting correctly copied over to the edge routes.

## How

* Added a test to exercise the bug
* Slight refactor to fix the bug

## Breaking Changes
No

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,6 +21,7 @@
           yq
           kyverno-chainsaw
         ];
+        CGO_ENABLED = "0";
       };
     }));
 }
--- a/pkg/managerdriver/driver.go
+++ b/pkg/managerdriver/driver.go
@@ -606,7 +606,7 @@ func (d *Driver) getTrafficPolicyJSON(ingress *netv1.Ingress, modSet *ingressv1a
 		return nil, err
 	}
 
-	if modSet == nil {
+	if modSet.IsEmpty() {
 		if trafficPolicy != nil {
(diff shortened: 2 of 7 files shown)


flake.nix
```diff
@@ -21,6 +21,7 @@
           yq
           kyverno-chainsaw
         ];
+        CGO_ENABLED = "0";
```
jonstacks, line 24: This was added to make delve work with nix. Before adding this, we were using CGO locally, but not in the docker image that publish. This makes the development environment similar to what we ship and also allows us not to link against nix's C libs, so that we can run the debugger in vscode.

pkg/managerdriver/driver.go
```diff
@@ -606,7 +606,7 @@ func (d *Driver) getTrafficPolicyJSON(ingress *netv1.Ingress, modSet *ingressv1a
 		return nil, err
 	}
 
-	if modSet == nil {
```
jonstacks, line 609: This was the actual bug. The `modSet` passed was never nil, because we always return a non nil moduleset from:
https://github.com/ngrok/ngrok-operator/blob/7d62d40b77c532334390b0aa14eebcd36a01fea7/pkg/managerdriver/utils.go#L331-L338

Adding the `IsEmpty()` utility seemed like the easiest route here.

alex-bezek (approved):

PR #621, opened 2025-03-28
add support for TCPRoute & TLSRoute

Alice-Lilith:
Adds support for TLSRoute and TCPRoute.

I was initially planning on using the listener. hostname from the Gateway to define the domain for these endpoints as we do with HTTPRoute. This is valid per the scheme of the Gateway API and would have made things nice and neat, but Gateway API has X-Validations on the listener hostname that forbid you from using it if the listener protocol is TCP/TLS. This was a major bummer and complicated things a fair bit.

My solution was to add support for Gateway. addresses. This way, you can specify the TCP/TLS addresses from ngrok that you want and then use the listeners to define the ports that all the addresses should listen on. Unfortunately, this limitation means that you will likely have to make several Gateway resources instead of being able to specify all of your TLS/TCP addresses in a single Gateway, but there's nothing we can do about that limitation.

I wasn't initially planning on adding support for Gateway. addresses, but since this introduces that support, I think in the future it might make more sense to move to always using this model, and then instead of using the listener hostnames to define endpoint domains (for HTTPRoutes), we instead use them to add match criteria for the SNI/Host header. That would standardize things but would be a shift in how we do things. I think that a decent middle-ground solution would be to keep the current behavior when Gateway. addresses are not supplied.

Aside from that, something to note is that since TLSRoute/TCPRoute are in the experimental channel, they are not in the standard channel set of CRDs, so we only enable the controllers for them if those CRDs are detected.

### DIFF
--- a/cmd/api/main.go
+++ b/cmd/api/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -67,6 +68,7 @@ import (
 	"github.com/ngrok/ngrok-operator/internal/version"
 	"github.com/ngrok/ngrok-operator/pkg/managerdriver"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	gatewayv1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
 	//+kubebuilder:scaffold:imports
 )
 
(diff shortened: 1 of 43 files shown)


Alice-Lilith:
While the line-count looks insane, most of this was just testdata scaffolding and updating the Gateway API CRDs yaml that we use for tests.

Alice-Lilith:
I dropped the commit that ignored domain creation for CloudEndpoint and AgentEndpoint recources with a url begining with tcp:// or tls:// because the changes @jonstacks has been making more appropriately address this issue.

jonstacks (approved):

PR #623, opened 2025-03-28
only make tunnels for edges strategy

Alice-Lilith:
Right now, we still always make tunnels when you create `Ingress` and Gateway API resources. This changes it so that we only make tunnels for them when you are using the `edges` mapping-strategy translation.


pkg/managerdriver/driver_test.go
```diff
@@ -290,6 +291,7 @@ var _ = Describe("Driver", func() {
 				Expect(err).ToNot(HaveOccurred())
 			})
 
+			// TODO: Alice, fix broken
```
jonstacks, line 294: Are these tests just incorrect because they relied on the old behavior which generated the tunnels no matter what? Or are they broken in some other way? If so, I don't mind fixing them since I was probably the one who broke them when I wrote them 😬

jonstacks (approved):
LGTM! It cleaned up all my old `Tunnel`s after I had migrated everything to endpoints.

Screenshot 2025-04-01 at 1 38 41 AM

pkg/managerdriver/driver_test.go
Alice-Lilith: Oops, I forgot to remove some of those comments 😓 
The tests are fixed now but were initially broken because they were expecting tunnels when there were no longer any tunnels (because the default mapping strategy on the ingresses creates endpoints). I fixed the test by adding the `mapping-strategy: edges` annotation and adding a new one to check for no tunnels when that annotation is not included, but I forgot to clean up the comments.

PR #626, opened 2025-04-08
bake ca into image

Megalonia:
1st step to moving to a let's encrypt managed cert. Throw out helm volume and bake cert into image


.dockerignore
```diff
@@ -14,6 +14,7 @@
 !internal/
 !pkg/
 !scripts/
+!certs/
```
Megalonia, line 17: maybe I need !certs/*?

.dockerignore
```diff
@@ -14,6 +14,7 @@
 !internal/
 !pkg/
 !scripts/
+!certs/*
```
jonstacks, line 17: If we're going to pass this into the docker build context, do we want to store the cert in a regular file in the certs directory, so we don't have to `<<EOF` it in the Dockerfile?

Dockerfile
```diff
@@ -18,12 +18,44 @@ ARG TARGETOS TARGETARCH
 # Build
 RUN --mount=type=cache,target=/go \
 	CGO_ENABLED=0 GOOS="${TARGETOS}" GOARCH="${TARGETARCH}" make _build
+# Prepare CA certificates
+FROM golang:1.23 AS certs
```
jonstacks, line 22: Do we need a separate stage for this, or could we just copy it into distroless in the same spot?

Dockerfile
```diff
@@ -18,12 +18,44 @@ ARG TARGETOS TARGETARCH
...
+EOF
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/bin/api-manager /workspace/bin/agent-manager /workspace/bin/bindings-forwarder-manager ./
 USER 65532:65532
 
+# Copy the intermediate CA into the system certificates directory
+COPY --from=certs /certs/ngrok-ca.crt /etc/ssl/certs/ca-certificates.crt
```
jonstacks, line 59: This is a nuance of how docker builds, but we might consider moving this up higher, before the `WORKDIR`, so that it can be cached better. The cert won't change as frequently our binaries. Having it here will result in us having to push this layer each time.

Also, I noticed that we were mounting in the CA to the `/etc/ssl/certs/ngrok` folder and setting `SSL_CERT_DIR=/etc/ssl/certs/ngrok` in the bindings-forwarder deployment. I think we may need to change that with this change

helm/ngrok-operator/templates/bindings-forwarder/deployment.yaml
```diff
@@ -95,24 +95,10 @@ spec:
               fieldPath: metadata.namespace
         - name: HELM_RELEASE_NAME
           value: {{ .Release.Name | quote }}
-        - name: SSL_CERT_DIR
-          value: /etc/ssl/certs/ngrok
         {{- range $key, $value := .Values.extraEnv }}
         - name: {{ $key }}
           value: {{- toYaml $value | nindent 12 }}
         {{- end }}
-        {{- if .Values.extraVolumeMounts  }}
```
masonj5n, line 104: I think we still want to support adding extra volume mounts, we just don't want to stick the `ngrok-ca` one in here now

helm/ngrok-operator/templates/bindings-forwarder/deployment.yaml
```diff
@@ -95,23 +95,11 @@ spec:
...
-        {{- toYaml .Values.extraVolumeMounts | nindent 8 }}
-        - name: ngrok-ca
-          mountPath: /etc/ssl/certs/ngrok
-          readOnly: true
-        {{- else }}
-        volumeMounts:
-        - name: ngrok-ca
-          mountPath: /etc/ssl/certs/ngrok
-          readOnly: true
+        {{- if .Values.extraVolumeMounts }}
```
masonj5n, line 102:
```suggestion
        {{- if .Values.extraVolumeMounts }}
        volumeMounts:
        {{- toYaml .Values.extraVolumeMounts | nindent 8 }}
```

masonj5n: I think we still want people to be able to post extra volume mounts right?

masonj5n (approved):

PR #641, opened 2025-04-30
feat: Add a reclaimPolicy to domains

jonstacks:
## What

Domains are somewhat unique among the ngrok API types, especially ngrok managed domains. Once you reserve one, it belongs to you. If you delete it, it is released back into the pool and someone else can reserve it. Since domains also count against your limits, sometimes you want to delete them if they are only used temporarily so it doesn't impact your account limits. Other times, you want to keep the domain around as it is a well known domain that you want to use again.

## How

This adds a reclaim policy to domains. If the `reclaimPolicy == "Delete"`, the domain is deleted from your account when the `Domain` CR is deleted. If the `reclaimPolicy == "Retain"`, the domain is not deleted from your account and only the CR is deleted.

In a subsequent PR, we'll be plumbing this value through the helm chart so that you can set the default domain reclaim policy for new domains that are created by the operator.

## Breaking Changes

The output format when you run `kubectl get domains.ingress.k8s.ngrok.com` has changed. The `CNAMETarget` will now only be shown with the `-o wide` flag.

### DIFF
--- a/api/ingress/v1alpha1/domain_types.go
+++ b/api/ingress/v1alpha1/domain_types.go
@@ -34,6 +34,13 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+type DomainReclaimPolicy string
+
+const (
+	DomainReclaimPolicyDelete DomainReclaimPolicy = "Delete"
+	DomainReclaimPolicyRetain DomainReclaimPolicy = "Retain"
+)
+
 // DomainSpec defines the desired state of Domain
 type DomainSpec struct {
 	ngrokAPICommon `json:",inline"`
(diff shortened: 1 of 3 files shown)


api/ingress/v1alpha1/domain_types.go
```diff
@@ -45,6 +52,11 @@ type DomainSpec struct {
 	// Region is the region in which to reserve the domain
 	// +kubebuilder:validation:Required
 	Region string `json:"region,omitempty"`
+
+	// DomainReclaimPolicy is the policy to use when the domain is deleted
+	// +kubebuilder:validation:Enum=Delete;Retain
+	// +kubebuilder:default=Delete
```
alex-bezek, line 58: should this default to Retain to be passive?

alex-bezek (approved):
just a question about passivity

api/ingress/v1alpha1/domain_types.go
jonstacks: The current behavior is to delete the domain, so this is passive. We just don't issue deletes on the domain CRs from the other controllers for this reason. 

Eventually, we can have a `defaultDomainReclaimPolicy` that is set on the controller when it creates the domains so that we can delete the domain CRs when reconciling ingress/gw/lb services if the reclaim policy is retain.

PR #643, opened 2025-05-01
chore(tests): Use t.Context() instead of context.Background() in tests

jonstacks:
## What
More refactors... use `t.Context()` in tests and the `GinkgoT().Context()` in tests instead of `context.Background()` and `context.TODO()`. It seems a bit more idiomatic.

## How
* Switch our go.mod to building with the semantics of go1.24.
* Replace `context.Background()` with `t.Context()` in places we are using `testing.T`.
* In Ginkgo tests, replace `context.Background()` with `GinkgoT().Context()` 

## Breaking Changes
No. Just impacts tests


pkg/managerdriver/driver_test.go
```diff
@@ -62,7 +61,7 @@ var _ = Describe("Driver", func() {
 
 	Describe("Seed", func() {
 		It("Should not error", func() {
-			err := driver.Seed(context.Background(), fake.NewClientBuilder().WithScheme(scheme).Build())
+			err := driver.Seed(GinkgoT().Context(), fake.NewClientBuilder().WithScheme(scheme).Build())
```
jtszalay, line 64: `GinkgoT().Context()` I used `ctx SpecContext`. I'm not sure the difference but 👍

jtszalay (approved):

internal/controller/gateway/gatewayclass_controller_test.go
```diff
@@ -54,7 +53,7 @@ var _ = Describe("GatewayClass controller", func() {
 	)
 
 	BeforeEach(func() {
-		ctx := context.Background()
+		ctx := GinkgoT().Context()
```
alex-bezek, line 56: OOC does this do anything different functionaly or is it more of a semantics thing?

PR #649, opened 2025-05-08
refactor: Update statuses concurrently and pre-calculate domains

jonstacks:
## What

While working on setting status for gateway API resources, I noticed we were listing all the domains more times than we need to be. 

We'd also return early if we encountered any type of error. This could potentially result in a domain not getting updated that should be updated.

## How
* Pre-compute the map of domains and pass it to the function that calculates the ingress status.
* When updating ingress status, if we encounter an error, keep going. This is accomplished by first calculating which ingresses need to be updated and then updating them in an errorgroup.
* Finally, I scaffolded out some of the pieces that we'll need for updating httproute and gateway status. (Will be implemented in a future PR).

## Breaking Changes
No


alex-bezek (approved):
go routines go _brrr_

PR #658, opened 2025-06-10
feat(gateway-api): Validate gateway listeners hostname and port

jonstacks:
## What

This PR validates Gateway API CRs and updates status conditions. Given how ngrok works, there are requirements we should validate and set the status conditions for appropriately:

* HTTP Listeners must be on port 80 and must include a hostname.
* HTTPS Listeners must be on port 443 and must include a hostname.

This is not an exhaustive list, but a first pass at better supporting Gateway API status conditions.

## How

* Validate gateway & routes and appropriately set their status conditions.

## Breaking Changes

~Yes, this PR does include breaking changes. Previously if a gateway had invalid listeners they may have been ignored and part of the gateway would be realized. Now, if a gateway is changed so that it becomes invalid, the entire configuration for the gateway is removed.~

No. We will accept the gateway if it has at least one valid listener. Either way, we continue to add it to the store

### DIFF
--- a/internal/controller/gateway/gateway_controller.go
+++ b/internal/controller/gateway/gateway_controller.go
@@ -129,13 +129,8 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, err
 	}
 
-	// Validate the Gateway
+	// Validate the Gateway, conditionally modifying the status of the Gateway
 	_ = r.validateGateway(ctx, gw)
-
-	// Update the gateway status
-	if err := r.updateGatewayStatus(ctx, gw); err != nil {
-		return ctrl.Result{}, err
-	}
 
 	// Update the gateway in the store
 	if _, err := r.Driver.UpdateGateway(gw); err != nil {
@@ -161,6 +156,13 @@ const (
 func (r *GatewayReconciler) validateGateway(ctx context.Context, gw *gatewayv1.Gateway) error {
 	log := ctrl.LoggerFrom(ctx)
 	log.V(5).Info("Validating Gateway")
+
+	var changed bool
+	setStatusCondition := func(conditions *[]metav1.Condition, newCondition metav1.Condition) {
+		if meta.SetStatusCondition(conditions, newCondition) {
+			changed = true
+		}
+	}
 
 	// In order to preserve existing status conditions, we need copy an existing listener status
 	// if there is one.
@@ -198,14 +200,14 @@ func (r *GatewayReconciler) validateGateway(ctx context.Context, gw *gatewayv1.G
 		case gatewayv1.HTTPProtocolType:
 			switch {
 			case l.Port != 80:
-				meta.SetStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
+				setStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
 					gw,
 					gatewayv1.ListenerConditionAccepted,
 					gatewayv1.ListenerReasonPortUnavailable,
 					"ngrok only supports HTTP on port 80",
 				))
 			case l.Hostname == nil || *l.Hostname == "":
-				meta.SetStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
+				setStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
 					gw,
 					gatewayv1.ListenerConditionAccepted,
 					ListenerReasonHostnameRequired,
@@ -218,14 +220,14 @@ func (r *GatewayReconciler) validateGateway(ctx context.Context, gw *gatewayv1.G
 			switch {
 
 			case l.Port != 443:
-				meta.SetStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
+				setStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
 					gw,
 					gatewayv1.ListenerConditionAccepted,
 					gatewayv1.ListenerReasonPortUnavailable,
 					"ngrok only supports HTTPS on port 443",
 				))
 			case l.Hostname == nil || *l.Hostname == "":
-				meta.SetStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
+				setStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
 					gw,
 					gatewayv1.ListenerConditionAccepted,
 					ListenerReasonHostnameRequired,
@@ -235,7 +237,7 @@ func (r *GatewayReconciler) validateGateway(ctx context.Context, gw *gatewayv1.G
 				listenerValid = true
 			}
 		case gatewayv1.UDPProtocolType:
-			meta.SetStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
+			setStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
 				gw,
 				gatewayv1.ListenerConditionAccepted,
 				gatewayv1.ListenerReasonUnsupportedProtocol,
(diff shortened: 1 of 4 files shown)


internal/controller/gateway/gateway_controller.go
```diff
@@ -145,26 +159,188 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+					gw,
+					gatewayv1.ListenerConditionAccepted,
+					ListenerReasonHostnameRequired,
+					"ngrok does not support HTTPS listeners without a hostname",
+				))
+			default:
+				listenerValid = true
+			}
+		default:
+			listenerValid = true
```
kate-osborn, line 247: Do we support other protocol types?

internal/controller/gateway/gateway_controller.go
```diff
@@ -145,26 +159,188 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		}
+
+		if listenerValid {
+			meta.SetStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
+				gw,
+				gatewayv1.ListenerConditionAccepted,
+				gatewayv1.ListenerReasonAccepted,
+				"",
```
kate-osborn: nit: we could throw a happy message in here like we do when the gateway is accepted.

internal/controller/gateway/gateway_controller.go
```diff
@@ -145,26 +159,188 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			))
+			meta.SetStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
+				gw,
+				gatewayv1.ListenerConditionProgrammed,
```
kate-osborn: Are we setting Programmed to false when the listener is invalid?

internal/controller/gateway/gateway_controller.go
```diff
@@ -145,26 +159,188 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 
+	listenersAreValid := true
+	for _, l := range newStatus.Listeners {
+		if meta.IsStatusConditionFalse(l.Conditions, string(gatewayv1.ListenerConditionAccepted)) {
+			listenersAreValid = false
+			break
+		}
+	}
+
+	if !listenersAreValid {
```
kate-osborn: Some implementations choose to accept a Gateway as long as at least one listener is valid. The valid listeners are configured, while the invalid listeners are ignored. Is that something we can and want to support?

internal/controller/gateway/gateway_controller.go
```diff
@@ -145,26 +159,188 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	} else {
+		meta.SetStatusCondition(&newStatus.Conditions, metav1.Condition{
+			Type:               string(gatewayv1.GatewayConditionAccepted),
+			Status:             metav1.ConditionTrue,
+			Reason:             string(gatewayv1.GatewayReasonAccepted),
+			Message:            "gateway accepted by the ngrok controller",
+			ObservedGeneration: gw.Generation,
+		})
+	}
+
```
kate-osborn, line 306: Do we set the `GatewayConditionProgrammed` condition on the Gateway in another part of the code?

internal/controller/gateway/gateway_controller.go
```diff
@@ -145,26 +159,188 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	var gatewayValidationError error
+	if meta.IsStatusConditionTrue(newStatus.Conditions, string(gatewayv1.GatewayConditionAccepted)) {
+		gatewayValidationError = fmt.Errorf("gateway validation failed")
```
kate-osborn: nit: 
```suggestion
		gatewayValidationError = errors.New("gateway validation failed")
```

internal/controller/gateway/gateway_controller.go
```diff
@@ -145,26 +159,188 @@ func (r *GatewayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				gw,
+				gatewayv1.ListenerConditionAccepted,
+				gatewayv1.ListenerReasonAccepted,
+				"",
+			))
+			meta.SetStatusCondition(&listenerStatus.Conditions, r.newListenerCondition(
+				gw,
+				gatewayv1.ListenerConditionProgrammed,
+				gatewayv1.ListenerReasonPending,
+				"",
```
kate-osborn: nit: is there a helpful message we could add here?

internal/controller/gateway/gateway_controller.go
Alice-Lilith: IMO it's better to not throw away the whole gateway when only a few listeners are invalid.

internal/controller/gateway/gateway_controller.go
Alice-Lilith: We support TCP/TLS, but the implementation is a little weird since Gateway API doesn't really mesh with how ngrok handles the gateway infrastructure

internal/controller/gateway/gateway_controller.go
Alice-Lilith: No, prior to this PR, we aren't handling anything with gateway API status/conditions. This seems like some early validation in the controller layer for the most common reasons for a listener to be invalid, but later on in the translation layer, the config could still be invalid for other reasons, so there is more to do after this either way.

internal/controller/gateway/gateway_controller.go
Alice-Lilith: What is the benefit to setting Programmed to false on top of already setting `ListenerConditionAccepted`?

jonstacks: > Are we setting Programmed to false when the listener is invalid?

We aren't current setting it when the listener is invalid and it will just have the default `Unknown` status: https://github.com/ngrok/ngrok-operator/blob/3015076d8c90987afc4c106f05420ac50069bef2/internal/controller/gateway/testdata/gatewayapi-crds.yaml#L2585-L2595

We could set it to false when the gateway is not accepted.

> What is the benefit to setting Programmed to false on top of already setting ListenerConditionAccepted?

My thought was that once the gateway passes validation and is accepted, the programmed condition is used to wait for the gateway to be ready. Setting it to false provides some feedback that our operator has seen it and is working on realizing the configuration.

internal/controller/gateway/gateway_controller.go
jonstacks: Thank you both for the feedback. I'll look into how much work it is to support this and will give it a shot. I originally missed:
```go
	// This reason is used with the "Accepted" condition when one or
	// more Listeners have an invalid or unsupported configuration
	// and cannot be configured on the Gateway.
	// This can be the reason when "Accepted" is "True" or "False", depending on whether
	// the listener being invalid causes the entire Gateway to not be accepted.
	GatewayReasonListenersNotValid GatewayConditionReason = "ListenersNotValid"
```

internal/controller/gateway/gateway_controller.go
jonstacks: I'm going to pull this out into a separate PR and add a linter rule for it so we can enforce it programatically.

internal/controller/gateway/gateway_controller.go
kate-osborn: Gotcha. One easy add (could be a different PR) would be to set the `Accepted/False/UnsupportedProtocol` condition here if the protocol is UDP.

internal/controller/gateway/gateway_controller.go
kate-osborn: > What is the benefit to setting Programmed to false on top of already setting ListenerConditionAccepted?

I don't think it's strictly necessary since `Accepted:False` implies `Programmed:False`, but I could see a scenario where a user is querying the status of a listener specifically for the Programmed condition and might be confused by the `Pending` status. IMHO if we know the listener can't be programmed because its config is invalid, it doesn't hurt to set the Programmed condition to false.

internal/controller/gateway/gateway_controller.go
kate-osborn: Fair enough!

internal/controller/gateway/gateway_controller.go
jonstacks: I'll rebase this PR once https://github.com/ngrok/ngrok-operator/pull/659 merges

internal/controller/gateway/gateway_controller.go
jonstacks: UDP and associated status conditions should be handled now.

internal/controller/gateway/gateway_controller.go
jonstacks: This is fixed now

internal/controller/gateway/gateway_controller.go
jonstacks: Message has been added

internal/controller/gateway/gateway_controller.go
jonstacks: I'm currently setting it in the driver where we add the addresses for the gateway for external DNS. I'm not 100% happy with the split of where we are setting these in different places right now, but I haven't found a better alternative yet.

PR #667, opened 2025-07-16
feat: Migrate to ngrok-go v2

jonstacks:
## What
There is a new version of the ngrok-go SDK that will help us get rid of some of the code we were maintaining for forwarding connections.

## How
* Update to v2 of the ngrok-go SDK.
* Removes our logic for forwarding in favor of the new forwarder in the SDK.
* Refactors the tunnel driver into an agent driver

## Breaking Changes
No.


pkg/agent/driver.go
```diff
@@ -0,0 +1,332 @@
...
+		ngrok.WithAuthtoken(os.Getenv("NGROK_AUTHTOKEN")),
+		ngrok.WithLogger(slog.New(logr.ToSlogHandler(logger))),
+		ngrok.WithRPCHandler(func(_ context.Context, session ngrok.AgentSession, req rpc.Request) ([]byte, error) {
+			switch req.Method() {
+			case rpc.UpdateAgentMethod:
+				// Since the version of the embedded agent is fixed based on the k8s manifest,
+				// we don't support updating the agent.
+				err := errors.New("UpdateAgentMethod is not supported for the ngrok operator. Please update the ngrok-operator instead.")
+				return []byte(err.Error()), err
+			case rpc.StopAgentMethod, rpc.RestartAgentMethod:
```
Alice-Lilith, line 129: I think this makes sense for disconnect, but for restart, I'm not sure what the mechanism here is for restarting the pod or reconnecting the agent after it disconnects.

jonstacks: Currently there is no path to stop the agent and just have it sit idle until the pod is manually restarted. The mechanism for both of the RPCs is currently:
* Add a new error to the alive channel. (This channel is what informs the pods livenessprobe)
* Close the done chan. (This is used to prevent reconciles for trying to start a new endpoint, which triggers a new session automatically).
* The pods liveness probe starts to fail
* k8s restarts it

This is pretty close to what we did before v2 of the SDK. Alternatively, we could also `os.Exit(1)` directly from agent driver. Or send a signal and have the exit be done in the main package.

PR #670, opened 2025-07-31
reload certs on any changes and handle symlinks to certs

kmvachhani:
## What
This PR will handle 2 things:
- the operator will notice if the ca-certs file changes and force the pod to exit and restart so any changes can be loaded
- updates the cert loading code to properly resolve symlinks and eliminates the need for the `subPath` hack and allow automatic reloads 

## How
*Describe the solution*
- use fsnotify to watch the certs directory for any changes, enabling the operator to detect updates as soon as the ConfigMap refreshes the files. We then exit the process, forcing Kubernetes to restart the pod
- use `os.Stat` which resolves symlinks so that the operator reads the actual cert files instead of placeholders 
## Breaking Changes

no breaking changes

### DIFF
--- a/internal/util/certs.go
+++ b/internal/util/certs.go
@@ -116,21 +116,21 @@ func watchCertsDir() {
 			ctrl.Log.Info("Detected changes in custom certs directory, reloading certs")
 			os.Exit(0)
 		} else {
-			ctrl.Log.Info("Detected changes in custom certs directory, but NGROK_OPERATOR_EXIT_ON_CERT_CHANGE is not set to true, so not exiting")
+			ctrl.Log.Info("Detected changes in custom certs directory, but NGROK_OPERATOR_RESTART_ON_CERT_CHANGE is not set to true, so not restarting")
 		}
 	})
 }
 
 // shouldExitOnCertChange checks if the environment variable indicates we should exit on cert changes
 func shouldExitOnCertChange() bool {
-	envVar := os.Getenv("NGROK_OPERATOR_EXIT_ON_CERT_CHANGE")
+	envVar := os.Getenv("NGROK_OPERATOR_RESTART_ON_CERT_CHANGE")
 	if envVar == "" {
 		return false
 	}
 
 	shouldExit, err := strconv.ParseBool(envVar)
 	if err != nil {
-		ctrl.Log.Info("Invalid boolean value for NGROK_OPERATOR_EXIT_ON_CERT_CHANGE, defaulting to false", "value", envVar)
+		ctrl.Log.Info("Invalid boolean value for NGROK_OPERATOR_RESTART_ON_CERT_CHANGE, defaulting to false", "value", envVar)
 		return false
 	}
 
--- a/helm/ngrok-operator/values.yaml
+++ b/helm/ngrok-operator/values.yaml
@@ -185,7 +185,7 @@ extraEnv: {}
 ##     secretKeyRef:
 ##       key: test-key
 ##       value: test-value
-##   NGROK_OPERATOR_EXIT_ON_CERT_CHANGE: "false"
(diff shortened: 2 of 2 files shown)


jonstacks (commented):
It looks like we only load the certs once when we create a new agent driver:

https://github.com/ngrok/ngrok-operator/blob/c2a51289c9e0c2c4761f341684acecb95fee7149/pkg/agent/driver.go#L202-L206

When the certs change on disk, I think we either need to load them and swap out the driver instance for a new one so that the ngrok go SDK is using the new certs, or terminate the pod and rely on the deployment restarting the pod.

alex-bezek:
> good call out, I've changed the code to restart the pod on changes to the certs directory & tested in my local box

I think one concern with this approach is how it handles High Availability during a rollout. If the cert is a k8s secret or config map mapped to a file on the system then when it gets updated, it would get updated in each of our agent pods at roughly the same time. The agent-manager pods we typically run multiple of and each one has a duplicated copy of pooled agent endpoints. This ensures during a normal deploy were we roll through 1 by 1 we always have 1 pod up at any time. But wiith this os.exit, all of them could shut down at the same time creating downtime. 

@jonstacks did mention another option of creating a new instance of the Driver and swapping it out. we initialize it and pass it to the reconciler here https://github.com/ngrok/ngrok-operator/blob/8251a186193dac40fc6988ada0fa5d0987f7421c/cmd/agent-manager.go#L161-L166 . I'm not sure if we would need to setup some form of a mutex to pause the reconciler while we indina jones swap things out. 

The other consideration is that this is a bit of an undocumented feature I think that you can mount other certs at `/etc/ssl/certs/ngrok`. I believe there are legitimate use cases for this where some people run a proxy between their agents and our tunnel servers and want to use their own cert. There might be a different way to make this a first class feature like taking a config map name or secret name to mount in that location and the operator api manager pod could watch these and do a rolling restart of the agent-manager pods.

alex-bezek:
I discussed this with @jonstacks some more the other day. There are some more options which would be a bit futher down the road and need some more discussions like pushing the cert change functionality into the ngrok-go sdk, or having an operator component handle rolling pods on cert changes. 

We thought for now we could just put this implementation where the pods restart on change behind a feature flag

internal/util/certs.go
```diff
@@ -61,3 +78,61 @@ func LoadCerts() (*x509.CertPool, error) {
...
+			}
+			return err
+		}
+	}
+}
+
+// watchCertsDir watches for changes in the certs directory and exits on any update
+func watchCertsDir() {
+	_ = watchCertsDirWithHandler(customCertsPath, func() {
+		if shouldExitOnCertChange() {
```
alex-bezek, line 115: I was originally thinking we wouldn't even setup the file system watcher if the feature flag wasn't set, but this might be better so if someone does update certs and doesn't have this set, the message is kind of self documenting to lead them to this feature flag so I think this is probably better 👍

internal/util/certs.go
```diff
@@ -61,3 +78,61 @@ func LoadCerts() (*x509.CertPool, error) {
...
+			ctrl.Log.Info("Detected changes in custom certs directory, reloading certs")
+			os.Exit(0)
+		} else {
+			ctrl.Log.Info("Detected changes in custom certs directory, but NGROK_OPERATOR_EXIT_ON_CERT_CHANGE is not set to true, so not exiting")
```
alex-bezek: Nit: I think Exit is how the implementation detail here works, but as a user "Exit" would concern me vs something like "Restart". we just happen to know here that its the process exiting which k8s will restart.

kmvachhani: changed it to `NGROK_OPERATOR_RESTART_ON_CERT_CHANGE` 👍

PR #679, opened 2025-09-25
Alex/domain ref status

alex-bezek:
## What

- changes cloud endpoint to store a domainRef instead of the domain status
- has the cloud endpoint controller populate both for passivity
- create a separate domain status type that has just the fields we had prior to this change https://github.com/ngrok/ngrok-operator/pull/678
- add the same domainRef to agentEndpoints
- update the service controller to fetch the domain instead of just using the cname

## How
populate the new field and use it to get the same value

## Breaking Changes
No not really. Technically this is reducing the fields in the CloudEndpoint.Status.Domain field but only back to what it was prior to https://github.com/ngrok/ngrok-operator/pull/678 

## Validation

Created some local test cases

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-server
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: test-server
  template:
    metadata:
      labels:
        app.kubernetes.io/name: test-server
    spec:
      containers:
        - name: test-server
          image: alpine:latest
          command: ["/bin/sh"]
          args:
            - -c
            - |
              apk add --no-cache netcat-openbsd
              while true; do
                echo "Hello from TCP server - $(date)" | nc -l -p 8080
              done
          ports:
            - containerPort: 8080
              name: tcp

---
# TCP LoadBalancer (no URL annotation = auto-assigned)
apiVersion: v1
kind: Service
metadata:
  name: test-tcp
  namespace: default
  annotations:
    k8s.ngrok.com/mapping-strategy: endpoints-verbose
  labels:
    app.kubernetes.io/name: test-server
spec:
  allocateLoadBalancerNodePorts: false
  loadBalancerClass: ngrok
  ports:
    - name: tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
  selector:
    app.kubernetes.io/name: test-server
  type: LoadBalancer

---
# TLS LoadBalancer (with custom domain)
apiVersion: v1
kind: Service
metadata:
  name: test-tls
  namespace: default
  annotations:
    k8s.ngrok.com/domain: test-tcp.alexbezek.me
    k8s.ngrok.com/mapping-strategy: endpoints-verbose
  labels:
    app.kubernetes.io/name: test-server
spec:
  allocateLoadBalancerNodePorts: false
  loadBalancerClass: ngrok
  ports:
    - name: tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
  selector:
    app.kubernetes.io/name: test-server
  type: LoadBalancer
```

And i could see the statuses were set

```
  status:
    loadBalancer:
      ingress:
      - hostname: 7.tcp.ngrok.io
        ports:
        - port: 22457
          protocol: TCP

  status:
    loadBalancer:
      ingress:
      - hostname: 2kapoyyjhr5reymrz.3kg7imcniwfgy1ozt.ngrok-cname.com
        ports:
        - port: 443
          protocol: TCP
```

I also created chainsaw tests to validate this in an e2e setup. I modified the code to hardcode the value to "incorrect" to show the tests fail

```
=== NAME  chainsaw/loadbalancer-services
    | 00:18:22 | loadbalancer-services | verify TLS LoadBalancer service gets status | ASSERT    | ERROR | v1/Service @ chainsaw-firm-man/test-tls-lb
        === ERROR
        ----------------------------------------
        v1/Service/chainsaw-firm-man/test-tls-lb
        ----------------------------------------
        * status.(length(to_array(loadBalancer.ingress)[? hostname && (pattern_match('*.ngrok.app', hostname) || pattern_match('*.ngrok-cname.com', hostname)) ]) > `0`): Invalid value: false: Expected value: true

        --- expected
        +++ actual
        @@ -3,10 +3,5 @@
         metadata:
           name: test-tls-lb
           namespace: chainsaw-firm-man
        -status:
        -  ? (length(to_array(loadBalancer.ingress)[? hostname && (pattern_match('*.ngrok.app',
        -    hostname) || pattern_match('*.ngrok-cname.com', hostname)) ]) > `0`)
        -  : true
        -  (length(to_array(loadBalancer.ingress)[? length(to_array(ports)) > `0` && to_number(to_array(ports)[0].port) > `0`]) > `0`): true
        -  (length(to_array(loadBalancer.ingress)[?hostname]) > `0`): true
        +status: {}
```

### DIFF
--- a/tools/make/deploy.mk
+++ b/tools/make/deploy.mk
@@ -61,6 +61,7 @@ deploy_with_bindings: _deploy-check-env-vars docker-build manifests kustomize _h
 
 .PHONY: deploy_for_e2e
 deploy_for_e2e: _deploy-check-env-vars docker-build manifests kustomize _helm_setup ## Deploy controller to the K8s cluster specified in ~/.kube/config.
+	kind load docker-image $(IMG)
 	helm upgrade $(HELM_RELEASE_NAME) $(HELM_CHART_DIR) --install \
 		--namespace $(KUBE_NAMESPACE) \
 		--create-namespace \


tools/make/deploy.mk
```diff
@@ -61,7 +61,6 @@ deploy_with_bindings: _deploy-check-env-vars docker-build manifests kustomize _h
 
 .PHONY: deploy_for_e2e
 deploy_for_e2e: _deploy-check-env-vars docker-build manifests kustomize _helm_setup ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	kind load docker-image $(IMG)
```
alex-bezek, line 64: @sabrina-ngrok FYI I removed this since i'm using minikube and this fails. But in theory this is what you'll need in the `deploy` make target. 
I think we have 2 options going forward:
- allow our make targets to conditionally do this
- standardize on kind and have all our make commands automatically do this. 

We had avoided the later in the past to just let people use what they want, but i think everyone has their own unique setup at this point and so onboarding is a bit trickier. I'm feeling more inclined to pick 1 tool to document how to get started with quickly. cc @jonstacks if you have opinions

jonstacks: FWIW, I think this will cause your PR to fail e2e in the merge queue and on main since we won't be able to load the image into the kind cluster.

We probably need to come up with a blessed path or 2 for local development to make it easier to onboard and to automate. Ideally local dev has parity with CI, or at least as close as possible.

alex-bezek: oh interesting, i saw it was running this command in the CI for this branch and had failed https://github.com/ngrok/ngrok-operator/actions/runs/18011065235/job/51243643129#step:3:865 and then passed once i fixed it. But i'm assuming its just pulling the public docker image. 

I'll just revert this for now. I may just swap over to kind and we can make that our default that we use for local since we use it in CI if that sounds good to you.

PR #691, opened 2025-10-22
refactor condition consts to be typed and available in api package

alex-bezek:
## What

refactors conditions consts and helpers to have types for each condition and reason string. This helps us generate docs for each type and prevents footgun issues of using the wrong condition or reason in the wrong situation

## How
- create specific types
- move them into the api package
- creates a reusable generic FindCondition function that helps with each
- Unfortunately, has lots of casting from our types to strings in the condition test files

## Breaking Changes
No


internal/controller/agent/agent_endpoint_conditions.go
```diff
@@ -7,35 +7,17 @@ import (
...
-// Standard condition types for AgentEndpoint
-const (
-	ConditionReady           = "Ready"
-	ConditionEndpointCreated = "EndpointCreated"
-	ConditionTrafficPolicy   = "TrafficPolicyApplied"
-)
-
-// Standard condition reasons
-const (
-	ReasonEndpointActive     = "EndpointActive"
```
jonstacks, line 19: **nit**: We could alias these from the `ngrokv1alpha1` package to make the code easier to read. Just a thought

api/ngrok/v1alpha1/cloudendpoint_conditions.go
```diff
@@ -0,0 +1,52 @@
...
+	// This condition will be True when endpoint creation succeeds,
+	// and False if endpoint creation fails.
+	CloudEndpointConditionCreated CloudEndpointConditionType = "CloudEndpointCreated"
+)
+
+// Reasons for Ready condition
+const (
+	// CloudEndpointReasonActive is used when the CloudEndpoint is fully active
+	// and ready in the ngrok cloud.
+	CloudEndpointReasonActive CloudEndpointConditionReadyReason = "CloudEndpointActive"
```
jonstacks, line 29: We can always do this later, but maybe consolidate the conditions between cloud & agent endpoint to just `EndpointActive` like the agent endpoint does

internal/controller/agent/agent_endpoint_conditions.go
alex-bezek: what strat would you use here? We already import the CRDs through the same package, and each package contains multiple CRDs so i don't think i can alias it to something CRD specific or condition specific

api/ngrok/v1alpha1/cloudendpoint_conditions.go
alex-bezek: You mean to update the string value to just be EndpoitnActive and get rid of this variable in favor of a common `EndpointReasonActive` variable? I actually had in my todo notes to do the opposite and make the agent endpoint string more specific to follow this pattern. 

I'm a bit hesitant to combine them because:
- they are very similar but have some differences and might diverge more in minor ways so i'm weary of trying to dry things up too much when there are only 2
- everything else is a 1-1 pattern so it makes a one off pattern. 

I think i need to fix the consistency one way or another though. If you have a strong opinion one way or another I'm happy to do that, otherwise I might update the agent endpoint strings to specify agentendpoint in them to be consistent with this and domains

internal/controller/agent/agent_endpoint_conditions.go
jonstacks: I left it as a nit because its more of a personal preference on easier reading. Here's a diff of what it would look like https://github.com/ngrok/ngrok-operator/compare/alex/conditions-types-refactoring...jonstacks:ngrok-operator:stacks/condition-refactoring?expand=1

My main thought was the lines were getting really long and the with the import alias in there already it was hard to read. We're also in the `agent` package already, so for the aliases: `ngrokv1alpha1.AgentEndpointReasonActive` -> `ReasonActive` in this package was the thought.

If you think its an improvement, feel free to merge it into your branch

api/ngrok/v1alpha1/cloudendpoint_conditions.go
jonstacks: I thought they were more similar and just trying to collapse the surface area where possible. If there is enough nuance in the conditions, then +1 with keeping them separate, but either way I think we should shoot for consistency.

jonstacks (approved):

internal/controller/agent/agent_endpoint_conditions.go
alex-bezek: Looking at this more closely, i do like the readability, but I'm not a big fan of the aliases.go file. The branch has it setup for the agent conditions, but we would need to do this for each of the conditions in each package that has conditions. I worry this will just be more maintenance and confusing on where to add conditions. Additionally, it looks like you lose out on the original type documentation
image
image

If you just want to reduce the line length, we could import the v1alpha1 package like this instead

`. "github.com/ngrok/ngrok-operator/api/ngrok/v1alpha1"`
and drop the `ngrokv1alpha1` from a bunch of lines
image

Although in theory hopefully these become just `v1` soon

I also considered if the variable names should just drop the resource specific prefix since agent endpoint and cloud endpoint are in the same package and could be combined, but i believe we also plan to collapse the different api packages down into a single one which would change all that so we would need to distriguish between endpoints, domains, boundendpoints, etc

PR #694, opened 2025-10-24
fix(lb-services): Fix too frequent reconciles for lb-services

jonstacks:
## What

The change in #693 to fix a bug with Cloud/Agent Endpoint changes not triggering re-reconciles had the adverse affect of watching for all changes on cloud/agent endpoints and re-conciling too frequently. We now use a better predicate filter on the service to determine if it needs to be updated.

## How

The status update was also not deterministic since the cloud and agent endpoint status are different. We now correctly determine the right object to use for status update.

In addition, we now also guard against trying to update the status if it is already up to date.

## Breaking Changes
No

### DIFF
--- a/internal/controller/service/controller.go
+++ b/internal/controller/service/controller.go
@@ -88,6 +88,34 @@ type ServiceReconciler struct {
 	TCPAddresses     ngrokapi.TCPAddressesClient
 }
 
+type ShouldHandleServicePredicate = TypedShouldHandleServicePredicate[client.Object]
+
+type TypedShouldHandleServicePredicate[object client.Object] struct {
+	predicate.TypedFuncs[object]
+}
+
+func (p TypedShouldHandleServicePredicate[object]) Create(e event.CreateEvent) bool {
+	svc, ok := e.Object.(*corev1.Service)
+	if !ok {
+		return false
+	}
+	return shouldHandleService(svc)
+}
+
+func (p TypedShouldHandleServicePredicate[object]) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil || e.ObjectNew == nil {
+		return false
+	}
+
+	oldSvc, ok1 := e.ObjectOld.(*corev1.Service)
+	newSvc, ok2 := e.ObjectNew.(*corev1.Service)
+	if !ok1 || !ok2 {
+		return false
+	}
+	// We need to reconcile if either the old or new service should be handled
+	return shouldHandleService(oldSvc) || shouldHandleService(newSvc)
+}
+
 func (r *ServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	if r.ClusterDomain == "" {
 		r.ClusterDomain = common.DefaultClusterDomain
@@ -111,16 +139,12 @@ func (r *ServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 
 	controller := ctrl.NewControllerManagedBy(mgr).
-		For(&corev1.Service{}, builder.WithPredicates(predicate.Funcs{
-			// Only handle services that are of type LoadBalancer and have the correct load balancer class
-			CreateFunc: func(e event.CreateEvent) bool {
-				svc, ok := e.Object.(*corev1.Service)
-				if !ok {
-					return false
-				}
-				return shouldHandleService(svc)
-			},
-		})).
+		For(&corev1.Service{}, builder.WithPredicates(
+			predicate.And(
+				ShouldHandleServicePredicate{},
+				predicate.ResourceVersionChangedPredicate{},
+			),
+		)).
 		// Watch traffic policies for changes
 		Watches(
 			&ngrokv1alpha1.NgrokTrafficPolicy{},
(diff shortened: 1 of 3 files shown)


internal/controller/service/controller.go
```diff
@@ -283,13 +312,19 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	if err != nil {
+		log.Error(err, "Failed to determine object for status update")
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
+	if err := subResourceReconcilers.UpdateServiceStatus(ctx, r.Client, svc, statusObject); err != nil {
+		return ctrl.Result{}, errors.Errorf("failed to update service status: %w", err)
+	}
+
+	r.Recorder.Event(svc, corev1.EventTypeNormal, "Reconciled", "Successfully reconciled service and its ngrok resources")
```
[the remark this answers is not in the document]
jonstacks: Since events only persist for an hour by default, I think this will be useful to see when the service was last reconciled successfully. Thanks for suggesting it as a nitpick

internal/controller/service/controller.go
```diff
@@ -283,13 +312,19 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+
+	// Determine which object to use for updating the service status based on mapping strategy
+	statusObject, err := r.getObjectForStatusUpdate(mappingStrategy, ownedResources)
+	if err != nil {
+		log.Error(err, "Failed to determine object for status update")
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
+	if err := subResourceReconcilers.UpdateServiceStatus(ctx, r.Client, svc, statusObject); err != nil {
+		return ctrl.Result{}, errors.Errorf("failed to update service status: %w", err)
```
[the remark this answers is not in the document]
jonstacks: Fixed with amend in 369b240

internal/controller/service/controller.go
```diff
@@ -283,13 +312,19 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	if err != nil {
+		log.Error(err, "Failed to determine object for status update")
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
+	if err := subResourceReconcilers.UpdateServiceStatus(ctx, r.Client, svc, statusObject); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to update service status: %w", err)
+	}
+
+	r.Recorder.Event(svc, corev1.EventTypeNormal, "Reconciled", "Successfully reconciled service and its ngrok resources")
```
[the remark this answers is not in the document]
jonstacks: We already covered this in your previous comment copilot....

internal/controller/service/controller.go
```diff
@@ -88,6 +88,34 @@ type ServiceReconciler struct {
...
+func (p TypedShouldHandleServicePredicate[object]) Create(e event.CreateEvent) bool {
+	svc, ok := e.Object.(*corev1.Service)
+	if !ok {
+		return false
+	}
+	return shouldHandleService(svc)
+}
+
+func (p TypedShouldHandleServicePredicate[object]) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil || e.ObjectNew == nil {
```
alex-bezek, line 106: would it be worth checking the generations are different as well or even leveraging this helper function https://github.com/kubernetes-sigs/controller-runtime/blob/v0.22.3/pkg/predicate/predicate.go#L213-L224 ?

alex-bezek (approved):

internal/controller/service/controller.go
jonstacks: Right now, I have https://github.com/ngrok/ngrok-operator/blob/369b240af71e917cc4cf3e3d67e062d04d03d7a8/internal/controller/service/controller.go#L142-L147

Basically making our own predicate that can be used with the `predicate.And` or `predicate.Or` functions. I want to keep this one we made tightly scoped to to only if we should handle the service(either the old or new version). This will filter out changes to non load balancer service, or services that aren't our LB class, so we only trigger reconciles for things that are or were a LB service and our LB class.

If we want to check the generation changed as well, we should use it in conjunction with the `predicate.And` or `predicate.Or` to keep it cleaner IMO.

This might be a case we don't need to handle, but I had it like that in case the status is changed by a user or tool, changing the resource version. Might be an over optimization against bad behavior by a user or tool that we don't need to handle and could be removed.

jonstacks: Doing some more research, I think you are right here. I'm being overly defensive when I don't think there is generally that expectation. I'll update it

jonstacks: After further investigation,

It looks like this is the best we can do unless we right our own custom diffing logic. The API server doesn't increment generation on services like it does for other types :confused: 

This logic lives under the `registry` package. Here is the logic for [Pod](https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/pod/strategy.go#L409) and here is the logic for [Service](https://github.com/kubernetes/kubernetes/blob/master/pkg/registry/core/service/strategy.go), noting a lack of the word Generation in the second :disappointed:

PR #702, opened 2025-10-30
fix: Created condition on existing IP Policy Rules

jonstacks:
Opening in draft until I have tests to re-create this scenario.

## What

1. Add the `Ready` column to the printer output so we can easily see if an IP Policy is ready or not.
2. In cases where the IP Policy already existed, it was setting the created condition to False. If the policy can be found by ID, it exists, so we should add the condition. I think we also shouldn't set `ReasonIPPolicyRulesConfigured` to true until after we've configured all of the rules, so this moves that to the end where we've processed all of the rules.

## How

Update the code to match the logic above.

## Breaking Changes
No

### DIFF
--- a/internal/controller/ingress/ippolicy_controllers_test.go
+++ b/internal/controller/ingress/ippolicy_controllers_test.go
@@ -123,6 +123,67 @@ var _ = Describe("IPPolicyReconciler", func() {
 			return count
 		}, timeout, interval).Should(Equal(0))
 	})
+
+	It("sets Created condition for existing IP Policy", func() {
+		By("creating an IP Policy")
+		ip := &ingressv1alpha1.IPPolicy{
+			ObjectMeta: metav1.ObjectMeta{Name: "test-existing-policy", Namespace: "default"},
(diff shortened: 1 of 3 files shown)


internal/controller/ingress/ippolicy_controller.go
```diff
@@ -135,6 +135,12 @@ func (r *IPPolicyReconciler) update(ctx context.Context, policy *ingressv1alpha1
 		return r.controller.ReconcileStatus(ctx, policy, err)
 	}
 
+	// If we found the remote policy, set the Created condition to true
+	// if it was not already set.
+	if !hasIPPolicyCreatedCondition(policy) {
+		setIPPolicyCreatedCondition(policy, true, ReasonIPPolicyCreated, "IP Policy already exists")
+	}
```
jonstacks: After talking @alex-bezek, he pointed out that we probably don't even need the has condition check. The controller-runtime helper for this should gracefully handle that. 👍

alex-bezek (approved):

PR #705, opened 2025-11-06
Alex/create endpoints without domains ready

alex-bezek:
## What
- makes the bindings string array only allow 1 value via the kube api {ngrok api only allows 1)
- agentendpoints and cloudendpoints with kubernetes bindings no longer get domains created or domainRef status fields
- agentendpoints and cloudendpoints now get created even if their domains are not ready. The big bug is that the condition changes made them not get created if the domain isn't ready. 
  - this meant endpoints with invalid domains that won't get certs but could still be used over http wouldn't get created
  - kubernetes bound endpoints created via CRDs also got domain CRs created for their k8s bound domain. These won't go ready since their TLD is the k8s namespace and not a domain that is owned. 
 

## How
- adds `MaxItems=1` to bindings array. The ngrok api rejects any array with more than 1 element so this should match and provide quicker feedback
- EnsureDomain no longer returns an error when the domain is not done reconciling. This prevents the early return in the create functions that blocks the creation of the resources. Instead it will just be reflected in the ready status but won't affect the creation process. 
- skip creating domains for kubernetes bound endpoints. 
- add a watcher for domains using the domain ref to find matches and enque events

## Breaking Changes
yes, this makes the bindings array only allow 1 item. but we pass all of them into the SDK and the API only accepts arrays with 1 value. So if someone was using a CRD with multiple bindings, those wouldn't be working anyways

## Validation 

It all works from a fresh state, but i'm having an issue i haven't figured out for migrating.  My steps are:
- deploy off main
- apply my test manifests https://gist.github.com/alex-bezek/e0e952e5a49e1dd12c430521c186681b
- see the bad state 
  - agent and cloud endpoints with not ready domains not being created
  - kubernetes bound endpoint gets created
- checkout branch and deploy
- all agent and cloud endpoints are created ✅ 
- all cloud endpoints are ready and corresponding domains are gone ✅ 
- agent endpoint with k8s binding get their domain deleted but recreated ❌ <--- NOW WORKS!! 🎉 

after upgrading to this branch i have this as my state. I would expect the agent endpoint k8s bound domain to have been deleted like the cloudendpoint one and for its status to be the same. 

```
kubectl get agentendpoints -A -o custom-columns="NAME:.metadata.name,URL:.spec.url,BINDINGS:.spec.bindings,READY:.status.conditions[?(@.type=='Ready')].status,EP-CREATED:.status.conditions[?(@.type=='EndpointCreated')].status,DOMAIN-READY:.status.conditions[?(@.type=='DomainReady')].status,MESSAGE:.status.conditions[?(@.type=='Ready')].message"

kubectl get cloudendpoints -A -o custom-columns="NAME:.metadata.name,URL:.spec.url,BINDINGS:.spec.bindings,READY:.status.conditions[?(@.type=='Ready')].status,EP-CREATED:.status.conditions[?(@.type=='CloudEndpointCreated')].status,DOMAIN-READY:.status.conditions[?(@.type=='DomainReady')].status,MESSAGE:.status.conditions[?(@.type=='Ready')].message"

kubectl get domains -A -o custom-columns="NAME:.metadata.name,DOMAIN:.spec.domain,READY:.status.conditions[?(@.type=='Ready')].status,REASON:.status.conditions[?(@.type=='Ready')].reason,MESSAGE:.status.conditions[?(@.type=='Ready')].message" | grep ngrok-operator
NAME                               URL                                               BINDINGS       READY   EP-CREATED   DOMAIN-READY   MESSAGE
ngrok-operator-bindings-test-aep   http://ngrok-operator-bindings-test-aep.default   [kubernetes]   False   True         False          Domain is being created
ngrok-operator-internal-aep        http://ngrok-operator-internal-aep.internal       [internal]     True    True         True           AgentEndpoint is active and ready
ngrok-operator-public-aep          https://ngrok-operator-public-aep.example.com     [public]       False   True         False          DNS_ERROR Reserved domain "ngrok-operator-public-aep.example.com" CNAME DNS record is not pointing at 4s5mnfledhnnk5u5r.3kg7imcniwfgy1ozt.ngrok-cname.com. Started at 2025-11-06T04:25:59Z Retries at 2025-11-06T04:46:14Z
NAME                                URL                                                BINDINGS       READY   EP-CREATED   DOMAIN-READY   MESSAGE
ngrok-operator-bindings-test-clep   http://ngrok-operator-bindings-test-clep.default   [kubernetes]   True    True         True           CloudEndpoint is active and ready
ngrok-operator-internal-clep        http://ngrok-operator-internal-clep.internal       [internal]     True    True         True           CloudEndpoint is active and ready
ngrok-operator-public-clep          https://ngrok-operator-public-clep.example.com     [public]       False   True         False          DNS_ERROR Reserved domain "ngrok-operator-public-clep.example.com" CNAME DNS record is not pointing at 39shjynerkafl5kxs.3kg7imcniwfgy1ozt.ngrok-cname.com. Started at 2025-11-06T04:25:58Z Retries at 2025-11-06T04:46:13Z
ngrok-operator-bindings-test-aep-default   ngrok-operator-bindings-test-aep.default   False   ProvisioningError   DNS_ERROR Reserved domain "ngrok-operator-bindings-test-aep.default" CNAME DNS record is not pointing at bht3tygql6bpeyvh.3kg7imcniwfgy1ozt.ngrok-cname.com. Started at 2025-11-06T04:29:29Z Retries at 2025-11-06T04:45:39Z
ngrok-operator-public-aep-example-com      ngrok-operator-public-aep.example.com      False   ProvisioningError   DNS_ERROR Reserved domain "ngrok-operator-public-aep.example.com" CNAME DNS record is not pointing at 4s5mnfledhnnk5u5r.3kg7imcniwfgy1ozt.ngrok-cname.com. Started at 2025-11-06T04:25:59Z Retries at 2025-11-06T04:46:14Z
ngrok-operator-public-clep-example-com     ngrok-operator-public-clep.example.com     False   ProvisioningError   DNS_ERROR Reserved domain "ngrok-operator-public-clep.example.com" CNAME DNS record is not pointing at 39shjynerkafl5kxs.3kg7imcniwfgy1ozt.ngrok-cname.com. Started at 2025-11-06T04:25:58Z Retries at 2025-11-06T04:46:13Z
```

Edit: Keeping the above notes for history but this has been resolved now!! thanks @jonstacks !

The issue was that during the rollout we spin up the new agent manager pod before tearing down the old one in order to keep things highly available. However there is no leader election in these pods so both try to do the same job. These pods started off as just read only pods that would launch a session in response to the CRD which works well. Then we added in the ability to create domains which also was fine since thats idempotent. 
The issue showed up here because now i'm trying to delete a domain while the other one is also reconciling and choosing to recreate it. 

The long term solution is to move any of this domain management and likely status writing stuff into a leader elected controller. 
For now, to help with this migration, it seems we can just fix the findAgentEndpointsForDomain function. Previously it would just check if a domain matched an endpoints domain ref. However there is a race condition:
- New pod starts spinning up. When it does it reconciles the agent endpoint with k8s binding and deletes the domain
- The old pod though sees this so it reconciles its agent endpoint and re-adds the domain
- The new pod does get the event but the agent endpoint it has cached has no domainRef since it just cleared it. So the findAgentEndpointsForDomain finds nothing that matches and doesn't reconcile.

The fix here is for the function to also match just on the URL matching between the 2 resources. This lets it trigger a reconcile of the agent endpoint again and delete the domain.

### DIFF
--- a/api/ngrok/v1alpha1/common_types.go
+++ b/api/ngrok/v1alpha1/common_types.go
@@ -83,7 +83,6 @@ func (ref *K8sObjectRefOptionalNamespace) Matches(obj client.Object) bool {
 		return false
 	}
 
-	// return true
 	ns := ptr.Deref(ref.Namespace, "")
 	return ns == "" || ns == obj.GetNamespace()
 }
--- a/internal/domain/manager.go
+++ b/internal/domain/manager.go
@@ -257,8 +257,5 @@ func (m *Manager) deleteStaleBindingDomain(ctx context.Context, endpoint ngrokv1
 	domain.SetName(domainRef.Name)
 
 	log.Info("Deleting stale domain for binding-based endpoint", "domain", client.ObjectKeyFromObject(domain), "endpoint", client.ObjectKeyFromObject(endpoint))
-	if err := m.Client.Delete(ctx, domain); client.IgnoreNotFound(err) != nil {
-		return err
-	}
-	return nil
+	return client.IgnoreNotFound(m.Client.Delete(ctx, domain))
 }
--- a/internal/controller/agent/agent_endpoint_controller.go
+++ b/internal/controller/agent/agent_endpoint_controller.go
@@ -238,7 +238,6 @@ func (r *AgentEndpointReconciler) update(ctx context.Context, endpoint *ngrokv1a
 		setTrafficPolicyCondition(endpoint, true, "TrafficPolicyApplied", "Traffic policy successfully applied")
 	}
 
-	// Update status (includes requeue check for domain readiness)
 	return r.updateStatus(ctx, endpoint, result, trafficPolicy, domainResult, nil)
 }
 
--- a/internal/controller/ngrok/cloudendpoint_controller.go
+++ b/internal/controller/ngrok/cloudendpoint_controller.go
@@ -200,7 +200,6 @@ func (r *CloudEndpointReconciler) create(ctx context.Context, clep *ngrokv1alpha
 	// Set success condition
 	setCloudEndpointCreatedCondition(clep, true, ReasonCloudEndpointCreated, "CloudEndpoint created successfully")
 
-	// Update status (includes requeue check for domain readiness)
 	return r.updateStatus(ctx, clep, ngrokClep, domainResult, nil)
 }
 
@@ -243,7 +242,6 @@ func (r *CloudEndpointReconciler) update(ctx context.Context, clep *ngrokv1alpha
 	// Set success condition
 	setCloudEndpointCreatedCondition(clep, true, ReasonCloudEndpointCreated, "CloudEndpoint updated successfully")
 
-	// Update status (includes requeue check for domain readiness)
 	return r.updateStatus(ctx, clep, ngrokClep, domainResult, nil)
 }
 


api/ngrok/v1alpha1/common_types.go
```diff
@@ -68,3 +68,22 @@ func (ref *K8sObjectRefOptionalNamespace) ToClientObjectKey(defaultNamespace str
...
+	if ref.Name != obj.GetName() {
+		return false
+	}
+
+	// nil or empty namespace means same namespace as the object
+	if ref.Namespace != nil && *ref.Namespace != "" && *ref.Namespace != obj.GetNamespace() {
+		return false
+	}
+
+	return true
```
jonstacks: *nit*:

```suggestion
	ns := ptr.Deref(ref.Namespace, "")	
	return ns == "" || ns == obj.GetNamespace()
```

internal/domain/manager_test.go
```diff
@@ -95,7 +96,10 @@ func TestManager_EnsureDomainExists_SkipsTCPDomains(t *testing.T) {
 
 	endpoint := createTestEndpoint("test-endpoint", "default", "tcp://1.tcp.ngrok.io:12345")
 
-	result, err := manager.EnsureDomainExists(context.TODO(), endpoint, endpoint.Spec.URL)
+	result, err := manager.EnsureDomainExists(context.TODO(), endpoint, DomainCheckParams{
```
jonstacks: We're on a new enough version of go that we can use `t.Context()` for this

internal/domain/manager.go
```diff
@@ -49,13 +64,13 @@ type Manager struct {
 }
 
 // EnsureDomainExists checks if the Domain CRD exists, creates it if needed, and sets conditions/domainRef
-func (m *Manager) EnsureDomainExists(ctx context.Context, endpoint ngrokv1alpha1.EndpointWithDomain, url string) (*DomainResult, error) {
-	parsedURL, err := m.parseAndValidateURL(endpoint, url)
+func (m *Manager) EnsureDomainExists(ctx context.Context, endpoint ngrokv1alpha1.EndpointWithDomain, params DomainCheckParams) (*DomainResult, error) {
```
jonstacks: It looks like everywhere this is called we're passing `DomainCheckParams` with the Endpoint's `spec.URL` and `spec.Bindings`. We're also already passing in the `EndpointWithDomain` interface in. It seems like both CloudEndpoints and AgentEndpoints both have a URL and bindings. What do you think about something like https://github.com/ngrok/ngrok-operator/commit/189a23790db4f9d7cebf5c0766b2a2e384b66118

api/ngrok/v1alpha1/common_types.go
```diff
@@ -68,3 +71,19 @@ func (ref *K8sObjectRefOptionalNamespace) ToClientObjectKey(defaultNamespace str
...
+func (ref *K8sObjectRefOptionalNamespace) Matches(obj client.Object) bool {
+	if ref == nil {
+		return false
+	}
+
+	if ref.Name != obj.GetName() {
+		return false
+	}
+
+	// return true
```
jonstacks: *nit*: This looks like it was code that is commented out and not an actual comment. If so, let's remove it to avoid confusion.

jonstacks (approved):

PR #706, opened 2025-11-06
fix: kginkgo helpers

jonstacks:
## What

There are some issues with the `kginkgo` helpers I made. Some of them don't correctly take into account the types of the `client.Object` being passed in. This fixes it as well as pulls some changes out of a future PR.

## How
* Fix them by getting using `unstructured.Unstructured` to get the object back

## Breaking Changes
No, this is only impacts tests and our testutils package

### DIFF
--- a/internal/testutils/kginkgo.go
+++ b/internal/testutils/kginkgo.go
@@ -2,6 +2,7 @@ package testutils
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	ingressv1alpha1 "github.com/ngrok/ngrok-operator/api/ingress/v1alpha1"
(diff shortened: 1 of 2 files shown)


internal/testutils/kginkgo.go
```diff
@@ -162,9 +170,13 @@ func (k *KGinkgo) ExpectFinalizerToBeAdded(ctx context.Context, obj client.Objec
 
 	eo := makeKGinkgoOptions(opts...)
 	key := client.ObjectKeyFromObject(obj)
+	gvk := k.getGVKForObject(obj)
 
+	By(fmt.Sprintf("expecting finalizer %s to be added to %s/%s", finalizer, obj.GetNamespace(), obj.GetName()))
 	Eventually(func(g Gomega) {
-		fetched := &corev1.Service{}
```
jonstacks, line 167: This was only being used in service tests and hence it worked. Because we want it to be useful across other tests, we need use unstructured.

internal/testutils/kginkgo.go
```diff
@@ -2,21 +2,25 @@ package testutils
...
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
-	DefaultTimeout  = 20 * time.Second
+	DefaultTimeout  = 10 * time.Second
```
jonstacks, line 23: The thought here was that changing the default timeout to be lower will help keep our tests faster when using the helpers and not specifying a timeout. Tests taking longer will fail and make us look at them closer

alex-bezek (approved):

PR #707, opened 2025-11-06
feat: Add util documentation and tests

jonstacks:
## What

Pulling these changes out of an existing branch to make it easier to review.

When trying to use these functions, the language server didn't have any docs on them which was annoying. I noticed they also didn't have any tests. While auditing them, I also noticed that we had some redundant code that doesn't do anything.

## How
* Add documentation on the package level functions in `utils.go`
* Add tests for those functions from `utils.go`
* `RemoveAndSyncFinalizer` will only update the object if it has has the finalizer, so the existing checks are not needed.

## Breaking Changes
No

### DIFF
--- a/internal/controller/util.go
+++ b/internal/controller/util.go
@@ -1,3 +1,26 @@
+/*
+MIT License
+
+Copyright (c) 2025 ngrok, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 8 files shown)


internal/controller/util.go
```diff
@@ -11,39 +34,55 @@ const (
...
 func RemoveFinalizer(o client.Object) bool {
 	return controllerutil.RemoveFinalizer(o, FinalizerName)
 }
 
+// RegisterAndSyncFinalizer adds the ngrok finalizer to the object if not already present.
+// If it adds the finalizer, it updates the object in the Kubernetes API.
 func RegisterAndSyncFinalizer(ctx context.Context, c client.Writer, o client.Object) error {
-	if !HasFinalizer(o) {
-		AddFinalizer(o)
+	if AddFinalizer(o) {
```
jonstacks, line 70: `AddFinalizer` returns true if it modified the finalizers, so we can just check the return of that instead of also calling `HasFinalizer` first.

internal/controller/util.go
```diff
@@ -11,39 +34,55 @@ const (
...
 	}
 	return nil
 }
 
+// RemoveAndSyncFinalizer removes the ngrok finalizer from the object if present.
+// If it removes the finalizer, it updates the object in the Kubernetes API.
 func RemoveAndSyncFinalizer(ctx context.Context, c client.Writer, o client.Object) error {
-	RemoveFinalizer(o)
-	return c.Update(ctx, o)
+	if RemoveFinalizer(o) {
```
jonstacks, line 79: Same thing as above. We don't need to update the object unless the finalizer was removed. I think we could also eventually make this a patch instead of an update.

Issue #681, opened 2025-09-26
The domain resource fails to finalize when defaultDomainReclaimPolicy = Retain


glennpow:
I'm using the ngrok-operator with my free ngrok account.
I've manually created a custom domain in ngrok's dashboard, which I use when spinning up my local cluster.
Since I don't want ngrok-operator to delete this custom domain when spinning back down, I set `defaultDomainReclaimPolicy` to `Retain` in the helm values.

This appears to preserve the domain when everything is torn down, however, then the `domain.ingress.k8s.ngrok.com` resource fails to complete its finalizer, which prevents the application namespace from terminating cleanly and ultimately requires manual deletion to restart the cluster.  My best guess is that the domain resources isn't finalizing because of the `Retain` setting.

Is that the case?  Is that expected behavior, or is something likely misconfigured?  What might my options so that I can retain the custom domain in ngrok, but cleanly finalize the cluster?

jonstacks:
Hi @glennpow, can you provide us what version of the operator you are on or any relevant logs from the manager pod?

I've tried to repro this issue in my cluster by:

1. Creating a domain CR with a reclaim policy of `Retain`:
```yaml
# test-domain-5.yaml
apiVersion: ingress.k8s.ngrok.com/v1alpha1
kind: Domain
metadata:
  name: stacks-test-app-5-ngrok-io
  namespace: default
spec:
  description: Created by the ngrok-operator
  domain: stacks-test-app-5.ngrok.io
  reclaimPolicy: Retain
```
2. Applying the manifest with `kubectl apply -f test-domain-5.yaml`
3. Waiting for the ngrok-operator to create it in ngrok and populate the ID field: `kubectl get domain stacks-test-app-5-ngrok-io`
```sh
NAME                         ID                               RECLAIM POLICY   REGION   DOMAIN                       AGE
stacks-test-app-5-ngrok-io   rd_33F0S4nBSpSocLZgdWwbNoYtxyR   Retain                    stacks-test-app-5.ngrok.io   8s
```
4. Delete it with `kubectl delete domain stacks-test-app-5-ngrok-io`. This returns immediately for me and doesn't hang. I can even verify it is delete by trying to get it again:
```sh
Error from server (NotFound): domains.ingress.k8s.ngrok.com "stacks-test-app-5-ngrok-io" not found
```

From my testing, I don't think it has to due with the reclaim policy of `Retain`. 

[...]

Issue #597, opened 2025-03-01
Possible race condition?


felipeamarante:
### Kubernetes Version

Server Version: v1.30.0

### Helm Chart Version

ngrok-operator-0.17.0

### Operator Configuration

Not found?

### System Info

Running Kind k8s on Podman.

### Kubernetes Logs

```
W0301 17:43:31.805641       1 reflector.go:569] k8s.io/client-go@v0.32.1/tools/cache/reflector.go:251: failed to list *v1alpha1.AgentEndpoint: the server could not find the requested resource (get agentendpoints.ngrok.k8s.ngrok.com)
E0301 17:43:31.805847       1 reflector.go:166] "Unhandled Error" err="k8s.io/client-go@v0.32.1/tools/cache/reflector.go:251: Failed to watch *v1alpha1.AgentEndpoint: failed to list *v1alpha1.AgentEndpoint: the server could not find the requested resource (get agentendpoints.ngrok.k8s.ngrok.com)" logger="UnhandledError"
W0301 17:44:15.232620       1 reflector.go:569] k8s.io/client-go@v0.32.1/tools/cache/reflector.go:251: failed to list *v1alpha1.Tunnel: the server could not find the requested resource (get tunnels.ingress.k8s.ngrok.com)
E0301 17:44:15.232781       1 reflector.go:166] "Unhandled Error" err="k8s.io/client-go@v0.32.1/tools/cache/reflector.go:251: Failed to watch *v1alpha1.Tunnel: failed to list *v1alpha1.Tunnel: the server could not find the requested resource (get tunnels.ingress.k8s.ngrok.com)" logger="UnhandledError"
W0301 17:44:31.612877       1 reflector.go:569] k8s.io/client-go@v0.32.1/tools/cache/reflector.go:251: failed to list *v1alpha1.AgentEndpoint: the server could not find the requested resource (get agentendpoints.ngrok.k8s.ngrok.com)
E0301 17:44:31.613076       1 reflector.go:166] "Unhandled Error" err="k8s.io/client-go@v0.32.1/tools/cache/reflector.go:251: Failed to watch *v1alpha1.AgentEndpoint: failed to list *v1alpha1.AgentEndpoint: the server could not find the requested resource (get agentendpoints.ngrok.k8s.ngrok.com)" logger="UnhandledError"
[...]

Issue #563, opened 2025-01-08
Add docs about how to avoid hanging finalizers during shut down


kaganjd:
### Description

If you are running the operator in the same namespace as your application, and you delete the entire namespace, you may end up with hanging finalizers like this:
```
  NamespaceContentRemaining                    True    Tue, 07 Jan 2025 14:55:19 -0800  SomeResourcesRemain   Some resources are remaining: domains.ingress.k8s.ngrok.com has 1 resource instances, httpsedges.ingress.k8s.ngrok.com has 1 resource instances, ingresses.networking.k8s.io has 1 resource instances, tunnels.ingress.k8s.ngrok.com has 1 resource instances
  NamespaceFinalizersRemaining                 True    Tue, 07 Jan 2025 14:55:19 -0800  SomeFinalizersRemain  Some content in the namespace has finalizers remaining: k8s.ngrok.com/finalizer in 4 resource instances
```

We should add docs suggesting:
1. running the ngrok-operator in its own namespace
2. shut down ngrok-operator namespace first, then application namespace 
3. if this isn't possible, run a script to delete all the finalizers on ngrok custom resources. @Alice-Lilith had the idea to make this a Helm `pre-delete` hook

script:
1. first, set your namespace: namespace=set-your-namespace
2. then run:
```bash
declare -a crds=("domains" "ingresses" "tunnels" "httpsedges")

for i in "${crds[@]}"; do for j in $(kubectl get $i -n $namespace -o name | cut -d '/' -f 2-); do kubectl patch $i $j -n $namespace -p '{"metadata":{"finalizers":[]}}' --type=merge; done; done
```

jonstacks:
This should also be fixed by #750 and will go out with the next release. We currently adding docs https://github.com/ngrok/ngrok-docs/pull/1896/changes for this as well

PR #709, opened 2025-11-07
feat: helm pre-delete hook

jonstacks:
Relates to:
* #681
* #563 

### Background
The ngrok-operator watches `Service[type=Loadbalancer]`, `Ingress`, Gateway API resources, and ngrok CRD types. When it creates resources in the ngrok API for these, it adds the ngrok finalizer to these resources to make sure that those ngrok resources get cleaned up on deletion of the resources.
These controllers are responsible for removing the finalizers from the resources they manage after cleaning up the ngrok resources. 

When the manager pods start terminating, their context is cancelled and they stop reconciling resources. Once these manager pods are terminated they can no longer cleanup resources or remove the finalizers.

### Problem

When uninstalling the helm chart, there is a race condition where the manager pods are deleted before they have time to clean up the resources they owned. This leaves resources in the cluster with the ngrok finalizer present, but no way remove the finalizer. This makes the uninstall fail and wedges these resources in a bad state.

What is more, is that if you you run helm uninstall and some of the resources are cleaned up, but others like the `KubernetesOperator` don't get processed in time, when you try to re-install, the new manager pods will clean up the finalizer from the existing resources and then the CRD will be deleted after install, leaving you in a bad state. This is what may be happening happening in #597. 

### Helm Hook

This PR adds a helm pre-delete hook, initially disabled by default in the chart while we continue to test it. The pre-delete hook installs a cluster role to give the helm delete hook permissions to find resources with the ngrok finalizer and then annotation those resources with a `k8s.ngrok.com/cleanup: "true"` annotation.

### `k8s.ngrok.com/cleanup: "true"` annotation support for controllers

This PR adds support for handling a `k8s.ngrok.com/cleanup: "true"` annotation to our controllers. When objects are annotated with this annotation:

* The ngrok resources are deleted
* Any internal annotations are cleaned up
* The finalizer is removed
* The status conditions are appropriately reset

### Alternatives considered

I was contemplating just triggering deletes for all of the resources that have our finalizer and letting our controllers handle the cleanup, removing the annotation, and then letting them be garbage collected. The reason I didn't opt for this approach is that we don't create Ingress, Service, and Gateway API resources. They are created by users. It feels unexpected for us to delete them since we didn't create them.

### Additional Concerns

This is currently only implemented as a helm pre-delete hook. This does not cover the case where the manager deployments or ngrok-operator namespace is deleted directly. That case is difficult to handle. When the namespace id deleted, it starts deleting all namespaced resources including the service account which grants the managers permissions to remove the finalizers. It also terminates the manager pods which prevent finalizers on resources in different namespaces from being cleaned up.

I have some other thoughts on how we can handle some of these cases that will make use of the annotation work, but need to do more testing. This also only works when using helm since this hook isn't supported by other tools like ArgoCD, for example. I think the annotation work can be kept, but we might need to supplement the pre-delete hook with other methods to make uninstall cleanup easier.

## Breaking Changes
There should be no breaking changes. This feature is disabled by default currently in the helm chart and has to be enabled with `cleanupHook.enabled=true` (We do enable it for our devenv to test it). Nothing in the codebase annotates these currently, so those paths should be unaffected.

### DIFF
--- a/docs/cleanup-hook.md
+++ b/docs/cleanup-hook.md
@@ -1,0 +1,131 @@
+# Cleanup Hook
+
+## Overview
+
+The ngrok-operator includes an optional Helm pre-delete hook that ensures proper cleanup of ngrok resources when
+uninstalling the operator. Without this hook, the operator's deployment and pods may be deleted before they can
+clean up ngrok API resources, leaving orphaned resources in your ngrok account.
+
+## How It Works
+
+When `helm uninstall` is executed, the cleanup hook:
+
+1. **Runs before deletion**: As a `pre-delete` hook, it executes before any operator resources are removed
+2. **Annotates resources**: Adds `k8s.ngrok.com/cleanup=true` to all Kubernetes resources managed by the operator that have the ngrok finalizer
+3. **Processes in order**:
+   - Gateway API Routes (HTTPRoute, TCPRoute, TLSRoute)
+   - Core resources (Ingress, Service)
+   - Gateway API Gateways
+   - ngrok CRDs (CloudEndpoint, AgentEndpoint, Domain, IPPolicy, etc.)
+4. **Waits for cleanup**: Monitors each resource until the operator removes the finalizer, indicating cleanup is complete
+5. **Retries on failure**: Automatically retries operations if they fail
+
+This ensures the operator managers stay running long enough to properly clean up all ngrok resources before the operator itself is removed.
+
+## Configuration
+
+The cleanup hook is configured via Helm values:
+
+```yaml
+cleanupHook:
+  # Enable or disable the cleanup hook
+  enabled: true
+
+  # Maximum time (in seconds) to wait for all resources to be cleaned up
+  timeout: 300  # 5 minutes
+
+  # Number of times to retry on failure
+  retries: 3
+
+  # Time to wait (in seconds) between retries
+  retryInterval: 10  # 10 seconds
+
+  # Resource requests/limits for the cleanup job pod
+  resources:
+    limits:
+      cpu: 100m
+      memory: 128Mi
+    requests:
+      cpu: 50m
+      memory: 64Mi
+```
+
+## When to Disable
+
+You may want to disable the cleanup hook if:
+
+- You want to retain ngrok resources after uninstalling the operator
+- You're troubleshooting and need to prevent automatic cleanup
+- You have a custom cleanup process
+
+To disable:
+
+```bash
+helm install ngrok-operator ngrok/ngrok-operator \
+  --set cleanupHook.enabled=false
+```
+
+## Troubleshooting
+
+### Timeout Issues
+
+If the cleanup hook times out, you can increase the timeout:
+
+```yaml
+cleanupHook:
+  timeout: 600s  # 10 minutes
+```
+
+### Hook Failures
+
+Check the cleanup job logs:
+
+```bash
+kubectl logs -n ngrok-operator job/ngrok-operator-cleanup
+```
+
+### Manual Cleanup
+
+If the hook fails, you can manually trigger cleanup by annotating resources:
+
+```bash
+# Annotate a specific ingress
+kubectl annotate ingress my-ingress k8s.ngrok.com/cleanup=true
+
+# Annotate all services with the ngrok finalizer
+kubectl get svc -A -o json | \
+  jq -r '.items[] | select(.metadata.finalizers[] | contains("k8s.ngrok.com/finalizer")) | "\(.metadata.namespace)/\(.metadata.name)"' | \
+  xargs -I {} kubectl annotate svc {} k8s.ngrok.com/cleanup=true
+```
+
+## Resource Processing Order
+
+The hook processes resources in a specific order to handle dependencies:
+
+1. **Routes first**: Gateway API routes are cleaned up before their parent gateways
+2. **Core resources**: Ingress and Service resources that create ngrok CRDs
+3. **Gateways**: After routes are cleaned up
+4. **ngrok CRDs**: Finally, any remaining operator-managed custom resources
+
+This ordering ensures that dependent resources are cleaned up before their dependencies, preventing validation errors.
+
+## Permissions
+
+The cleanup hook requires cluster-wide permissions to:
+- List and update all resource types managed by the operator
+- Check if optional CRDs (like Gateway API) exist
+
+These permissions are automatically granted via the `ngrok-operator-cleanup` ClusterRole, which is created as part of the hook.
+
+## Implementation Details
+
+The cleanup hook is implemented as:
+- A Kubernetes Job with `helm.sh/hook: pre-delete`
+- Uses a bash script (stored in ConfigMap) with kubectl to annotate and monitor resources
+- Runs in a minimal `bitnami/kubectl` container image
+- ServiceAccount with ClusterRole permissions
+- Automatic cleanup via `helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded`
+
+The bash script is stored at `helm/ngrok-operator/scripts/cleanup.sh` and can be reviewed or customized before installation.
+
+The hook will be automatically removed after successful completion or before the next hook execution.
--- a/helm/ngrok-operator/scripts/cleanup.sh
+++ b/helm/ngrok-operator/scripts/cleanup.sh
@@ -1,0 +1,315 @@
+#!/bin/bash
+set -euo pipefail
+
+# Configuration from environment variables
+TIMEOUT="${CLEANUP_TIMEOUT:-300}"
+RETRIES="${CLEANUP_RETRIES:-3}"
+RETRY_INTERVAL="${CLEANUP_RETRY_INTERVAL:-10}"
+POLL_INTERVAL="${POLL_INTERVAL:-2}"
+FINALIZER="k8s.ngrok.com/finalizer"
+CLEANUP_ANNOTATION="k8s.ngrok.com/cleanup"
+
+log() {
+  echo "[$(date -u +"%Y-%m-%dT%H:%M:%SZ")] $*"
+}
+
+log_error() {
+  echo "[$(date -u +"%Y-%m-%dT%H:%M:%SZ")] ERROR: $*" >&2
+}
+
+# Check if a CRD exists
+crd_exists() {
+  local group="$1"
+  local version="$2"
+  local resource="$3"
+
+  kubectl get crd "${resource}.${group}" &>/dev/null
+}
+
+# Check if a resource type is available (handles both CRDs and core types)
+resource_type_exists() {
+  local api_version="$1"
+  local kind="$2"
+
+  # For core types (empty group), just try to list
+  if [[ "$api_version" == "v1" ]] || [[ "$api_version" == "networking.k8s.io/v1" ]]; then
+    return 0
+  fi
+
+  # For CRDs, check if they exist
+  local group="${api_version%/*}"
+  local plural="${kind,,}s"  # Simple pluralization
+  # Handle special plural forms.
+  # Example: ends in 'y' -> 'ies'
+  if [[ "$kind" == *y ]]; then
+    plural="${kind%y}ies"
+  elif [[ "$kind" == "Ingress" ]]; then
+    plural="ingresses"
+  fi
+  kubectl get crd "${plural}.${group}" &>/dev/null
+}
+
+# List resources with the ngrok finalizer
+list_resources_with_finalizer() {
+  local api_version="$1"
+  local kind="$2"
+  local namespaced="${3:-true}"
+
+  local output
+  if [[ "$namespaced" == "true" ]]; then
+    if ! output=$(kubectl get "${kind}" -A -o json 2>&1); then
+      log_error "Failed to list ${kind} resources: ${output}"
+      return 1
+    fi
+    echo "$output" | jq -r --arg finalizer "$FINALIZER" \
+      '.items[] | select(.metadata.finalizers[]? == $finalizer) | "\(.metadata.namespace)/\(.metadata.name)"'
+  else
+    if ! output=$(kubectl get "${kind}" -o json 2>&1); then
+      log_error "Failed to list ${kind} resources: ${output}"
+      return 1
+    fi
+    echo "$output" | jq -r --arg finalizer "$FINALIZER" \
+      '.items[] | select(.metadata.finalizers[]? == $finalizer) | .metadata.name'
+  fi
+}
+
+# Annotate a resource for cleanup
+annotate_resource() {
+  local kind="$1"
+  local namespace="$2"
+  local name="$3"
+
+  if [[ -n "$namespace" ]]; then
+    # Check if already annotated
+    local current_annotation
+    current_annotation=$(kubectl get "${kind}" -n "${namespace}" "${name}" -o jsonpath="{.metadata.annotations['${CLEANUP_ANNOTATION}']}" 2>/dev/null || echo "")
+
+    if [[ "$current_annotation" == "true" ]]; then
+      log "Resource ${kind}/${namespace}/${name} already has cleanup annotation"
+      return 0
+    fi
+
+    log "Annotating ${kind}/${namespace}/${name} for cleanup"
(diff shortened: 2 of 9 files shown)


helm/ngrok-operator/templates/cleanup-hook/job.yaml
```diff
@@ -0,0 +1,66 @@
...
+kind: Job
+metadata:
+  name: {{ include "ngrok-operator.fullname" . }}-cleanup
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "ngrok-operator.labels" . | nindent 4 }}
+    app.kubernetes.io/component: cleanup-hook
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "-5"
```
alex-bezek, line 12: Out of curiosity, are the `-5` and `-10` arbitrary outside the fact that we need the config map, service account, etc to install before the job?

docs/cleanup-hook.md
```diff
@@ -0,0 +1,131 @@
...
+kubectl logs -n ngrok-operator job/ngrok-operator-cleanup
+```
+
+### Manual Cleanup
+
+If the hook fails, you can manually trigger cleanup by annotating resources:
+
+```bash
+# Annotate a specific ingress
+kubectl annotate ingress my-ingress k8s.ngrok.com/cleanup=true
```
alex-bezek, line 93: would it be worth including this script in our manager pod docker image so someone could easily run the shell script without actually uninstalling the whole chart? maybe an idea for later?

helm/ngrok-operator/templates/cleanup-hook/job.yaml
```diff
@@ -0,0 +1,66 @@
...
+      restartPolicy: Never
+      serviceAccountName: {{ include "ngrok-operator.fullname" . }}-cleanup
+      {{- with .Values.image.pullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+      - name: cleanup
+        # Use kubectl image for running the cleanup script
+        image: bitnami/kubectl:latest
```
alex-bezek, line 33: should we pin this to a specific tag instead of using latest? I doubt it would cause a problem, but it's generally not best practice to use latest typically.

helm/ngrok-operator/scripts/cleanup.sh
```diff
@@ -0,0 +1,315 @@
...
+    # Gateway API Gateways
+    "Gateway|gateway.networking.k8s.io/v1|gateway|true|true"
+
+    # ngrok CRDs
+    "CloudEndpoint|ngrok.k8s.ngrok.com/v1alpha1|cloudendpoint|true|false"
+    "AgentEndpoint|ngrok.k8s.ngrok.com/v1alpha1|agentendpoint|true|false"
+    "Domain|ingress.k8s.ngrok.com/v1alpha1|domain|true|false"
+    "IPPolicy|ingress.k8s.ngrok.com/v1alpha1|ippolicy|true|false"
+    "NgrokTrafficPolicy|ngrok.k8s.ngrok.com/v1alpha1|ngroktrafficpolicy|true|false"
+    "KubernetesOperator|ngrok.k8s.ngrok.com/v1alpha1|kubernetesoperator|true|false"
```
alex-bezek, line 290: do we also need to clean up BoundEndpoints since they have finalizers? 

```
kubectl get boundendpoints -A -o yaml | grep finalizer -A 2
    finalizers:
    - k8s.ngrok.com/finalizer
    generation: 1
    name: ngrok-1708d1e0-0982-55c5-8c11-668dee45634b
--
    finalizers:
    - k8s.ngrok.com/finalizer
    generation: 1
    name: ngrok-d30e730c-d44b-5746-a308-a58507613e80
```

In theory when the cloud endpoints and agent endpoints get deleted the boundendpoints should get removed by the operator once the API reflects it, but that seems like it would be racey. I see the `boundendpoint_controller` got updated with the status clearing method, but I don't see it listed here or in the RBAC file

internal/controller/gateway/gateway_controller_test.go
```diff
@@ -84,6 +86,50 @@ var _ = Describe("Gateway controller", Ordered, func() {
 						g.Expect(obj.Status.Addresses[0].Value).To(Equal(domain))
 					})
 				})
+
+				When("the gateway is annoated for cleanup", func() {
```
alex-bezek:
```suggestion
				When("the gateway is annotated for cleanup", func() {
```

internal/controller/gateway/httproute_controller_test.go
```diff
@@ -103,6 +105,42 @@ var _ = Describe("HTTPRoute controller", Ordered, func() {
 
 					}, timeout, interval).Should(Succeed())
 				})
+
+				When("the httproute is annoated for cleanup", func() {
```
alex-bezek:
```suggestion
				When("the httproute is annotated for cleanup", func() {
```

helm/ngrok-operator/values.yaml
```diff
@@ -428,3 +428,40 @@ bindings:
...
+## @param cleanupHook.enabled Enable the pre-delete cleanup hook
+## @param cleanupHook.timeout Maximum time to wait for cleanup to complete
+## @param cleanupHook.retries Number of times to retry on failure
+## @param cleanupHook.retryInterval Time to wait between retries
+##
+cleanupHook:
+  enabled: false
+  timeout: "300" # Timeout in seconds
+  retries: "3"
+  retryInterval: "10" # Retry interval in seconds
```
alex-bezek: Nit: these seem like they should be int's not strings so the values.schema.json file validates them correctly

internal/annotations/annotations.go
```diff
@@ -46,6 +46,9 @@ const (
 	TrafficPolicyAnnotation    = "k8s.ngrok.com/traffic-policy"
 	TrafficPolicyAnnotationKey = "traffic-policy"
 
+	CleanupAnnotation    = "k8s.ngrok.com/cleanup"
+	CleanupAnnotationKey = "cleanup"
```
alex-bezek: This looks like it's unused

internal/controller/base_controller.go
```diff
@@ -73,27 +78,14 @@ func (self *BaseController[T]) Reconcile(ctx context.Context, req ctrl.Request,
 
 	log.V(1).Info("Reconciling Resource", "ID", self.StatusID(obj))
 
-	if IsUpsert(obj) {
-		if err := RegisterAndSyncFinalizer(ctx, self.Kube, obj); err != nil {
-			return ctrl.Result{}, err
-		}
+	// If the cleanup annotation is present and the finalizer has been removed, we've already clened up
```
alex-bezek:
```suggestion
	// If the cleanup annotation is present and the finalizer has been removed, we've already cleaned up
```

helm/ngrok-operator/scripts/cleanup.sh
```diff
@@ -0,0 +1,315 @@
...
+log() {
+  echo "[$(date -u +"%Y-%m-%dT%H:%M:%SZ")] $*"
+}
+
+log_error() {
+  echo "[$(date -u +"%Y-%m-%dT%H:%M:%SZ")] ERROR: $*" >&2
+}
+
+# Check if a CRD exists
+crd_exists() {
```
alex-bezek, line 21: I don't see this being used anywhere

pkg/managerdriver/controller-handler.go
```diff
@@ -48,6 +56,14 @@ func (e *ControllerEventHandler) Create(_ context.Context, evt event.CreateEvent
...
 // Update is called in response to an update event -  e.g. Edge Updated.
 func (e *ControllerEventHandler) Update(ctx context.Context, evt event.UpdateEvent, _ workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+	if annotations.HasCleanupAnnotation(evt.ObjectNew) {
+		e.log.V(5).Info("object has cleanup annotation, skipping update", "object", evt.ObjectNew)
+		if err := e.store.Delete(evt.ObjectNew); err != nil {
+			e.log.Error(err, "error deleting object with cleanup annotation", "object", evt.ObjectNew)
+		}
+		return
+	}
+
```
alex-bezek, line 66: I can't comment down below, but should the `Generic` function also include these checks?

internal/controller/service/controller.go
```diff
@@ -244,6 +251,18 @@ func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		svc.Status = corev1.ServiceStatus{}
+		anns := svc.GetAnnotations()
+		delete(anns, annotations.ComputedURLAnnotation)
+		svc.SetAnnotations(anns)
+
+		if err := r.Client.Status().Update(ctx, svc); err != nil {
+			log.Error(err, "Failed to clear service status after cleanup")
+			return ctrl.Result{}, err
+		}
+
```
alex-bezek, line 265: This looks like we clear the status and call the Status().Update but we are also deleting the annotation and i don't see a call to update the resource with those annotation changes. Is deleting the computed url annotation even necessary if the object is marked to be cleaned up?

internal/controller/gateway/tlsroute_controller.go
```diff
@@ -77,28 +77,45 @@ func (r *TLSRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, err
 	}
 
-	tcpRoute, err = r.Driver.UpdateTLSRoute(tcpRoute)
```
alex-bezek: nit: unrelated to your change, but looks like there was some copy-pasta with `tcpRoute` and `tlsRoute` getting mixed up here in the past.

internal/controller/gateway/tlsroute_controller.go
```diff
@@ -77,28 +77,45 @@ func (r *TLSRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
-			log.Error(err, "Failed to remove finalizer")
+
+		tcpRoute.Status = gatewayv1alpha2.TLSRouteStatus{
+			RouteStatus: gatewayv1alpha2.RouteStatus{
+				Parents: []gatewayv1alpha2.RouteParentStatus{},
+			},
+		}
+
+		if err := r.Client.Status().Update(ctx, tcpRoute); err != nil {
+			log.Error(err, "Failed to update TLSRoute status")
```
alex-bezek: comparing the tlsroute, tcproute, and httproute updates, this looks like the only one that clears the route status and calls the status().update function.

internal/controller/gateway/tlsroute_controller.go
```diff
@@ -77,28 +77,45 @@ func (r *TLSRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+
+	tcpRoute, err = r.Driver.UpdateTLSRoute(tcpRoute)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// The object is not being deleted, so register and sync finalizer
+	if err := controller.RegisterAndSyncFinalizer(ctx, r.Client, tcpRoute); err != nil {
+		log.Error(err, "Failed to register finalizer")
+		return ctrl.Result{}, err
```
alex-bezek, line 118: super minor nit: i'm just comparing the different route controllers and noticed these do the same operations but in a different order
image

I'm assuming updating the store before removing the finalizer from the object won't really affect anything though fwiw

internal/controller/agent/agent_endpoint_controller_test.go
```diff
@@ -45,6 +47,55 @@ import (
...
+	return ns + "/" + name
+}
+
+// AgentDriverDeletedEndpoint returns a function that checks if the mock driver has deleted the given endpoint.
+// It is meant to be used with Eventually in tests.
+//
+// Example usage:
+//
+//	Eventually(AgentDriverDeletedEndpoint(myEndpoint), timeout, interval).Should(BeTrue())
+func AgentDriverDeletedEndpoint(endpoint *ngrokv1alpha1.AgentEndpoint) func() bool {
```
alex-bezek, line 62: nice helper functions and great doc strings!

internal/controller/ngrok/cloudendpoint_controller.go
```diff
@@ -136,9 +139,14 @@ func (r *CloudEndpointReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&ngrokv1alpha1.CloudEndpoint{}, builder.WithPredicates(
-			predicate.GenerationChangedPredicate{},
-		)).
+		For(&ngrokv1alpha1.CloudEndpoint{},
+			builder.WithPredicates(
+				predicate.Or(
+					predicate.GenerationChangedPredicate{},
+					predicate.AnnotationChangedPredicate{},
```
alex-bezek, line 146: nice catch, these predicate filters can cause tricky bugs :)

alex-bezek (commented):
This looks pretty good to me 👍 I mostly had some nits, comments about consistency, and some questions

My only other thought was that the helm templates NOTES.txt might be a spot we could add information about how this works and how to avoid issues. But... it is already quite verbose so i'm not sure how effective that would be

helm/ngrok-operator/templates/cleanup-hook/job.yaml
jonstacks: Yes, mainly just so that we have the configmap and service account in place before the job runs. The weights are discussed at the bottom of [this section](https://helm.sh/docs/topics/charts_hooks/#writing-a-hook) in the helm docs

helm/ngrok-operator/templates/cleanup-hook/job.yaml
jonstacks: probably worth at least moving it to the values.yaml so it can be customized. Looking at the available [docker tags](https://hub.docker.com/r/bitnami/kubectl/tags?page=5), I can't tell if they are deleting older tags which may cause more problems.

docs/cleanup-hook.md
jonstacks: Our manager pod doesn't have bash in it right now and is based on `gcr.io/distroless/static:nonroot`. I opted for not including it and putting the script into a configmap and only pulling the `bitnami/kubectl` image when we need it as part of the cleanup job. 

The other option I explored was making a cleanup binary and putting that into the image, but we'd also have to give the manager a service account with elevated permissions all the time to be able to run it when it needs, which didn't feel great.

internal/controller/gateway/tlsroute_controller.go
jonstacks: Good catch. I'll fix this in a PR and rebase the changes off that

jonstacks: Have https://github.com/ngrok/ngrok-operator/pull/739 open to fix this

internal/controller/gateway/gateway_controller_test.go
jonstacks: Good catch. Fixed

internal/controller/gateway/httproute_controller_test.go
jonstacks: Fixed

helm/ngrok-operator/values.yaml
jonstacks: Fixed

internal/annotations/annotations.go
jonstacks: Good catch. Removed it.

internal/controller/base_controller.go
jonstacks: Fixed

internal/controller/gateway/tlsroute_controller.go
jonstacks: Good catch, I'll change the order around for the tls controller so that its consistent with the others.

pkg/managerdriver/controller-handler.go
jonstacks: I don't believe so. If its being deleted, the logic would be the same for deleting it from the store.

internal/controller/service/controller.go
jonstacks: When we remove and sync the finalizer below this, it will remove the annotation. There is a test for this that was added in the service controller.

jonstacks:
Closing in favor of #750

Issue #266, opened 2023-07-03
Redeploys or restarts never complete if ngrok sessions are limited to 1


jrobsonchase:
### What happened

A `deploy` or `rollout restart` causes a new replicaset to get created with the new pod spec in parallel with the old one. Once this is marked "Ready", the old one is terminated.

If agent sessions are limited to 1, as with free ngrok accounts today, this will never actually occur, since the ingress controller pod's rediness is contingent on the agent session starting successfully.

There are also no logs to indicate why the new pods get stuck.

### What you think should happen instead

Hard to say. Maybe we should detect the "too many sessions" error, and allow the pod to ready up if it persists for a certain amount of time? Forcing the old one to close before the new one can start will result in a blip in uptime, so we still want to wait on the session to fully start if we can, but we still shouldn't block forever in the event that it never will.

### How to reproduce

`make deploy` with a free account is sufficient, since it does a `rollout restart` to force a re-pull. You'll see two ingress controller pods and two replicasets. The deploy will list one as the "old" and the other as "new". The "new" one's pod will be stuck unready.

alex-bezek:
> Forcing the old one to close before the new one can start will result in a blip in uptime, so we still want to wait on the session to fully start if we can, but we still shouldn't block forever in the event that it never will.

Is it possible to wait for the new session to start before closing the old one if there is a limit of 1 agent session for free accounts? If you are limited to a single active session, it seems we'll always have to tear one down before creating the new one which will always have a small blip of downtime for free accounts. 

We technically could make a helm configuration that allows free accounts to specify a [Recreate Rollout Strategy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#recreate-deployment) instead of a Rolling Update, but the DevX for free accounts to remember to add a specific helm override isn't the greatest.

jrobsonchase:
> Is it possible to wait for the new session to start before closing the old one if there is a limit of 1 agent session for free accounts? If you are limited to a single active session, it seems we'll always have to tear one down before creating the new one which will always have a small blip of downtime for free accounts.

That's basically what we're doing already, but there's no way to tell the first session that it's time to shut down while the new one is coming up. The new instance can start, but it'll never get marked Ready until its tunnel session starts, which means the original will never get told to shut down.

[...]

PR #711, opened 2025-11-11
Adding strategy type of RollingUpdate to prevent session doubling during deployments

sabrina-ngrok:
#266 

## What
Redeploys or restarts never complete if ngrok sessions are limited to 1

## How
Changes made: https://ampcode.com/threads/T-a0987420-692d-47c2-b885-0d2c7456b717

[agent/deployment.yaml](file:///Users/sabrina/workspace/ngrok-operator/helm/ngrok-operator/templates/agent/deployment.yaml#L17-L22) - Added rolling update strategy with maxSurge: 0 and maxUnavailable: 1 to prevent session doubling during deployments

[values.yaml](file:///Users/sabrina/workspace/ngrok-operator/helm/ngrok-operator/values.yaml#L307-L313) - Added default agent.lifecycle PreStop hook that sleeps for 15 seconds, giving ngrok time to gracefully disconnect

[agent/deployment.yaml](file:///Users/sabrina/workspace/ngrok-operator/helm/ngrok-operator/templates/agent/deployment.yaml#L124-L127) - Updated lifecycle reference to use agent.lifecycle with fallback to global lifecycle

This ensures old pods disconnect from ngrok before new ones start, preventing session limit issues during redeploys.

Created [agent-deployment_test.yaml](file:///Users/sabrina/workspace/ngrok-operator/helm/ngrok-operator/tests/agent-deployment_test.yaml) with tests that verify:

✅ Rolling update strategy has maxSurge: 0 and maxUnavailable: 1
✅ Default lifecycle PreStop hook sleeps for 15 seconds
✅ Global lifecycle value overrides agent.lifecycle when set
✅ Default terminationGracePeriodSeconds is 30
✅ Custom terminationGracePeriodSeconds can be configured
✅ Agent deployment not created when ingress.enabled: false

## Breaking Changes
[None that I know of :)]

### DIFF
--- a/helm/ngrok-operator/templates/agent/deployment.yaml
+++ b/helm/ngrok-operator/templates/agent/deployment.yaml
@@ -16,7 +16,7 @@ metadata:
 spec:
   replicas: {{ $agent.replicaCount }}
   {{- if $agent.updateStrategy }}
-  updateStrategy:
+  strategy:
     {{- toYaml $agent.updateStrategy | nindent 4 }}
   {{- end }}
   selector:
--- a/helm/ngrok-operator/tests/agent/deployment_test.yaml
+++ b/helm/ngrok-operator/tests/agent/deployment_test.yaml
@@ -17,7 +17,7 @@ tests:
   template: agent/deployment.yaml
   asserts:
   - equal:
-      path: spec.updateStrategy.type
+      path: spec.strategy.type
       value: RollingUpdate
 - it: Adds .Values.podLabels to the controller deployment podspec
   set:
--- a/helm/ngrok-operator/templates/bindings-forwarder/deployment.yaml
+++ b/helm/ngrok-operator/templates/bindings-forwarder/deployment.yaml
@@ -16,7 +16,7 @@ metadata:
 spec:
   replicas: {{ $forwarder.replicaCount }}
   {{- if $forwarder.updateStrategy }}
-  updateStrategy:
+  strategy:
     {{- toYaml $forwarder.updateStrategy | nindent 4 }}
   {{- end }}
   selector:
--- a/helm/ngrok-operator/tests/agent/__snapshot__/deployment_test.yaml.snap
+++ b/helm/ngrok-operator/tests/agent/__snapshot__/deployment_test.yaml.snap
@@ -22,6 +22,8 @@ Should match snapshot:
           app.kubernetes.io/component: agent
           app.kubernetes.io/instance: RELEASE-NAME
           app.kubernetes.io/name: ngrok-operator
+      strategy:
+        type: RollingUpdate
       template:
         metadata:
           annotations:
@@ -97,8 +99,6 @@ Should match snapshot:
                 allowPrivilegeEscalation: false
           serviceAccountName: RELEASE-NAME-ngrok-operator-agent
           terminationGracePeriodSeconds: 30
-      updateStrategy:
-        type: RollingUpdate
   2: |
     apiVersion: rbac.authorization.k8s.io/v1
     kind: ClusterRole
--- a/helm/ngrok-operator/tests/bindings-forwarder/__snapshot__/deployment_test.yaml.snap
+++ b/helm/ngrok-operator/tests/bindings-forwarder/__snapshot__/deployment_test.yaml.snap
@@ -22,6 +22,8 @@ Should match snapshot:
           app.kubernetes.io/component: bindings-forwarder
           app.kubernetes.io/instance: RELEASE-NAME
           app.kubernetes.io/name: ngrok-operator
+      strategy:
+        type: RollingUpdate
       template:
         metadata:
           annotations:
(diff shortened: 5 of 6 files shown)


helm/ngrok-operator/templates/agent/deployment.yaml
```diff
@@ -15,6 +15,11 @@ metadata:
     checksum/rbac: {{ $rbacChecksum }}
 spec:
   replicas: {{ $agent.replicaCount }}
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
```
jonstacks: This hardcodes the `strategy.type` to `RollingUpdate` which is the [default](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy). We want the user to be able to override the `strategy.type` if they wish in the values.yaml to be `Recreate`. We should also stick with the defaults for the `RollingUpdate` strategy, but allow them to be supplied by the user if they wish

helm/ngrok-operator/values.yaml
```diff
@@ -304,6 +304,15 @@ agent:
...
 
+  ## @param agent.lifecycle Lifecycle hooks for the agent container
+  ## ref: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
+  ## PreStop hook gives the agent time to gracefully disconnect from ngrok before termination
+  ##
+  lifecycle:
+    preStop:
+      exec:
+        command: ["/bin/sh", "-c", "sleep 15"]
+
```
jonstacks: Is there a reason that we need this over just letting the agent get the SIGTERM, and shutdown normally? It should have the default 30s grace period before it is force killed, giving it time to shut down gracefully

helm/ngrok-operator/values.yaml
```diff
@@ -298,12 +298,27 @@ agent:
...
   ## @param agent.terminationGracePeriodSeconds The amount of time to wait for the agent pod to gracefully terminate
   ## ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination
   ## Increase this value to give the operator more time to clean up finalizers during shutdown
   ##
   terminationGracePeriodSeconds: 30
 
+  ## @param agent.lifecycle Lifecycle hooks for the agent container
+  ## ref: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
+
```
sabrina-ngrok: Will remove this, seems like we don't need it.

helm/ngrok-operator/tests/agent-deployment_test.yaml
```diff
@@ -0,0 +1,92 @@
+suite: test agent-deployment
```
jonstacks: We already have a file for agent deployment tests https://github.com/ngrok/ngrok-operator/blob/bad8fd8e0e91f30f6ef9ae790a00a968c3709331/helm/ngrok-operator/tests/agent/deployment_test.yaml#L1

Can we move the relevant tests out of here and into that test file?

helm/ngrok-operator/templates/agent/deployment.yaml
```diff
@@ -15,6 +15,10 @@ metadata:
     checksum/rbac: {{ $rbacChecksum }}
 spec:
   replicas: {{ $agent.replicaCount }}
+  {{- if $agent.strategy }}
+  strategy:
+    {{- toYaml $agent.strategy | nindent 4 }}
+  {{- end }}
```
jonstacks, line 21: This looks good. We'll also want to make this change in this file https://github.com/ngrok/ngrok-operator/blob/main/helm/ngrok-operator/templates/bindings-forwarder/deployment.yaml#L38 since we're adding it to the bindings forwarder as well

helm/ngrok-operator/templates/agent/deployment.yaml
```diff
@@ -117,9 +121,9 @@ spec:
         volumeMounts:
         {{ toYaml .Values.extraVolumeMounts | nindent 10 }}
         {{- end }}
-        {{- if .Values.lifecycle }}
+        {{- if or .Values.lifecycle $agent.lifecycle }}
         lifecycle:
-        {{ toYaml .Values.lifecycle | nindent 10 }}
+        {{ toYaml (default $agent.lifecycle .Values.lifecycle) | nindent 10 }}
```
jonstacks: Let's undo these 2 changes since we don't have an agent specific lifecycle configuration yet.

helm/ngrok-operator/tests/agent/deployment_test.yaml
```diff
@@ -6,9 +6,6 @@ set:
   ingress:
     enabled: true
 tests:
-- it: Should match snapshot
-  asserts:
-  - matchSnapshot: {}
```
jonstacks, line 11: Whats the reason to remove this?

sabrina-ngrok: Was failing this test, found out needed to run command `make update-snapshots` for it to pass.

PR #718, opened 2025-12-01
Create NewTrafficPolicyFromJson deserializer checking for unknown keys and implement it in translating ingresses

sabrina-ngrok:
#K8SOP-147

## What
it looks like we don't handle traffic policy with an inbound phase appropriately. Instead we just discard all of it, leading to no user supplied policy being merged with our generated traffic policy.

This could just be an edge case that you only hit with the inbound key

## How
Amp suggested that we create a new JSON deserializer that checks the valid traffic policy keys (on_http_request, on_http_response, and on_tcp_connect) and have a bunch of error handling for edge cases such as this.

## Breaking Changes
¯\\(ツ)/¯

### DIFF
--- a/internal/trafficpolicy/policy_test.go
+++ b/internal/trafficpolicy/policy_test.go
@@ -202,6 +202,12 @@ func TestNewTrafficPolicyFromJSON(t *testing.T) {
 			expectError: true,
 			errContains: "failed to unmarshal traffic policy: invalid character 'i' looking for beginning of object key string. raw traffic policy: {invalid}",
 		},
+		{
+			name:        "valid JSON with invalid structure should error on parse including original traffic policy",
+			input:       `{"on_http_request": "not-an-array"}`,
+			expectError: true,
+			errContains: "failed to parse traffic policy: json: cannot unmarshal string into Go struct field TrafficPolicy.on_http_request of type []trafficpolicy.Rule. traffic policy: {\"on_http_request\": \"not-an-array\"}",
+		},
 	}
 
 	for _, tc := range tests {


internal/trafficpolicy/policy.go
```diff
@@ -78,11 +81,48 @@ func NewTrafficPolicy() *TrafficPolicy {
 	}
 }
 
+// validTrafficPolicyKeys are the only keys allowed at the top level of a traffic policy document.
+// Any other keys indicate a malformed policy that could lead to security issues if silently ignored.
+var validTrafficPolicyKeys = map[string]bool{
+	"on_http_request":  true,
+	"on_http_response": true,
+	"on_tcp_connect":   true,
+}
```
alex-bezek, line 90: I think this would work to throw an error if any phase other than than these 3 were supplied, however this changes a design principle we've had with traffic policy so far. The TrafficPolicy type today is just a raw json message https://github.com/ngrok/ngrok-operator/blob/e8d5d7d39c0bf89965339a6c44341e92a1260165/api/ngrok/v1alpha1/ngroktrafficpolicy_types.go#L45
We did this because new capabilities are continually being added to traffic policies and we don't want to have to make a operator code change, cut a release, and force users to upgrade in order to use them. 

This primarily affects all the new actions being added continually, however we are also considering adding more [phases](https://ngrok.com/docs/traffic-policy/concepts/phases#phases) such as an `on_http_error` phase. If we put this hardcoded list here, we'll have to keep it in sync with the product changes and users would need to update their operator just to use it.  

The issue that triggered this was that someone was using one of the old phases like `inbound`, `outbound` and it was failing silently. If the requirements are to protect against this situation, we could look explicitly for deprecated phase names and throw an error then which would still let new phases be added and consumed immediately. 

However, if the requirement is to protect against fat fingering `on_http_reques` or something that requires using an allowlist, this wouldn't work and is at odds with the goal of allowing new additions to traffic policy without forcing updates in the operator. This might be worth talking to @iambmelt about for requirements.

pkg/managerdriver/translate_ingresses.go
```diff
@@ -257,19 +256,17 @@ func (t *translator) ingressBackendToIR(ingress *netv1.Ingress, backend *netv1.I
...
 		}
 
-		var routeTrafficPolicy trafficpolicy.TrafficPolicy
-		if len(routePolicyCfg.Spec.Policy) > 0 {
-			if err := json.Unmarshal(routePolicyCfg.Spec.Policy, &routeTrafficPolicy); err != nil {
-				return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, routePolicyCfg.Spec.Policy)
-			}
+		routeTrafficPolicy, err := trafficpolicy.NewTrafficPolicyFromJSON(routePolicyCfg.Spec.Policy)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse traffic policy: %w", err)
```
jonstacks: I think we can just return the original error here since wrapping it isn't providing us much additional value here. It probably makes the error message more verbose since we use the same message in `NewTrafficPolicyFromJSON`

internal/trafficpolicy/policy.go
```diff
@@ -78,11 +81,48 @@ func NewTrafficPolicy() *TrafficPolicy {
...
+// if unknown keys are present (which would otherwise be silently ignored).
 func NewTrafficPolicyFromJSON(data []byte) (*TrafficPolicy, error) {
+	if len(data) == 0 {
+		return NewTrafficPolicy(), nil
+	}
+
+	// First unmarshal to a map to detect unknown keys
+	var rawPolicy map[string]json.RawMessage
+	if err := json.Unmarshal(data, &rawPolicy); err != nil {
+		return nil, fmt.Errorf("failed to parse traffic policy: %w", err)
```
jonstacks: I think we want to keep the original context from the error message we had:
```go
return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, routePolicyCfg.Spec.Policy)
```

In this case, we probably want to include the bytes here and at the bottom where we return the same error message

internal/trafficpolicy/policy.go
alex-bezek: I discussed this with @jonstacks and he pointed me to https://github.com/ngrok/ngrok-operator/blob/main/internal/trafficpolicy/policy.go#L67 and gave some context. 
While we do want the traffic policy to accept any action types without having to update our code, we have other requirements for the phases. We need to support 2 separate ingress objects each with the same host but different traffic policies. So in the code today, we have the different phases so we can unmarshall different policies and merge their rules. 

Since we only support these existing phases today, I think its a separate issue we can log to follow up with later to allow using additional phases without having to update and release the operator. @iambmelt can help us undersstand how important this is, if we plan to have like 1 or 2 new phases in the future, or if we plan to add a bunch of new ones.

iambmelt: Thanks for the tag @alex-bezek 

A few thoughts.
- **Avoiding silent failure:** As I understand it from reading this thread, the original issue was that using a deprecated phase like `inbound` caused us to silently drop the traffic policy, which seems definitely not the behavior we want. From that standpoint, failing fast on known bad phases seems better than ignoring them. 
- **Forward compat for new phases:** We do want to allow new additions to traffic policy without operator updates. A few principles that seem aligned with this:
    * Reject known disallowed/deprecated phases (e.g., `inbound`, `outbound`)
    * _Consider_ explicitly _allowing_ known supported phases (caveat: this seems like potentially a maintenance burden, so avoiding it if possible seems preferable)
    * Treat everything else as unrecognized, possibly legitimate inputs

So then the question: what do with unknowns?

At a glance, I think this comes down to if we can correctly merge the unrecognized inputs. Can we?

It seems safest to me to reject the known bad and allow all else.

alex-bezek: > the original issue was that using a deprecated phase like inbound caused us to silently drop the traffic policy, which seems definitely not the behavior we want. From that standpoint, failing fast on known bad phases seems better than ignoring them.

Cool, this PR addresses that 👍 

I'll log a separate issue for enhancing the operator to allow unknown phases in traffic policies to be used. From talking with @jonstacks we should be able to do something like:
- define the known phases the operator needs to parse out in order to add in specific rules (like adding forward internal to a cloud endpoints http request phase)
- take the unknown phases and just merge their rules if their keys match to allow using new phases without code changes.

internal/trafficpolicy/policy.go
sabrina-ngrok:
```suggestion
		return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, rawPolicy)
```

internal/trafficpolicy/policy_test.go
```diff
@@ -139,6 +139,86 @@ func TestContainsAction(t *testing.T) {
...
+			name:        "mix of valid and invalid keys should error",
+			input:       `{"on_http_request": [], "inbound": []}`,
+			expectError: true,
+			errContains: "inbound",
+		},
+		{
+			name:        "invalid JSON should error",
+			input:       `{invalid}`,
+			expectError: true,
+			errContains: "failed to parse",
```
sabrina-ngrok:
```suggestion
			errContains: "failed to unmarshal traffic policy: invalid character 'i' looking for beginning of object key string. raw traffic policy: map[]",
```

pkg/managerdriver/translate_ingresses.go
sabrina-ngrok:
```suggestion
			return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, routePolicyCfg.Spec.Policy)
```

internal/trafficpolicy/policy.go
sabrina-ngrok: Implemented in commit.

pkg/managerdriver/translate_ingresses.go
sabrina-ngrok: Implemented in commit

sabrina-ngrok:
I tested the changes in my local, but I was having issues with SSH to push my commit and it was a small change so I just did a commit via suggestions on here.

internal/trafficpolicy/policy.go
```diff
@@ -78,11 +81,48 @@ func NewTrafficPolicy() *TrafficPolicy {
...
+		return nil, fmt.Errorf("traffic policy contains unknown keys that would be ignored: %s; valid keys are: on_http_request, on_http_response, on_tcp_connect",
+			strings.Join(unknownKeys, ", "))
+	}
+
+	// Now unmarshal into the typed struct
 	tp := NewTrafficPolicy()
-	err := json.Unmarshal(data, tp)
-	return tp, err
+	if err := json.Unmarshal(data, tp); err != nil {
+		return nil, fmt.Errorf("failed to parse traffic policy: %w", err)
```
sabrina-ngrok:
```suggestion
		return nil, fmt.Errorf("failed to parse traffic policy: %w. traffic policy: %v", err, tp)
```

internal/trafficpolicy/policy.go
```diff
@@ -78,11 +81,48 @@ func NewTrafficPolicy() *TrafficPolicy {
...
+// if unknown keys are present (which would otherwise be silently ignored).
 func NewTrafficPolicyFromJSON(data []byte) (*TrafficPolicy, error) {
+	if len(data) == 0 {
+		return NewTrafficPolicy(), nil
+	}
+
+	// First unmarshal to a map to detect unknown keys
+	var rawPolicy map[string]json.RawMessage
+	if err := json.Unmarshal(data, &rawPolicy); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, rawPolicy)
```
jonstacks:
```suggestion
		return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, data)
```

If it fails to unmarshal, it looks like we'll just print the empty rawPolicy as opposed to what we are trying to unmarshal.

internal/trafficpolicy/policy.go
```diff
@@ -78,11 +81,48 @@ func NewTrafficPolicy() *TrafficPolicy {
...
+		return nil, fmt.Errorf("traffic policy contains unknown keys that would be ignored: %s; valid keys are: on_http_request, on_http_response, on_tcp_connect",
+			strings.Join(unknownKeys, ", "))
+	}
+
+	// Now unmarshal into the typed struct
 	tp := NewTrafficPolicy()
-	err := json.Unmarshal(data, tp)
-	return tp, err
+	if err := json.Unmarshal(data, tp); err != nil {
+		return nil, fmt.Errorf("failed to parse traffic policy: %w. traffic policy: %v", err, tp)
```
jonstacks:
```suggestion
		return nil, fmt.Errorf("failed to parse traffic policy: %w. traffic policy: %v", err, data)
```

pkg/managerdriver/translate_ingresses.go
```diff
@@ -257,19 +256,17 @@ func (t *translator) ingressBackendToIR(ingress *netv1.Ingress, backend *netv1.I
...
 		}
 
-		var routeTrafficPolicy trafficpolicy.TrafficPolicy
-		if len(routePolicyCfg.Spec.Policy) > 0 {
-			if err := json.Unmarshal(routePolicyCfg.Spec.Policy, &routeTrafficPolicy); err != nil {
-				return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, routePolicyCfg.Spec.Policy)
-			}
+		routeTrafficPolicy, err := trafficpolicy.NewTrafficPolicyFromJSON(routePolicyCfg.Spec.Policy)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, routePolicyCfg.Spec.Policy)
```
jonstacks:
```suggestion
			return nil, err
```

I think we get most of this error message from the `NewTrafficPolicyFromJSON` function. Wrapping seems like it'll just be overly verbose here. My suggestions is to just return this error here unless we add additional context.

internal/trafficpolicy/policy.go
```diff
@@ -78,11 +81,48 @@ func NewTrafficPolicy() *TrafficPolicy {
...
+// if unknown keys are present (which would otherwise be silently ignored).
 func NewTrafficPolicyFromJSON(data []byte) (*TrafficPolicy, error) {
+	if len(data) == 0 {
+		return NewTrafficPolicy(), nil
+	}
+
+	// First unmarshal to a map to detect unknown keys
+	var rawPolicy map[string]json.RawMessage
+	if err := json.Unmarshal(data, &rawPolicy); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, data)
```
jonstacks: Sorry I gave you a bad suggestion earlier. We want the raw traffic policy to be output as a string in both this case and the one at the end of the function so that the error messages are more helpful. Thanks for adding the test that helped catch this mistake on my part.

sabrina-ngrok:
```suggestion
		return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, rawPolicy)
```

internal/trafficpolicy/policy.go
```diff
@@ -78,11 +81,48 @@ func NewTrafficPolicy() *TrafficPolicy {
...
+		return nil, fmt.Errorf("traffic policy contains unknown keys that would be ignored: %s; valid keys are: on_http_request, on_http_response, on_tcp_connect",
+			strings.Join(unknownKeys, ", "))
+	}
+
+	// Now unmarshal into the typed struct
 	tp := NewTrafficPolicy()
-	err := json.Unmarshal(data, tp)
-	return tp, err
+	if err := json.Unmarshal(data, tp); err != nil {
+		return nil, fmt.Errorf("failed to parse traffic policy: %w. traffic policy: %v", err, data)
```
sabrina-ngrok:
```suggestion
		return nil, fmt.Errorf("failed to parse traffic policy: %w. traffic policy: %v", err, tp)
```

internal/trafficpolicy/policy_test.go
```diff
@@ -139,6 +139,86 @@ func TestContainsAction(t *testing.T) {
...
+			name:        "mix of valid and invalid keys should error",
+			input:       `{"on_http_request": [], "inbound": []}`,
+			expectError: true,
+			errContains: "inbound",
+		},
+		{
+			name:        "invalid JSON should error",
+			input:       `{invalid}`,
+			expectError: true,
+			errContains: "failed to unmarshal traffic policy: invalid character 'i' looking for beginning of object key string. raw traffic policy: [123 105 110 118 97 108 105 100 125]",
```
sabrina-ngrok:
```suggestion
			errContains: "failed to unmarshal traffic policy: invalid character 'i' looking for beginning of object key string. raw traffic policy: map[]",
```

internal/trafficpolicy/policy.go
sabrina-ngrok: Reverted back.

internal/trafficpolicy/policy_test.go
```diff
@@ -139,6 +139,86 @@ func TestContainsAction(t *testing.T) {
...
+			name:        "mix of valid and invalid keys should error",
+			input:       `{"on_http_request": [], "inbound": []}`,
+			expectError: true,
+			errContains: "inbound",
+		},
+		{
+			name:        "invalid JSON should error",
+			input:       `{invalid}`,
+			expectError: true,
+			errContains: "failed to unmarshal traffic policy: invalid character 'i' looking for beginning of object key string. raw traffic policy: map[]",
```
jonstacks: I think we would expect the original traffic policy `{invalid}` to be printed so the user knows what wasn't able to be parsed. I think what you had before was close, it just needed to be cast as a string rather than a `[]byte`

sabrina-ngrok: I tested out in my local and it seems that the only variation where we can get `{invalid}` as the output is if we cast `data` to string instead of `rawPolicy` because `rawPolicy` is coming up as empty, `data` has the information.

jonstacks: Sorry, that is what I meant when I left this suggestion https://github.com/ngrok/ngrok-operator/pull/718#discussion_r2590732380. I missed data being a `[]byte` instead of a string. If you cast `data` to a string for the error messages and update the invalid test, I think that is all that is needed before we can ship it.

internal/trafficpolicy/policy.go
sabrina-ngrok:
```suggestion
		return nil, fmt.Errorf("failed to unmarshal traffic policy: %w. raw traffic policy: %v", err, string(data))
```

internal/trafficpolicy/policy_test.go
sabrina-ngrok:
```suggestion
			errContains: "failed to unmarshal traffic policy: invalid character 'i' looking for beginning of object key string. raw traffic policy: {invalid}",
```

PR #722, opened 2025-12-05
Remove the deprecated domain annotation on Load balancer services in …

alex-bezek:
…favor of the url annotation

## What

This has been deprecated for a while in favor of the k8s.ngrok.com/url annotation. We should remove this

## How

Simply removes the old code paths and updates some comments and documentation. 

## Breaking Changes
Yes, this deletes the `k8s.ngrok.com/domain` annotation support from the LoadBalancer Service. Its been deprecated for some time now in favor of `k8s.ngrok.com/url`

## Validation

Created a couple of load balancer services

```
# TCP LoadBalancer (no URL annotation = auto-assigned)
apiVersion: v1
kind: Service
metadata:
  name: test-tcp
  namespace: default
  annotations:
    k8s.ngrok.com/mapping-strategy: endpoints-verbose
  labels:
    app.kubernetes.io/name: test-server
spec:
  allocateLoadBalancerNodePorts: false
  loadBalancerClass: ngrok
  ports:
    - name: tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
  selector:
    app.kubernetes.io/name: test-server
  type: LoadBalancer

---
# TLS LoadBalancer (with custom domain)
apiVersion: v1
kind: Service
metadata:
  name: test-tls
  namespace: default
  annotations:
    k8s.ngrok.com/url: tls://test-tcp.alexbezek.me:443
    k8s.ngrok.com/mapping-strategy: endpoints-verbose
  labels:
    app.kubernetes.io/name: test-server
spec:
  allocateLoadBalancerNodePorts: false
  loadBalancerClass: ngrok
  ports:
    - name: tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
  selector:
    app.kubernetes.io/name: test-server
  type: LoadBalancer

```

and confirmed the status is set correctly

```
k get svc -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: Service
  metadata:
    creationTimestamp: "2025-12-05T16:19:59Z"
    labels:
      component: apiserver
      provider: kubernetes
    name: kubernetes
    namespace: default
    resourceVersion: "195"
    uid: 5470f946-dbec-4c47-9a70-aa6b21d09ee5
  spec:
    clusterIP: 10.96.0.1
    clusterIPs:
    - 10.96.0.1
    internalTrafficPolicy: Cluster
    ipFamilies:
    - IPv4
    ipFamilyPolicy: SingleStack
    ports:
    - name: https
      port: 443
      protocol: TCP
      targetPort: 6443
    sessionAffinity: None
    type: ClusterIP
  status:
    loadBalancer: {}
- apiVersion: v1
  kind: Service
  metadata:
    annotations:
      annotation1: val1
      annotation2: val2
    creationTimestamp: "2025-12-11T19:57:07Z"
    labels:
      app.kubernetes.io/managed-by: ngrok-operator
      bindings.k8s.ngrok.com/endpoint-binding-name: ngrok-d30e730c-d44b-5746-a308-a58507613e80
      bindings.k8s.ngrok.com/endpoint-binding-namespace: ngrok-operator
      label1: val1
    name: ngrok-operator-bindings-test-clep
    namespace: default
    resourceVersion: "28387"
    uid: 0a41d823-5845-4d7c-b286-29df81d389dc
  spec:
    externalName: ngrok-d30e730c-d44b-5746-a308-a58507613e80.ngrok-operator.svc.cluster.local
    ports:
    - name: http
      port: 80
      protocol: TCP
      targetPort: 80
    sessionAffinity: ClientIP
    sessionAffinityConfig:
      clientIP:
        timeoutSeconds: 10800
    type: ExternalName
  status:
    loadBalancer: {}
- apiVersion: v1
  kind: Service
  metadata:
    annotations:
      k8s.ngrok.com/computed-url: tcp://5.tcp.ngrok.io:24266
      k8s.ngrok.com/mapping-strategy: endpoints-verbose
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{"k8s.ngrok.com/mapping-strategy":"endpoints-verbose"},"labels":{"app.kubernetes.io/name":"test-server"},"name":"test-tcp","namespace":"default"},"spec":{"allocateLoadBalancerNodePorts":false,"loadBalancerClass":"ngrok","ports":[{"name":"tcp","port":8080,"protocol":"TCP","targetPort":8080}],"selector":{"app.kubernetes.io/name":"test-server"},"type":"LoadBalancer"}}
    creationTimestamp: "2025-12-11T21:17:41Z"
    finalizers:
    - k8s.ngrok.com/finalizer
    labels:
      app.kubernetes.io/name: test-server
    name: test-tcp
    namespace: default
    resourceVersion: "37896"
    uid: 421b56b7-5bb3-434c-9cf1-7af7ac2ba999
  spec:
    allocateLoadBalancerNodePorts: false
    clusterIP: 10.96.224.124
    clusterIPs:
    - 10.96.224.124
    externalTrafficPolicy: Cluster
    internalTrafficPolicy: Cluster
    ipFamilies:
    - IPv4
    ipFamilyPolicy: SingleStack
    loadBalancerClass: ngrok
    ports:
    - name: tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
    selector:
      app.kubernetes.io/name: test-server
    sessionAffinity: None
    type: LoadBalancer
  status:
    loadBalancer:
      ingress:
      - hostname: 5.tcp.ngrok.io
        ports:
        - port: 24266
          protocol: TCP
- apiVersion: v1
  kind: Service
  metadata:
    annotations:
      k8s.ngrok.com/computed-url: tls://test-tcp.alexbezek.me:443
      k8s.ngrok.com/mapping-strategy: endpoints-verbose
      k8s.ngrok.com/url: tls://test-tcp.alexbezek.me:443
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{"k8s.ngrok.com/mapping-strategy":"endpoints-verbose","k8s.ngrok.com/url":"tls://test-tcp.alexbezek.me:443"},"labels":{"app.kubernetes.io/name":"test-server"},"name":"test-tls","namespace":"default"},"spec":{"allocateLoadBalancerNodePorts":false,"loadBalancerClass":"ngrok","ports":[{"name":"tcp","port":8080,"protocol":"TCP","targetPort":8080}],"selector":{"app.kubernetes.io/name":"test-server"},"type":"LoadBalancer"}}
    creationTimestamp: "2025-12-11T21:17:41Z"
    finalizers:
    - k8s.ngrok.com/finalizer
    labels:
      app.kubernetes.io/name: test-server
    name: test-tls
    namespace: default
    resourceVersion: "59839"
    uid: 87674c17-b861-4187-a1b0-cc86eb1416d0
  spec:
    allocateLoadBalancerNodePorts: false
    clusterIP: 10.96.172.116
    clusterIPs:
    - 10.96.172.116
    externalTrafficPolicy: Cluster
    internalTrafficPolicy: Cluster
    ipFamilies:
    - IPv4
    ipFamilyPolicy: SingleStack
    loadBalancerClass: ngrok
    ports:
    - name: tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
    selector:
      app.kubernetes.io/name: test-server
    sessionAffinity: None
    type: LoadBalancer
  status:
    loadBalancer:
      ingress:
      - hostname: 2kapoyyjhr5reymrz.3kg7imcniwfgy1ozt.ngrok-cname.com
        ports:
        - port: 443
          protocol: TCP
kind: List
metadata:
  resourceVersion: ""
```

### DIFF
--- a/tests/chainsaw/loadbalancer-services/chainsaw-test.yaml
+++ b/tests/chainsaw/loadbalancer-services/chainsaw-test.yaml
@@ -45,21 +45,22 @@ spec:
             # Positive port present (null-safe)
             "(length(to_array(loadBalancer.ingress)[? length(to_array(ports)) > `0` && to_number(to_array(ports)[0].port) > `0`]) > `0`)": true
 
-  - name: verify TLS LoadBalancer with custom domain gets CNAME target in status
-    try:
-    - assert:
-        # Custom domain reservation can take longer on first run (needs ngrok API call to reserve)
-        timeout: 60s
-        resource:
-          apiVersion: v1
-          kind: Service
-          metadata:
-            name: test-tls-custom-domain-lb
-          status:
-            # Ingress entry exists with a hostname (null-safe)
-            "(length(to_array(loadBalancer.ingress)[?hostname]) > `0`)": true
-            # For custom domains, hostname MUST be the CNAME target (*.ngrok-cname.com), NOT the custom domain itself
-            # This ensures users can configure their DNS properly
-            "(length(to_array(loadBalancer.ingress)[? hostname && pattern_match('*.ngrok-cname.com', hostname) ]) > `0`)": true
-            # Positive port present (null-safe)
-            "(length(to_array(loadBalancer.ingress)[? length(to_array(ports)) > `0` && to_number(to_array(ports)[0].port) > `0`]) > `0`)": true
+  # Disabled for now as the domain reservation conflicts between users and CI accounts.
+  # - name: verify TLS LoadBalancer with custom domain gets CNAME target in status
+  #   try:
+  #   - assert:
+  #       # Custom domain reservation can take longer on first run (needs ngrok API call to reserve)
+  #       timeout: 60s
+  #       resource:
+  #         apiVersion: v1
+  #         kind: Service
+  #         metadata:
+  #           name: test-tls-custom-domain-lb
+  #         status:
+  #           # Ingress entry exists with a hostname (null-safe)
+  #           "(length(to_array(loadBalancer.ingress)[?hostname]) > `0`)": true
+  #           # For custom domains, hostname MUST be the CNAME target (*.ngrok-cname.com), NOT the custom domain itself
+  #           # This ensures users can configure their DNS properly
+  #           "(length(to_array(loadBalancer.ingress)[? hostname && pattern_match('*.ngrok-cname.com', hostname) ]) > `0`)": true
+  #           # Positive port present (null-safe)
+  #           "(length(to_array(loadBalancer.ingress)[? length(to_array(ports)) > `0` && to_number(to_array(ports)[0].port) > `0`]) > `0`)": true


internal/controller/service/controller_test.go
```diff
@@ -580,10 +579,10 @@ var _ = Describe("ServiceController", func() {
 				})
 			})
 
-			When("service has domain annotation", func() {
```
jonstacks, line 583: I think you can remove this entire block, we should have a test up above for the TLS url annotation if I'm not mistaken.

internal/controller/service/controller.go
```diff
@@ -893,18 +872,19 @@ func updateStatus(ctx context.Context, c client.Client, svc *corev1.Service, end
 		}
 	case !errors.IsMissingAnnotations(err): // Some other error
 		return err
-	default: // computedURL not present, fallback to the domain annotation
-		domain, err := parser.GetStringAnnotation("domain", svc)
+	default: // computedURL not present, fallback to parsing the URL annotation
```
jonstacks: I think there is a tricky case here with removing domain.

* When we were using the `domain` annotation it implied that it was a TLS service. This also implied that the domain should be reserved and that we'd have a domain status to get. The core piece is that we were only looking at domain which would only be for TLS services.
* The reason we introduced the internal `computed-url` annotation is that we had to handle the `tcp://` case, where it means: "Please give me a random TCP url and keep using it". We don't own the spec Service and we don't currently have a `TCPAddr` CRD (maybe we should now), so we needed a place to store the reserved addr.
* Now that domain annotation is removed, if we don't have the `computed-url` annotation (the default case here), it looks like we are still going to try to parse `tcp://` instead and use that to update the status.

What do you think?

alex-bezek: The explanation for how we use computed-url makes sense (i think thats fine for now instead of a dedicated CRD). Since thats a step in the standup of the endpoint, we could just rely on the computed-url if present (seems like it may not be necessary to update the SVC status if we haven't computed and reserved the domain yet). 

I guess in the end, this is primarily just used by external-dns to know the CNAME to make DNS records for which is applicable to TLS but not the auto-reserved TCP addr. I could just remove this extra fallback logic and rely on the computer URL if that sounds better

jonstacks: We still can use external DNS for tcp endpoints. The port isn't able to be customized in that scenario, but you still can connect via `my-custom-domain.com:12345` as an example and some users may like to provide their own certificates for just this case.

Things like ArgoCD also use the load balancer status as a determination on whether or not the LB service is healthy or not, so we should probably make sure it is set even for TCP.

Like you said, I think the most correct thing to do here is to remove the fallback of parsing the `url` and rely on `computed-url`.

internal/controller/service/controller.go
```diff
@@ -865,8 +842,11 @@ func getNgrokTrafficPolicyForService(ctx context.Context, c client.Client, svc *
 	return policy, err
 }
 
-func updateStatus(ctx context.Context, c client.Client, svc *corev1.Service, endpoint ngrokv1alpha1.EndpointWithDomain) error {
+func updateStatus(ctx context.Context, c client.Client, svc *corev1.Service, _ ngrokv1alpha1.EndpointWithDomain) error {
 	clearIngressStatus := func(svc *corev1.Service) error {
+		if len(svc.Status.LoadBalancer.Ingress) == 0 {
+			return nil
+		}
```
jonstacks, line 860: Nice catch!

jonstacks (approved):

internal/controller/service/controller.go
alex-bezek: Oh, i guess i misunderstood things here. I thought in both the tcp and tls case we always created a `computed-url`. But i created some test services locally

```
# TCP LoadBalancer (no URL annotation = auto-assigned)
apiVersion: v1
kind: Service
metadata:
  name: test-tcp
  namespace: default
  annotations:
    k8s.ngrok.com/mapping-strategy: endpoints-verbose
  labels:
    app.kubernetes.io/name: test-server
spec:
  allocateLoadBalancerNodePorts: false
  loadBalancerClass: ngrok
  ports:
    - name: tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
  selector:
    app.kubernetes.io/name: test-server
  type: LoadBalancer

---
# TLS LoadBalancer (with custom domain)
apiVersion: v1
kind: Service
metadata:
  name: test-tls
  namespace: default
  annotations:
    k8s.ngrok.com/url: tls:///test-tcp.alexbezek.me:443
    k8s.ngrok.com/mapping-strategy: endpoints-verbose
  labels:
    app.kubernetes.io/name: test-server
spec:
  allocateLoadBalancerNodePorts: false
  loadBalancerClass: ngrok
  ports:
    - name: tcp
      port: 8080
      protocol: TCP
      targetPort: 8080
  selector:
    app.kubernetes.io/name: test-server
  type: LoadBalancer
```

and i found only the tcp one is getting a computed-url

```
k get svc test-tcp -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    k8s.ngrok.com/computed-url: tcp://5.tcp.ngrok.io:24266
    k8s.ngrok.com/mapping-strategy: endpoints-verbose

k get svc test-tls -o yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    k8s.ngrok.com/mapping-strategy: endpoints-verbose
    k8s.ngrok.com/url: tls:///test-tcp.alexbezek.me:443
```

Because of this code change, it doesn't have a have a computed-url so it doesn't set the status. When testing on main, when i change the annotation back to domain instead of url, that makes the status show. But it seems the url doesn't create a status on main either. 

I think here, our options are:
- go back to what i was doing before and parse the url to get its domain to put in the status
- also create a computed-url for these, however i would be the same value as the url annotation i believe so it seems a little weird, but consistent.

internal/controller/service/controller.go
```diff
@@ -893,30 +881,8 @@ func updateStatus(ctx context.Context, c client.Client, svc *corev1.Service, end
...
-		if dr != nil {
-			// Lookup the domain
-			domain := &ingressv1alpha1.Domain{}
-			if err := c.Get(ctx, client.ObjectKey{Namespace: *dr.Namespace, Name: dr.Name}, domain); err != nil {
-				return err
-			}
-			if domain.Status.CNAMETarget != nil {
-				hostname = *domain.Status.CNAMETarget
-			}
-		}
```
jonstacks, line 919: I broke this for the `url` annotation prior to this PR, but it was working for the `domain` annotation. Leaving the comment here, because I think its related to this.

I've deployed this branch and saw that my custom domains no longer get the `ngrok-cname` in the load balancer status. It just happens to copy the custom domain to the load balancer status and then external-dns tells me that the record is invalid.

tests/chainsaw/loadbalancer-services/test-tls-custom-domain-service.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: test-tls-custom-domain-lb
+  annotations:
+    k8s.ngrok.com/url: tls://service-lb-tls-test-ngrok-operator.example.com:443
```
alex-bezek, line 6: This test fails in the e2e merge queue. I picked a domain i thought would be fairly custom and not used to avoid conflicts with future clients. However, the issue is that i run it locally so my account reserved the domain and now the CI one will fail. I can't just unreserve it as we'll then face the same issue in our local systems. 

Instead i'm going to just comment this test out for now and will investigate how we can ensure these domains are custom either by having chainsaw create a unique namespace for each run and use that in its templating, or by injecting a unique value from the PR number or codespace

jonstacks (approved):

PR #724, opened 2025-12-08
chore(deps): Update go to 1.25.5

jonstacks:
## What

Mainly updates the version of go that we are using in our devshell and that we use to build the docker image. There is 1 caveat with this PR and that the version of go in the devshell is `1.25.4`, while the version we are building and testing with is `1.25.5`. This is because nixpkgs-unstable doesn't have 1.25.5 available yet and targeting staging is more trouble than its worth right now.

## How
Running `nix flake update` and formatting the flake.nix with `nixfmt ./flake.nix`

## Breaking Changes
Shouldn't be, primarily updates the go version from 1.24 -> 1.25.

### DIFF
--- a/.github/actions/build-and-test/action.yaml
+++ b/.github/actions/build-and-test/action.yaml
@@ -36,10 +36,6 @@ runs:
       shell: bash
       run: nix develop --command make build
 
-    - name: Lint
-      shell: bash
-      run: nix develop --command make lint
-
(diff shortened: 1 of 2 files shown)


jonstacks:
Moving to draft while we figure out the other updates needed

alex-bezek (approved):

PR #725, opened 2025-12-09
chore(deps): Use golangci-lint from nixpkgs

jonstacks:
## What

We have several places where we manage dependencies. We have some currently consumed via nix, some via github actions, some via Makefile, and naturally our go dependencies. Having to update our Makefile dependencies periodically as opposed to just running `nix flake update` like we already do for `go` version increases complexity. We're looking at moving many of these to nix and out of go tools where appropriate. 

## How
* Pull golangci-lint from nixpkgs instead of via makefile
* Updates CI to install nix so that `golangci-lint` is on the path for `make lint`
* As part of using the nix provided `golangci-lint`, we're using version 2 of `golangci-lint` now, so this updates the `.golangci-lint` version to match

## Breaking Changes
No, this is an internal build system only change

### DIFF
--- a/docs/developer-guide/README.md
+++ b/docs/developer-guide/README.md
@@ -61,8 +61,12 @@ devenv template so we can address it.
 
 #### Prerequisites
 
+If you are using `nix` with `direnv`, or running with `nix develop`, you can skip installing these tools manually as this is the same setup we
+use in our codespaces devcontainer.
+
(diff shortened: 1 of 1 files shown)


jonstacks:
> Should we also update this contributing guide for the non-nix users to have a list of things you need to install manually https://github.com/ngrok/ngrok-operator/tree/main/docs/developer-guide#non-codespaces

Good call out. Updated the developer documentation

PR #726, opened 2025-12-10
chore(devenv): Move more tools into our nix devenv

jonstacks:
Continues the work from #725 

## What

We have several places where we manage dependencies. We have some currently consumed via nix, some via github actions, some via Makefile, and naturally our go dependencies. Having to update our Makefile dependencies periodically as opposed to just running nix flake update like we already do for go version increases complexity. We're looking at moving many of these to nix and out of go tools where appropriate.

## How
* Updates the `actions/checkout` action from many various version to v6. Also removed specific setup actions for things we have in our nix flake already. As an example, we now run with the same version of chainsaw that runs locally as well as the same version of go(less places to update).
* Removes `bootstrap-tools` make target as everything is now moved to nix.
* Adds caching for our nix flake so that runs are faster
* Cleans up a lot of the indirection in the Makefile
* Helm and the helm-unittest plugin are now consumed via nix
* controller-gen is also consumed from nixpkgs. This caused the version to update to a recent version. (We were quite behind on versions and needed to update some of the kubebuilder markers. This also changed how the role gets generated. It appears the newer version collapsed some of the similar kinds where applicable to make the overall document smaller)
* setup-envtest is now also pulled from nixpkgs.

## Breaking Changes

There shouldn't be any. Most of this is all build related. The one place it could have impact is in the generated rbac since controller-gen gets updated. I looked thought it and it looked right though.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,6 @@ on:
     types: [checks_requested]
 
 env:
-  GO_VERSION: '1.24'
   DOCKER_BUILDX_PLATFORMS: linux/amd64,linux/arm64
 
 jobs:
@@ -27,7 +26,7 @@ jobs:
       tests: ${{ steps.changes.outputs.tests }}
       make: ${{ steps.changes.outputs.make }}
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v6
     - uses: "./.github/actions/changes"
       id: changes
 
@@ -55,7 +54,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v6
     - name: Set up QEMU
       uses: docker/setup-qemu-action@v2
     - id: buildx-setup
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yaml
```diff
@@ -81,11 +78,9 @@ jobs:
...
       (needs.changes.outputs.go == 'true')
     steps:
-    - uses: actions/checkout@v3
-    - name: Set up Go
-      uses: actions/setup-go@v3
-      with:
-        go-version: ${{ env.GO_VERSION }}
+    - uses: actions/checkout@v6
+    - uses: nixbuild/nix-quick-install-action@v34
+    - uses: DeterminateSystems/magic-nix-cache-action@b8276522d77f21bf19d3574e5bc99186a6c7aa6c
```
alex-bezek, line 83: image

scripts/preflight.sh
```diff
@@ -7,6 +7,27 @@ GOVERSION="$(go env GOVERSION || echo "not installed")"
 if ! [[ "$GOVERSION" == "go1.24" ||  "$GOVERSION" = "go1.24."* ]]; then
   echo "Detected go version $GOVERSION, but 1.24 is required"
   exit 1
-else
-  echo "Preflight passed ✅"
 fi
+
+if ! command -v controller-gen >/dev/null 2>&1; then
```
alex-bezek, line 12: should we also check that the kind is available? Or is that intentionally left off incase you want to setup your own cluster?

jonstacks: I left it off since its not technically needed to build the code or generate the Docker image. Like you mentioned, we really just need a k8s cluster of some sort and kind is just one way of getting to that.

PR #729, opened 2025-12-11
Alex/validation tag issues

alex-bezek:
## What
We noticed that some of these kubebuilder annotations were incorrect so this fixes them. 

We also called out in the issue that we should try to lint for these going forward. 

## How
Fixing the actual issue is very simple (just updating the annotation name)

Linting for this has proven more difficult. I did not find any well supported libraries that could lint for these values. I worked with amp and it came up with this soluation of hardcoding valid annotations but i'm not a big fan of them. 

I'm inclined to just not lint for this. 

## Breaking Changes
No


api/bindings/v1alpha1/boundendpoint_types.go
```diff
@@ -114,7 +114,7 @@ type EndpointTarget struct {
 	Port int32 `json:"port"`
 
 	// Metadata is a subset of metav1.ObjectMeta that is added to the Service
-	// +kube:validation:Optional
+	// +kubebuilder:validation:Optional
```
jonstacks, line 117: Even though it was invalid, it looks like this had no effect on the generated output.

scripts/lint-markers.sh
```diff
@@ -0,0 +1,70 @@
...
+    '+kube:default:'         # Should be +kubebuilder:default:
+    '+kube:object:'          # Should be +kubebuilder:object:
+    '+kube:subresource:'     # Should be +kubebuilder:subresource:
+    '+kube:printcolumn:'     # Should be +kubebuilder:printcolumn:
+    '+kube:resource:'        # Should be +kubebuilder:resource:
+    '+kube:rbac:'            # Should be +kubebuilder:rbac:
+    '+kube:webhook:'         # Should be +kubebuilder:webhook:
+    '+kube:skip'             # Should be +kubebuilder:skip
+    '+kube:storageversion'   # Should be +kubebuilder:storageversion
+)
```
jonstacks, line 37: This looks like it will only catch `+kube` making the marker invalid. It looks like we aren't handling the case of `+kuebuilder:object` or any other misspellings. Still feel like an improvement, just noting where it might bite us.

I think long term if we want something more robust, maybe we add a linter to golangci-lint for this? I'm sure others might be interested too.

tools/make/lint.mk
```diff
@@ -5,6 +5,11 @@ lint: ## Run golangci-lint linter & yamllint
 	golangci-lint run
 
 
+.PHONY: lint-markers
+lint-markers: ## Check for invalid kubebuilder marker prefixes
```
jonstacks, line 9: Should we add this as a dependency for the `lint` target so that it gets run in CI automatically?

PR #732, opened 2025-12-16
feat: Split ngrok-operators CRDs into their own chart

jonstacks:
## What

Split the ngrok Operator's CRDs into their own helm chart.

## Why

Splitting them into their own chart allows users to manage the lifecycle of the CRDs separate from the operator. This should have several benefits:
* With a few future changes it will allow for installing multiple versions of the ngrok-operator
* Installing the CRDs on their own and then helm uninstalling the ngrok-operator and re-installing it should not lead to missing CRDs. (Related to #597)

## How
* Update the make target to generate the ngrok CRDs into the ngrok-crds helm chart.
* Add a dependency on the ngrok-crds chart to the ngrok-operator.
* Update the GitHub actions to release and publish the ngrok-crds chart.

## Breaking Changes
No, the `installCRDs` flag defaults to true.


helm/ngrok-crds/README.md
```diff
@@ -0,0 +1,36 @@
...
+`helm delete ngrok-crds`
+
+### Using with ngrok-operator
+
+The `ngrok-operator` chart can install these CRDs automatically via the `installCRDs` value (enabled by default).
+You only need to install this chart separately if you want to manage CRD lifecycle independently from the operator.
+
+## More Information
+
+For more information about the ngrok Kubernetes Operator, see https://github.com/ngrok/ngrok-operator
```
alex-bezek, line 36: Should we queue up a docs site change to talk about this chart and link to that in the future?

PR #734, opened 2025-12-16
Fix invalid kubebuilder code gen markers

alex-bezek:
## What
We noticed some build markers were not using the right names/syntax. This fixes them

## How
We actually created a linter in this other PR to find these https://github.com/ngrok/ngrok-operator/pull/731
While we do want to add linting, the other PR will likely be closed in favor of making an open source plugin. 
This PR just fixes the issues quickly for now

## Breaking Changes
- Yes, we actually do validation of types now that we have the right syntax
  - agent and cloud endpoints correctly now only allow internal/public/k8s binding types
  - kubernetes operator restricts registration status to its enum types now (not really non-passive i guess)


api/bindings/v1alpha1/boundendpoint_types.go
```diff
@@ -114,7 +114,7 @@ type EndpointTarget struct {
 	Port int32 `json:"port"`
 
 	// Metadata is a subset of metav1.ObjectMeta that is added to the Service
-	// +kube:validation:Optional
+	// +kubebuilder:validation:Optional
```
alex-bezek: this resulted in no changes to the manifest because its redundant with the `omitempty`. We have 33 of these through our types but we aren't consistently using it. I could also just remove this marker from all the types if that sounds good

alex-bezek: https://github.com/ngrok/ngrok-operator/pull/735

jonstacks (approved):

api/ngrok/v1alpha1/cloudendpoint_types.go
```diff
@@ -91,12 +86,11 @@ type CloudEndpointStatus struct {
 	// DomainRef is a reference to the Domain resource associated with this endpoint.
 	// For internal endpoints, this will be nil.
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:Nullable
+	// +nullable
```
alex-bezek, line 89: domainRef is the unique one here where it doesn't have omitEmpty. This is because i explicitly needed to clear it from the status by setting it to null https://github.com/ngrok/ngrok-operator/blob/main/internal/domain/manager.go#L96

Fixing this nullable marker does update the CRD to include `nullable: true`. This should be passive as k8s already defaults to letting objects be null. This doesn't change the go struct marshalling so I believe its passive and redundant but is a bit more explicit for this situation.

jonstacks (approved):

PR #736, opened 2025-12-18
feat(helm): Allow multiple versions of the operator to be installed

jonstacks:
## What

We've broken out the CRDs into their own chart. With that work, this allows multiple versions of the ngrok-operator to be installed in the same cluster by first installing the CRDs and then installing the ngrok-operator chart twice with `installCRDs: false` for both the the helm installs.

## How

A few of the RBAC resources did not include the release name. This fixes it so that the `Role`s, `RoleBinding`s, `ClusterRole`s, and `ClusterRoleBinding`s have the release name included.

I've tested this by using the new make command: `make deploy_multi_namespace`. The eventual goal(not in this PR) is to have this path tested in CI as well.

## Breaking Changes
No.

### DIFF
--- a/tools/make/deploy.mk
+++ b/tools/make/deploy.mk
@@ -63,7 +63,7 @@ deploy_with_bindings: _deploy-check-env-vars docker-build manifests _helm_setup 
 		$(HELM_DESCRIPTION_FLAG)
 
 .PHONY: deploy_for_e2e
-deploy_for_e2e: _deploy-check-env-vars docker-build manifests _helm_setup kind-load-image ## Deploy controller to the K8s cluster specified in ~/.kube.config.
+deploy_for_e2e: _deploy-check-env-vars docker-build manifests _helm_setup kind-load-image ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	helm upgrade $(HELM_RELEASE_NAME) $(HELM_CHART_DIR) --install \
 		--namespace $(KUBE_NAMESPACE) \
 		--create-namespace \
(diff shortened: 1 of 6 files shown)


tools/make/deploy.mk
```diff
@@ -85,7 +85,52 @@ deploy_for_e2e: _deploy-check-env-vars docker-build manifests _helm_setup kind-l
...
+## 1. We want to install the CRDs only once at the beginning
+## 2. We want to make a namespace-a and namespace-b
+## 3. We want to install the helm chart twice, watching only the namespace belonging to each one
 deploy_multi_namespace: _deploy-check-env-vars docker-build manifests _helm_setup kind-load-image ## Deploy multiple copies of the controller to the K8s cluster specified in ~/.kube/config.
+	helm upgrade ngrok-operator-crds $(CRD_CHART_DIR) --install \
+		--kube-context=kind-$(KIND_CLUSTER_NAME) \
+	    --namespace kube-system \
+		--create-namespace
+	kubectl --context=kind-$(KIND_CLUSTER_NAME) create namespace namespace-a || true
+	kubectl --context=kind-$(KIND_CLUSTER_NAME) create namespace namespace-b || true
```
alex-bezek: Nit: You could use `--create-namespace` instead of this similar to how the other make targets work

tools/make/deploy.mk
```diff
@@ -85,7 +85,52 @@ deploy_for_e2e: _deploy-check-env-vars docker-build manifests _helm_setup kind-l
 		--set bindings.serviceLabels.label1="val1"
 
 .PHONY: deploy_multi_namespace
+## 1. We want to install the CRDs only once at the beginning
+## 2. We want to make a namespace-a and namespace-b
+## 3. We want to install the helm chart twice, watching only the namespace belonging to each one
 deploy_multi_namespace: _deploy-check-env-vars docker-build manifests _helm_setup kind-load-image ## Deploy multiple copies of the controller to the K8s cluster specified in ~/.kube/config.
+	helm upgrade ngrok-operator-crds $(CRD_CHART_DIR) --install \
+		--kube-context=kind-$(KIND_CLUSTER_NAME) \
+	    --namespace kube-system \
```
alex-bezek, line 94: Out of curiosity, is this even necessary? All the resources should be cluster scoped right?

jonstacks: Yes, this is just the namespace where the `helm` release info gets stored. For multi-install, I figured putting it in `kube-system` was good so we can test deleting `namespace-a` or `namespace-b`. Since the CRDs extend the kubernetes API, it felt natural to me to put the `helm` release in `kube-system`.

The other option I see would be bypassing helm and just doing a `kubectl apply -f` of the ngrok-crds templates directory. I was leaning the helm direction so that we test the chart install works like our users would. Let me know what you think

tools/make/deploy.mk
jonstacks: Fixed

tools/make/deploy.mk
jonstacks: I'm going to merge since this was just a nit. Let me know what you think and its something we can change pretty easily

alex-bezek: ahh, i wasn't considering the helm release secret. that makes sense, and I don't think it matters much for the secret so this is good 👍

PR #737, opened 2025-12-19
test: Add e2e tests for multiple installs in the same cluster

jonstacks:
## What

Adds CI tests that we can build off over time for testing installing multiple ngrok operators into a single cluster. Now that we have multiple installs working, we want to run tests to make sure we don't break that functionality.

## How
Adds a new `e2e-multi-namespace` that runs after the regular `e2e` tests.

## Breaking Changes
No, this is testing only.

### DIFF
--- a/tests/chainsaw-multi-ns/ingresses/chainsaw-test.yaml
+++ b/tests/chainsaw-multi-ns/ingresses/chainsaw-test.yaml
@@ -1,0 +1,48 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: multi-ns-ingresses
+spec:
+  steps:
+  - name: create resources in namespace-a
+    try:
+    - create: { file: ./namespace-a/deployment.yaml }
+    - create: { file: ./namespace-a/service.yaml }
+    - create: { file: ./namespace-a/ingress.yaml }
+
+  - name: create resources in namespace-b
+    try:
+    - create: { file: ./namespace-b/deployment.yaml }
+    - create: { file: ./namespace-b/service.yaml }
+    - create: { file: ./namespace-b/ingress.yaml }
+
+  - name: verify ingress in namespace-a gets created
+    try:
+    - assert:
+        resource:
+          apiVersion: networking.k8s.io/v1
+          kind: Ingress
+          metadata:
+            name: ingress-a
+            namespace: namespace-a
+          status:
+            # At least one ingress object with a hostname (null-safe; no [0] until it exists)
+            "(length(to_array(loadBalancer.ingress)[?hostname]) > `0`)": true
+            # At least one hostname matches "test-ngrok-operator-ingress-a.ngrok.app"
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -182,3 +182,50 @@ jobs:
...
+      run: nix develop --command make e2e-tests-multi-ns
+    - name: Cleanup
+      if: always()
+      run: |
+        echo "Deleting kubernetesoperators..."
+        kubectl -n namespace-a delete kubernetesoperator ngrok-operator-a || true
+        kubectl -n namespace-b delete kubernetesoperator ngrok-operator-b || true
+        echo "Uninstalling the operators..."
+        helm uninstall ngrok-operator-a -n namespace-a || true
+        helm uninstall ngrok-operator-b -n namespace-b || true
```
alex-bezek, line 231: would it be worth removing the `|| true` so it actually fails if they error? Or if we don't care about ensuring these uninstalls work and the goal is to just reset kind we can just delete the kind cluster

tests/chainsaw-multi-ns/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+These tests assume the cluster has been set up using `make deploy_multi_namespace`, which deploys:
+
+1. **CRDs** installed once in `kube-system` via `ngrok-operator-crds` helm chart
+2. **Operator A** (`ngrok-operator-a`) in `namespace-a`:
+   - Watches only `namespace-a`
+   - Uses ingress class `ngrok-a`
+   - Controller name: `k8s.ngrok.com/ingress-controller-a`
+3. **Operator B** (`ngrok-operator-b`) in `namespace-b`:
+   - Watches only `namespace-b`
```
alex-bezek, line 15: It might be simple to write a chainsaw test for these by making a set of ingress objects in a combination of namespaces + ingressClasses and asserting the cloud endpoint CRDs are created or checking the ingress' statuses

alex-bezek (approved):

.github/workflows/ci.yaml
jonstacks: The goal here wasn't to reset the kind cluster since this is run in a GitHub action and the runner disappears at the end of the job. This step specifies `if: always()` as a cleanup mechanism primarily for ngrok resources.

Now that we are installing 2 operators, it seemed like there were 4 different scenarios:
1. None of the operators install correctly.
2. `namespace-a` succeeds, `namespace-b` fails
3. `namespace-b` succeeds, `namespace-a` fails
4. both install correctly

The intent is to de-register any successfully registered operators and (try to) gracefully helm uninstall to handle cases 2, 3, 4 without writing complex logic to check for which state we are in.

tests/chainsaw-multi-ns/README.md
jonstacks: This seems reasonable. I'll update it to add the tests

.github/workflows/ci.yaml
alex-bezek: ahh ok, i wasn't considering the possibility of one of the installs failing and the need to clean up the ngrok resources. Thanks for the breakdown that makes sense.

tests/chainsaw-multi-ns/README.md
jonstacks: Done in 5e152dc

Issue #88, opened 2022-12-28
Allow adding annotations to helm cluster roles


alex-bezek:
### Description

We should create a helm value that can be used to pass in arbitrary annotations that get added to the cluster role. 

### Use Case

With a CRD, a user/serviceaccount needs permission to access a specific resource. Out of the box, we can access a crd easily because auth is defaulted to the `system:masters` role. It is a best practice to not use this role and to create more locked down roles. 

Allowing annotations to be added to the cluster role would allow users to add annotations like these to easily give themselves access to the CRDs

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: aggregate-cron-tabs-edit
  annotations:
    # Add these permissions to the "admin" and "edit" default roles.
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
```

PR #738, opened 2025-12-22
Allow adding annotations to helm cluster roles

sabrina-ngrok:
Issue: #88
## What
Create a helm value that can be used to pass in arbitrary annotations that get added to the cluster role.

## How
- `values.yaml`- Add definition for cluster role and adding the annotation
- `role.yaml rbac.yaml controller-rbac.yaml` - Update main ClusterRole templates with annotations
- `editor/viewer subagents` - Update editor/viewer ClusterRole templates with annotations
- `controller-rbac_test.yaml rbac_test.yaml` - Add unit test for adding annotation to cluster role and confirming it was set
 
Unit tests testing
```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: aggregate-cron-tabs-edit
  annotations:
    # Add these permissions to the "admin" and "edit" default roles.
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
    rbac.authorization.k8s.io/aggregate-to-edit: "true"
```

- Generate new version of `README` and snapshots

## Breaking Changes
¯\\(ツ)/¯

### DIFF
--- a/helm/ngrok-operator/values.yaml
+++ b/helm/ngrok-operator/values.yaml
@@ -205,7 +205,8 @@ serviceAccount:
   annotations: {}
 
 ## Cluster Role Settings
-## @param clusterRole.annotations Additional annotations to add to all ClusterRoles
+## @param clusterRole.annotations Additional annotations to add to all ClusterRoles.
+## Commonly used for RBAC role aggregation (e.g., aggregate this ClusterRole into the built-in `admin`/`edit` roles) to grant CRD access without relying on `system:masters`.
 ##
 clusterRole:
   annotations: {}
--- a/helm/ngrok-operator/README.md
+++ b/helm/ngrok-operator/README.md
@@ -121,7 +121,7 @@ To run multiple ngrok-operator instances in the same cluster (e.g., in different
 | `serviceAccount.create`              | Specifies whether a ServiceAccount should be created                                                                                           | `true`   |
 | `serviceAccount.name`                | The name of the ServiceAccount to use.                                                                                                         | `""`     |
 | `serviceAccount.annotations`         | Additional annotations to add to the ServiceAccount                                                                                            | `{}`     |
-| `clusterRole.annotations`            | Additional annotations to add to all ClusterRoles                                                                                              | `{}`     |
+| `clusterRole.annotations`            | Additional annotations to add to all ClusterRoles.                                                                                             | `{}`     |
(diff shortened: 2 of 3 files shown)


helm/ngrok-operator/values.yaml
```diff
@@ -204,6 +204,12 @@ serviceAccount:
   name: ""
   annotations: {}
 
+## Cluster Role Settings
+## @param clusterRole.annotations Additional annotations to add to all ClusterRoles
+##
+clusterRole:
```
alex-bezek, line 211: Since this value gets applied to all the cluster roles, what are your thoughts on making this plural? 
I think a single value to apply to all of them covers 90% of the use cases and is all we are looking to handle right now. But i'm trying to consider also if one day someone wanted to annotate the CRD's separately from the operator's roles, or have specific annotations for 1 CRD but not another (maybe users can make cloud endpoints, but can't delete domains for example). I think we could have this single value apply to all of them and in the future allow specifying `.clusterRoles[crdName].annotations` passively

helm/ngrok-operator/values.yaml
```diff
@@ -204,6 +204,12 @@ serviceAccount:
   name: ""
   annotations: {}
 
+## Cluster Role Settings
+## @param clusterRole.annotations Additional annotations to add to all ClusterRoles
```
alex-bezek: It might be worth putting a little explanation here on what you would use these annotations for. k8s has some decent docs here we could link to:
- https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles
- https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

This might be worth working with @iambmelt on some doc updates as well as a follow up. I could see us having a doc page on hardening the security of the operator/cluster and documenting how to restrict CRD access there.

helm/ngrok-operator/values.yaml
sabrina-ngrok: We had a conversation about this option and due to K8 documentation being singular `clusterRole`, we are going to go ahead with that option currently and then possibly add a plural version in the future if needed. https://chatgpt.com/share/e/695404c1-7594-8004-9296-15cd82998385

helm/ngrok-operator/values.yaml
sabrina-ngrok:
```suggestion
## @param clusterRole.annotations Additional annotations to add to all ClusterRoles.
## Commonly used for RBAC role aggregation (e.g., aggregate this ClusterRole into the built-in `admin`/`edit` roles) to grant CRD access without relying on `system:masters`.
```

PR #743, opened 2026-01-14
Updating proto to match ngrok-private, adding PodIdentity

sabrina-ngrok:
#K8SOP-202

## What
Project: Traffic Policy - Add `conn.client_id.k8s` variables
Milestone 2: Extend Proto

- Updating proto to match `ngrok-private` which adds PodIdentity: [link](https://github.com/ngrok-private/ngrok/blob/main/go/lib/pb_agent/conn_header.pb.go)

## How
- Copy and paste file from `ngrok-private`
- Updated use of `EndpointID` to `EndpointId` because that was changed in proto

### DIFF
--- a/internal/pb_agent/conn_header.pb.go
+++ b/internal/pb_agent/conn_header.pb.go
@@ -3,11 +3,11 @@
 package pb_agent
 
 import (
-	reflect "reflect"
-	sync "sync"
-
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
+	reflect "reflect"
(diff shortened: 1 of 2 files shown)


internal/pb_agent/conn_header.pb.go
```diff
@@ -3,11 +3,11 @@
 package pb_agent
```
alex-bezek, line 3: this is all just copy pasted from our internal protobuff definition i assume?

sabrina-ngrok: Yes I left the top TODO part and just copied everything under it from the newly generated proto go file from `ngrok-private`.

PR #744, opened 2026-01-15
Don't create internal domains for ingress and gateway objects

alex-bezek:
## What

For cloud endpoints and agent endpoints we created code that skips creating the domain CRD for internal domains as those don't need to be reserved and can't be

```
43m         Warning   CreateError         domain/webhooks-stripe-internal                                    Failed to Create Domain/webhooks-stripe-internal: HTTP 400: You may not reserve a name on the TLD '.internal'. This TLD is not reachable over the public internet. Failed to reserve 'webhooks-stripe.internal'. [ERR_NGROK_411]...
```

https://github.com/ngrok/ngrok-operator/blob/main/internal/domain/manager.go#L105

However, if the ingress object itself has the internal domain, we don't skip those

https://github.com/ngrok/ngrok-operator/blob/main/pkg/managerdriver/domains.go#L16

Deleting our internal domains in our clusters and then restarting the operator recreated all of them. 

## How
*Describe the solution*
- make a re-usable function to check if the domain is `internal`
- put this helper in utils. Unfortunately, there is a circular dependency if i put it in the `internal/domain` package between the api manager driver, the ingress package, and `internal/domain`. The issue is the ingress package imports the managerDriver package in order to call its `NewControllerEventhandler` function. the `internal/domain` package imports ingress to access its `IsDomainReady` function which uses conditions defined in the ingress package. If `managerDriver` also tries to import the `internal/domain` it makes a circular dependency
```
package github.com/ngrok/ngrok-operator
        imports github.com/ngrok/ngrok-operator/cmd from main.go
        imports github.com/ngrok/ngrok-operator/internal/controller/agent from agent-manager.go
        imports github.com/ngrok/ngrok-operator/internal/domain from agent_endpoint_conditions.go
        imports github.com/ngrok/ngrok-operator/internal/controller/ingress from manager.go
        imports github.com/ngrok/ngrok-operator/pkg/managerdriver from ingress_controller.go
        imports github.com/ngrok/ngrok-operator/internal/domain from domains.go: import cycle not allowed
```
I began refactoring this but it would mean moving all the conditions, which then i thought back to how we have discussed moving conditions to the api package instead after we refactor them. So i stopped with the refactoring for now. 
- also had the domain controller delete any `.internal` domain CRs it sees. These will just requeue indefinitely trying to get reserved which is what we see internally right now
- added various chainsaw and env tests. 
  - Didn't test gateway via chainsaw explicitly as we don't have the gateway stuff setup anywhere. 

## Breaking Changes

### DIFF
--- a/internal/controller/ingress/domain_controller.go
+++ b/internal/controller/ingress/domain_controller.go
@@ -127,12 +127,10 @@ func (r *DomainReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	// Delete internal domains - they cannot be reserved via ngrok API
 	if util.IsInternalDomain(domain.Spec.Domain) {
-		log.Info("Deleting Domain CRD with internal TLD - internal domains cannot be reserved via ngrok API",
-			"hostname", domain.Spec.Domain)
-		if err := r.Delete(ctx, domain); err != nil {
-			return ctrl.Result{}, client.IgnoreNotFound(err)
-		}
-		return ctrl.Result{}, nil
+		log.Info("Deleting Domain CRD with internal TLD - internal domains cannot be reserved via ngrok API", "hostname", domain.Spec.Domain)
+
+		err := r.Delete(ctx, domain)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
 	result, err := r.controller.Reconcile(ctx, req, new(v1alpha1.Domain))


internal/util/endpoints.go
```diff
@@ -8,6 +8,14 @@ import (
 	"strings"
 )
 
+// IsInternalDomain returns true if the given hostname ends with ".internal" TLD.
+// Internal domains cannot be reserved via the ngrok API (returns HTTP 400).
+// This function handles case-insensitivity and trailing dots.
+func IsInternalDomain(host string) bool {
```
jonstacks, line 14: Nice addition. Good idea to also have it work with a trailing dot.

internal/controller/ingress/domain_controller.go
```diff
@@ -116,14 +117,31 @@ func (r *DomainReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+
+	// Delete internal domains - they cannot be reserved via ngrok API
+	if util.IsInternalDomain(domain.Spec.Domain) {
+		log.Info("Deleting Domain CRD with internal TLD - internal domains cannot be reserved via ngrok API",
+			"hostname", domain.Spec.Domain)
+		if err := r.Delete(ctx, domain); err != nil {
+			return ctrl.Result{}, client.IgnoreNotFound(err)
+		}
+		return ctrl.Result{}, nil
```
jonstacks: *Nit*: What do you think about

```go
return ctrl.Result{}, client.IgnoreNotFound(r.Delete(ctx, domain))
```
or 
```go
err := r.Delete(ctx, domain)
return ctrl.Result{}, client.IgnoreNotFound(err)
```

PR #746, opened 2026-01-24
feat: support Domain `resolves_to` field

andrew-harris-at-ngrok:
## What

In https://github.com/ngrok/ngrok-api-go/releases/tag/v7.8.0 , the ngrok API now supports the `resolves_to` field on Reserved Domains.

We should support the field on the Operator as well.

## How

- Bumps `ngrok-api-go` to `v7.8.0`
- Adds support in code for parsing `resolves_to`.

### and also...

To get to `ngrok-api-go` `v7.8.0`, we also have to support changes that were made in `v7.7.0` around Cloud Endpoint `PoolingEnabled` now being an optional field. So, this PR also:

- Adds support in code for Cloud Endpoint `PoolingEnabled` as an optional field.

I'd be happy to split this out as a separate PR, as the changes are non-trivial.

## Breaking Changes

While Cloud Endpoint `PoolingEnabled` is now optional _in the API_, the signature of the relevant objects in this PR change to accommodate that shift. I don't _believe_ that leads to a breaking change for clients, as passing through `true` or `false` in YAML or other infra-as-code should work the same as before. However, I'd like someone to please check my assertion on that.


api/ingress/v1alpha1/domain_types.go
```diff
@@ -53,6 +53,9 @@ type DomainSpec struct {
 	// Region is the region in which to reserve the domain
 	Region string `json:"region,omitempty"`
 
+	// ResolvesTo is the list of resolving targets for the domain
+	ResolvesTo *[]DomainResolvesToEntry `json:"resolves_to,omitempty"`
```
jonstacks, line 57: Probably not worth limiting here since this will change as our infra changes. I was just curious on if we limit at all in the API

internal/annotations/annotations_test.go
```diff
@@ -82,40 +82,42 @@ func TestExtractNgrokTrafficPolicyFromAnnotations(t *testing.T) {
 }
 
 func TestExtractUseEndpointPooling(t *testing.T) {
+	boolPtr := func(b bool) *bool { return &b }
```
jonstacks: We're already using `ptr.To` in a lot of places. I would recommend that over a new `boolPtr` function.

internal/controller/ingress/domain_controller.go
```diff
@@ -175,9 +176,11 @@ func (r *DomainReconciler) update(ctx context.Context, domain *v1alpha1.Domain)
...
 	}
 
-	// Only update the domain if the description or metadata has changed
-	// These are the only fields that can be updated that we write to.
-	if domain.Spec.Description == resp.Description && domain.Spec.Metadata == resp.Metadata {
+	// Only update the domain if updatable fields have changed
+	specResolvesTo := buildResolvesToRequest(domain.Spec.ResolvesTo)
+	if domain.Spec.Description == resp.Description &&
+		domain.Spec.Metadata == resp.Metadata &&
+		resolvesToEqual(specResolvesTo, resp.ResolvesTo) {
```
jonstacks:
```suggestion
        reflect.DeepEqual(specResolvesTo, resp.ResolvesTo) {
```

pkg/managerdriver/translate_ingresses.go
```diff
@@ -14,6 +14,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// boolPtrEqual compares two *bool values for equality.
+// Returns true if both are nil, or both are non-nil with the same value.
+func boolPtrEqual(a, b *bool) bool {
```
jonstacks: Could replace this with https://pkg.go.dev/k8s.io/utils/ptr#Equal

alex-bezek: I believe we discussed yesterday using https://pkg.go.dev/k8s.io/utils/ptr#Equal since we already use `ptr.To` from this package and the implementation is pretty similar https://github.com/kubernetes/utils/blob/914a6e7505707ae6d13abe19730c24b4cfde9e6f/ptr/ptr.go#L65-L73

api/ingress/v1alpha1/domain_types.go
andrew-harris-at-ngrok: We do, and the limit is based on the pool of available valid resolving targets. In the current state of this feature, that's our [Points of Presence](https://ngrok.com/docs/universal-gateway/points-of-presence).

pkg/managerdriver/translate_ingresses.go
andrew-harris-at-ngrok: https://github.com/ngrok/ngrok-operator/pull/746/commits/d3fdc20ae70c07f54bd9ff8c539bbe05d2a07191

andrew-harris-at-ngrok: https://github.com/ngrok/ngrok-operator/pull/746/commits/d3fdc20ae70c07f54bd9ff8c539bbe05d2a07191

internal/controller/ingress/domain_controller.go
andrew-harris-at-ngrok: https://github.com/ngrok/ngrok-operator/pull/746/commits/d3fdc20ae70c07f54bd9ff8c539bbe05d2a07191

internal/annotations/annotations_test.go
andrew-harris-at-ngrok: https://github.com/ngrok/ngrok-operator/pull/746/commits/d3fdc20ae70c07f54bd9ff8c539bbe05d2a07191

PR #750, opened 2026-02-02
Allow helm uninstall to be configured to handle cleaning up api resou…

alex-bezek:
…rces and removing finalizers

## What

There have been multiple pain points for people when uninstalling the operator and getting left with resources like internal CRDs that are stuck in deletion due to finalizers. These resources getting stuck then results in the CRDs themselves from not being deleted and it blocks re-installing the operator. 

## How

This sets up a helm pre-delete hook that initates a "Drain" of the resources when helm uninstall is used. 
By default, this drain process tries to preserve ngrok-api resources where it can such as cloud endpoints aiming at supporting the product infrastructure use case to avoid tearing down infra when possible. 
its also configurable and can be set to `Delete` if you want to iterate locally on uninstalling multiple times and want it to delete resources from the ngrok account. 

This also works slightly differently depending on how the CRDs were installed. When the delete policy is used, it always tries to delete the CRs so they get removed from the ngrok-api. However if retain is used, the CRs will be preserved if the CRDs were installed separately and aren't getting uninstalled. But if the CRDs are bundled with the operator install, they'll be removed at the end. 

This also works in some multi-install scenarios. Specifically it works whenever you are installing the operator multiple times and having them watch different namespaces. It DOES NOT work when installing the operator multiple times to watch the same namespaces but different ingress classes as right now there is no way for the 2 operators to distinguish between the 2 different types of CRs created by the operator. 

This also made some changes to the docker image build. I found it annoying to have to rebuild everytime i edited the makefiles or the scripts folder while i was iterating on tests. So i broke that process up a bit. I may still remove this from this PR and do it in its own changeset, but i left it for now incase i need to itarate more

## Breaking Changes

Shouldn't be. Uninstalling behavior is different is the major call out. 

## Docs

Being created here https://github.com/ngrok/ngrok-docs/pull/1896/changes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,24 @@
 # Build the manager binary
 FROM --platform=$BUILDPLATFORM golang:1.25.5 AS builder
 
-ARG TARGETOS TARGETARCH
-ARG GIT_COMMIT=""
-
 WORKDIR /workspace
-
-# Copy the Go Modules manifests first for better layer caching
-COPY go.mod go.sum ./
+# Copy the Go Modules manifests
+COPY go.mod go.mod
+COPY go.sum go.sum
+# cache deps before building and copying source so that we don't need to re-download as much
+# and so that source changes don't invalidate our downloaded layer
 RUN --mount=type=cache,target=/go \
 	go mod download
 
-# Copy the rest
+# Copy the go source
 COPY . .
+
+ARG TARGETOS TARGETARCH
+ARG GIT_COMMIT=""
 
 # Build
 RUN --mount=type=cache,target=/go \
-	CGO_ENABLED=0 GOOS="${TARGETOS}" GOARCH="${TARGETARCH}" GIT_COMMIT="${GIT_COMMIT}" ./scripts/build.sh
+        CGO_ENABLED=0 GOOS="${TARGETOS}" GOARCH="${TARGETARCH}" GIT_COMMIT="${GIT_COMMIT}" ./scripts/build.sh
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
--- a/api/ngrok/v1alpha1/kubernetesoperator_types.go
+++ b/api/ngrok/v1alpha1/kubernetesoperator_types.go
@@ -83,8 +83,7 @@ type KubernetesOperatorStatus struct {
 	BindingsIngressEndpoint string `json:"bindingsIngressEndpoint,omitempty"`
 
 	// DrainStatus indicates the current state of the drain process
-	// +kubebuilder:validation:Enum=pending;draining;completed;failed
-	DrainStatus string `json:"drainStatus,omitempty"`
+	DrainStatus DrainStatus `json:"drainStatus,omitempty"`
 
 	// DrainMessage provides additional information about the drain status
 	DrainMessage string `json:"drainMessage,omitempty"`
@@ -110,11 +109,15 @@ const (
 	KubernetesOperatorFeatureBindings = "bindings"
 )
 
+// DrainStatus indicates the current state of the drain process
+// +kubebuilder:validation:Enum=pending;draining;completed;failed
+type DrainStatus string
+
 const (
-	DrainStatusPending   = "pending"
-	DrainStatusDraining  = "draining"
-	DrainStatusCompleted = "completed"
-	DrainStatusFailed    = "failed"
+	DrainStatusPending   DrainStatus = "pending"
+	DrainStatusDraining  DrainStatus = "draining"
+	DrainStatusCompleted DrainStatus = "completed"
+	DrainStatusFailed    DrainStatus = "failed"
 )
 
 // DrainPolicy determines how ngrok API resources are handled during drain
--- a/cmd/api-manager.go
+++ b/cmd/api-manager.go
@@ -785,13 +785,18 @@ func createKubernetesOperator(ctx context.Context, client client.Client, opts ap
 		},
 	}
 	_, err := controllerutil.CreateOrUpdate(ctx, client, k8sOperator, func() error {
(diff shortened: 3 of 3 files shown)


helm/ngrok-operator/values.yaml
```diff
@@ -428,3 +428,34 @@ bindings:
...
+## @param cleanupHook.timeout Timeout in seconds for the cleanup process
+## @param cleanupHook.image.repository The repository for the kubectl image used by the cleanup hook
+## @param cleanupHook.image.tag The tag for the kubectl image
+## @param cleanupHook.image.pullPolicy The pull policy for the cleanup hook image
+## @param cleanupHook.resources.limits.cpu CPU limit for the cleanup hook container
+## @param cleanupHook.resources.limits.memory Memory limit for the cleanup hook container
+## @param cleanupHook.resources.requests.cpu CPU request for the cleanup hook container
+## @param cleanupHook.resources.requests.memory Memory request for the cleanup hook container
+##
+drainPolicy: "Retain"  # "Delete" or "Retain"
```
alex-bezek, line 446: I considered nesting this under the cleanup hook, however this drain behavior happens regardles of if you are using the cleanup hook if you were to just delete your k8s operator CR

api/ngrok/v1alpha1/kubernetesoperator_types.go
```diff
@@ -95,6 +110,24 @@ const (
 	KubernetesOperatorFeatureBindings = "bindings"
 )
 
+const (
+	DrainStatusPending   = "pending"
+	DrainStatusDraining  = "draining"
+	DrainStatusCompleted = "completed"
+	DrainStatusFailed    = "failed"
```
jonstacks: What are your thoughts on making a

```go
type DrainStatus string
```

, similar to `DrainPolicy`, to make the options clear in the go code?.

cmd/api-manager.go
```diff
@@ -745,15 +785,13 @@ func createKubernetesOperator(ctx context.Context, client client.Client, opts ap
...
-				Version:   version.GetVersion(),
-			},
-			Region: opts.region,
+		k8sOperator.Spec.Description = opts.description
+		k8sOperator.Spec.Deployment = &ngrokv1alpha1.KubernetesOperatorDeployment{
+			Name:      opts.releaseName,
+			Namespace: opts.namespace,
+			Version:   version.GetVersion(),
 		}
+		k8sOperator.Spec.Region = opts.region
```
jonstacks: Was this a re-assignment issue where we were overwriting the entire spec?

internal/controller/ingress/ingress_controller.go
```diff
@@ -126,6 +125,29 @@ func (r *IngressReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	// Skip non-delete reconciles during drain to prevent adding new finalizers
+	if controller.IsDraining(ctx, r.DrainState) {
+		log.V(1).Info("Draining, skipping non-delete reconcile")
+		// Remove from store so no new resources are created for this ingress
+		if err := r.Driver.DeleteIngress(ingress); err != nil {
+			log.Error(err, "Failed to delete ingress from store during drain")
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
```
jonstacks, line 145: Might be reading this wrong, but if we aren't doing a delete and we are draining, do we need to remove the finalizer if it is present?

Dockerfile
```diff
@@ -1,23 +1,22 @@
...
+# Copy the Go Modules manifests first for better layer caching
+COPY go.mod go.sum ./
 RUN --mount=type=cache,target=/go \
 	go mod download
 
-# Copy the go source
+# Copy the rest
 COPY . .
 
-ARG TARGETOS TARGETARCH
```
jonstacks, line 16: I believe I had ARG down lower to not invalidate docker's layer cache since they aren't needed until you start building. Not sure if it still matters with buildkit and maybe newer docker versions handle it better, but just noting it

api/ngrok/v1alpha1/kubernetesoperator_types.go
alex-bezek: Sure thing

cmd/api-manager.go
alex-bezek: that was 1 piece, although to be honest, i'm not sure thats a good thing. I think the only 2 spec fields not set here are the metadata and region as those just end up with the defaults

```
    metadata: '{"owned-by":"ngrok-operator"}'
    region: global
```

so if someone edited that manually:
- in the old way, I believe we would call CreateOrUpdate and this mutateFN would set the whole spec potentially overriding missing fields with their defaults
- in this way it might preserve them

Honestly, the first one seems like how it should work ideally (fully reconciling this resource to what the operator wants it to be as it owns it) so i can revert this change and just set the drainPolicy spec field inside here. 

I think the other reason this changed is amp just thought it was better being more consistent, setting all the spec fields individually

internal/controller/ingress/ingress_controller.go
alex-bezek: I don't do a finalizer removal in any of the controllers just because its in draining mode. The drain.go code is responsible for removing finalizers from these objects. This check prevents it from re-adding the finalizer right back basically

cmd/api-manager.go
jonstacks: If we want to set them individually, that seems fine. I was just wondering if you hit anything in practice because of the full reassignment. Maybe it'll help prevent future foot guns? This is non-blocking IMO and just an OOC.

Dockerfile
alex-bezek: I'm reading that newer buildkit versions do make it so it only busts the cache layers where its used, not where the ARG line is placed. BUT things still recommend using the old organization approach of keeping the ARG declaration right above where you initially need it just incase someone builds with older docker versions or something. So i'll undo this part. Good eye 👁️

PR #766, opened 2026-02-23
Log warn messages from server-side

sabrina-ngrok:
Closes: K8SOP-226
## What
Emit warning logs in the operator if Notices are passed from Server-side through ConnResponse

## How
- Get `ConnResponse`
- Emit list of strings within `Notice` as warnings

## Testing
- Add a notice to the list using `resp.Notices = append(resp.Notices, &pb_agent.ConnResponse_Notice{Message: "This is a warning message"})`
```
2026-02-23T22:52:52Z    DEBUG   controllers.bindings-forwarder  WARNING: This is a warning message      {"resource": "bindings.k8s.ngrok.com/v1alpha1, Kind=BoundEndpoint Name=ngrok-f3624161-6020-5a17-8123-89d9631ff717", "endpoint-binding": {"name":"ngrok-f3624161-6020-5a17-8123-89d9631ff717","namespace":"ngrok-operator"}, "port": 10000, "remoteAddr": "10.244.0.15:57644", "ingress": {"endpoint":"kubernetes-binding-ingress.ngrok.io:443"}, "binding": {"host":"myservice.ngrok-operator","port":"80"}, "endpoint.id": "ep_3A5eJCIcK2TeYzNNy0rtnbVzUAu", "proto": "http", "pod identity": "uid:\"6dcf1a40-2dfb-4e5b-86ba-16d71a90649b\"  name:\"ngrok-operator-manager-6d54cb96f9-n5p8w\"  namespace:\"ngrok-operator\"  annotations:{key:\"k8s.ngrok.com/test\"  value:\"{\\\"env\\\": \\\"local\\\"}\"}"}
```

## Breaking Changes
¯\_(ツ)_/¯

### DIFF
--- a/internal/controller/bindings/forwarder_controller.go
+++ b/internal/controller/bindings/forwarder_controller.go
@@ -259,7 +259,7 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
 
 		// Upgrade the connection to a binding connection
 		resp, err := mux.UpgradeToBindingConnection(log, ngrokConn, host, port, podIdentity)
-		log = log.WithValues("endpoint.id", resp.EndpointId, "proto", resp.Proto)
+		log = log.WithValues("endpoint.id", resp.EndpointId, "proto", resp.Proto, "pod identity", podIdentity)
 		if err != nil {
 			log.Error(err, "failed to upgrade connection")
 			return err
@@ -272,7 +272,8 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
 		}
 
 		for _, notice := range resp.Notices {
-			log.Info("Warning", "message", notice.GetMessage())
+			noticeMsg := fmt.Sprintf("WARNING: %s", notice.GetMessage())
+			log.V(1).Info(noticeMsg)
 		}
 
 		log.Info("Bound connection")


internal/controller/bindings/forwarder_controller.go
```diff
@@ -271,6 +271,10 @@ func (r *ForwarderReconciler) update(ctx context.Context, epb *bindingsv1alpha1.
 			return err
 		}
 
+		for _, notice := range resp.Notices {
+			log.Info("Warning", "message", notice.GetMessage())
```
sabrina-ngrok: There is only `log.Info` and `log.Error`

alex-bezek: yea, this is that weird thing where you can also specify `log.V(1).Info` for debug and lower level logging. I think this is fine. 

A minor nit though would be to put the message in the first string and add additional meta information as the key value pairs similar to 

https://github.com/ngrok/ngrok-operator/blob/bc56b5211a22d25f12dbfb05f7bce67b48ca473b/cmd/api-manager.go#L324-L329

In the lines above we already add some additional details to the logger with `log.WithValues`. It would probably make sense to add the pod details to the logger as well either above using `WithValues` or explicitly log them here. The log would say something like `annotations is exceeding 1024 charactesrs and have been trucated.`. It might be nice to log the actual value in this case as well.

sabrina-ngrok: Updated description to show changes.

PR #773, opened 2026-03-05
Alex/object modified errors

alex-bezek:
## What

The operator will often have big scary error logs that say `the object has been modified; please apply your changes to the latest version and try again: %!s(<nil>)`

## How

This issue can happen when you have a reference to an object in memory and attempt to make an update to it but its been updated already by somethign else. When you call update, it checks that the resource version number hasn't changed and if it has it rejects it. The 2 main areas that can cause this are when we update an object to add its finalizer, and when we do updates to statuses. 

For the finalizers, they are simple and safe to just use patch on instead of update. This required adding patch permissions to their RBAC roles

For updating statuses, the RetryOnConflict can help with this and we already use it in places
https://github.com/ngrok/ngrok-operator/blob/aac9596856a5f834883ea50766511bb7f218aebb/pkg/managerdriver/driver.go#L838-L858

This adds RetryOnConflict to the driver's ingress status update and the base controller's status update

## Breaking Changes


internal/controller/base_controller.go
```diff
@@ -151,15 +153,39 @@ func (self *BaseController[T]) handleErr(op BaseControllerOp, obj T, err error)
...
+			// Re-fetch the latest resourceVersion for the next attempt.
+			// Status().Update() only writes the status subresource, so we only
+			// need the current resourceVersion — the spec/metadata are irrelevant.
+			latest := obj.DeepCopyObject().(T)
+			if getErr := self.Kube.Get(ctx, client.ObjectKeyFromObject(obj), latest); getErr != nil {
+				return getErr
+			}
+			obj.SetResourceVersion(latest.GetResourceVersion())
+		}
+		return err
```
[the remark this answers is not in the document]
alex-bezek: The concern about clobbering concurrent status updates doesn't apply here. For single-writer controllers (AgentEndpoint, CloudEndpoint, Domain, IPPolicy, etc.), ReconcileStatus is the sole status writer — the conflict is caused by metadata changes (e.g., the finalizer patch earlier in the same reconcile), not competing status writes. Refreshing only resourceVersion is safe in that case.

For the multi-writer case (BoundEndpoint), the caller already handles merge semantics itself — it re-GETs the latest object and copies only its owned fields onto the fresh copy before passing it to ReconcileStatus (see boundendpoint_controller.go:updateStatus). The poller does the same independently. So even on retry, the status payload already reflects the latest state of all writers.

Removing the retry would reintroduce the noisy "object modified" errors this PR is fixing

pkg/managerdriver/driver.go
```diff
@@ -706,11 +706,22 @@ func (d *Driver) updateIngressStatuses(ctx context.Context, c client.Client) err
...
+					return err
+				}
+
+				if reflect.DeepEqual(current.Status.LoadBalancer.Ingress, ingress.Status.LoadBalancer.Ingress) {
+					return nil
+				}
+
+				current.Status.LoadBalancer.Ingress = ingress.Status.LoadBalancer.Ingress
+				return c.Status().Update(ctx, current)
+			})
```
jonstacks: Looks like we dropped some error logging here. Do you think its worth capturing the err from retry.RetryOnConflict and logging within the goroutine?

jonstacks (approved):

pkg/managerdriver/driver.go
alex-bezek: updated

PR #782, opened 2026-03-16
Change full install manifests to main

sabrina-ngrok:
## What
manifest-bundle.yaml not being updated

## How
Existing GitHub action for this change branch to main

## Breaking Changes


jonstacks (changes requested):
We've had issues with using `pull_request_target` before. It can be a security foot gun if not handled properly since the permissions granted to it run within the context of our repo and not the fork's repo.

I think a good middle ground to unblock this for now and at least make sure we are keeping it up to date is to have a pre-flight check for it that runs on PRs like we do for `go mod tidy` in the main CI job and just fail the workflow if its not up to date.

.github/workflows/ci.yaml
```diff
@@ -49,6 +49,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs:
+    - changes
+    if: |
+      (needs.changes.outputs.charts == 'true') ||
+      (needs.changes.outputs.make == 'true')
+    steps:
+    - uses: actions/checkout@v6
+    - uses: nixbuild/nix-quick-install-action@v34
+    - uses: DeterminateSystems/magic-nix-cache-action@b8276522d77f21bf19d3574e5bc99186a6c7aa6c
```
jonstacks:
```suggestion
    - uses: ./.github/actions/nix-setup
```

.github/workflows/ci.yaml
```diff
@@ -49,6 +49,23 @@ jobs:
...
+    - name: Generate manifest bundle
+      run: nix develop --command make manifest-bundle
```
jonstacks:
```suggestion
      run: make manifest-bundle
```

.github/workflows/generate-full-install-manifests.yaml
jonstacks, line 1: Since the CI workflow above makes sure that this is up to date and runs in the merge-queue, we can just remove this file. I think the auto commit action at the bottom is going to have trouble pushing since we have merge queue checks.

jonstacks (changes requested):

jonstacks (approved):

PR #797, opened 2026-04-09
Fix/k8sop 246 kubernetesoperator fixes

alex-bezek:
related: K8SOP-240

## What
Three fixes to the KubernetesOperator controller that address error swallowing, nil pointer panics, and unreachable dead code.

- **Bug 1.4:** `findExisting()` returned `(nil, nil)` when `getNamespaceUID` failed, hiding RBAC/API server errors. Caller would create a duplicate KubernetesOperator.
- **Bug 1.6:** `Spec.Binding` and `Spec.Deployment` are optional pointers dereferenced without nil checks. A user setting `enabledFeatures: ["bindings"]` without a `binding:` block would panic.
- **Bug 2.2:** `findOrCreateTLSSecret` had inverted control flow making the validation block unreachable. Existing secrets were returned without checking for valid `tls.key`/`tls.csr`.

## How
- 1.4: Propagate the error: `return nil, fmt.Errorf("getting namespace UID: %w", err)`
- 1.6: Added nil guards before accessing optional pointer fields in `create`, `_update`, and `findExisting`
- 2.2: Restructured to a switch: check found (validate) → not-found (create) → other error (return)

Tests added: `TestFindExisting_GetNamespaceUIDError`, `TestNilBindingWithBindingsFeatureEnabled` (create + update), `TestNilDeployment_FindExisting`, `TestNilDeployment_Create`, `TestFindOrCreateTLSSecret` (5 table-driven cases)

## Breaking Changes
None


jonstacks (approved):
LGTM, no blocking issues.

Its a little bit of a departure from our normal tests that we have a separate spec file for the nil case, maybe consider consolidating it.

Also, not that it needs to be resolved in this PR, but we should make the tests for this operator similar to our other tests where we set up a an actual controller for this IMO

internal/controller/ngrok/kubernetesoperator_controller_test.go
```diff
@@ -43,3 +56,148 @@ func TestCalculateFeaturesEnabled(t *testing.T) {
...
+		interval = 500 * time.Millisecond
+	)
+
+	var kginkgo *testutils.KGinkgo
+
+	BeforeAll(func() {
+		kginkgo = testutils.NewKGinkgo(k8sClient)
+
+		// Ensure the controller namespace exists
+		ns := &v1.Namespace{
```
alex-bezek: @copilot please update these tests to use the kginkgo helpers more thuroughly. For example, this creation of a namespace and its checking that it succeeded could be using ExpectCreateNamespace https://github.com/ngrok/ngrok-operator/blob/main/internal/testutils/kginkgo.go#L126C19-L126C40

alex-bezek:
e2e failed uninstalling the operator after the e2e tests but there isn't much info to go off of. I've setup my local to use the same account and tried the same flow as ci multiple times and can't reproduce the issue. There is 1 leaked resource in the API so i'm going to delete that and try again since these code changes shouldn't effect the drain as far as i know
image

PR #798, opened 2026-04-09
Fix/k8sop 247 gateway controller fixes

alex-bezek:
related: K8SOP-240

## What
Three fixes to the Gateway and HTTPRoute controllers addressing validation logic, informer cache corruption, and cross-namespace matching.

- **Bug 2.1:** `validateGateway()` had inverted logic — returned an error when the gateway WAS valid (Accepted=True), and nil when invalid.
- **Bug 2.3:** `secretReferencedByGateway()` and `configMapReferencedByGateway()` mutated `certRef.Namespace` on cached Gateway objects via shared pointers, corrupting the informer cache.
- **Bug 2.17:** `findHTTPRouteForGateway` didn't default nil `parentRef.Namespace` to the route's namespace per Gateway API spec, matching any gateway with the same name across all namespaces.

## How
- 2.1: Added `!` to invert the condition: error when NOT accepted
- 2.3: Replaced cache mutations with local variable computation at 3 sites (BackendTLS cert ref, listener TLS cert refs, listener TLS CA cert refs)
- 2.17: Used `ptr.Deref(parentRef.Namespace, gatewayv1.Namespace(route.Namespace))` matching the correct pattern at line 239

Tests added: `validateGateway` (valid + invalid), `findHTTPRouteForGateway` cross-namespace test

## Breaking Changes
None


internal/controller/gateway/gateway_controller.go
```diff
@@ -449,13 +449,13 @@ func secretReferencedByGateway(secret *v1.Secret, c client.Client) bool {
 		// in the same layer as the rest of the translation offers a better user experience with understanding errors with their resources and why they happened.
 		if gw.Spec.BackendTLS != nil && gw.Spec.BackendTLS.ClientCertificateRef != nil {
 			certRef := gw.Spec.BackendTLS.ClientCertificateRef
-			if certRef.Namespace == nil {
-				certNs := gatewayv1.Namespace(gw.Namespace)
-				certRef.Namespace = &certNs
+			certNs := gw.Namespace
+			if certRef.Namespace != nil {
+				certNs = string(*certRef.Namespace)
 			}
```
jonstacks: I've seen us do both styles, maybe it makes sense to standardize on one or the other:

certNs := string(ptr.Deref(certRef.Namespace, gw.Namespace))

WDYT?

internal/controller/gateway/gateway_controller.go
```diff
@@ -465,12 +465,12 @@ func secretReferencedByGateway(secret *v1.Secret, c client.Client) bool {
 				continue
 			}
 			for _, certRef := range listener.TLS.CertificateRefs {
-				if certRef.Namespace == nil {
-					certNs := gatewayv1.Namespace(gw.Namespace)
-					certRef.Namespace = &certNs
+				certNs := gw.Namespace
+				if certRef.Namespace != nil {
+					certNs = string(*certRef.Namespace)
```
jonstacks: **nit**: Seeing some duplication in here that we might be able to clean up

internal/controller/gateway/gateway_controller_test.go
```diff
@@ -340,6 +340,71 @@ func newGateway(gwc *gatewayv1.GatewayClass) *gatewayv1.Gateway {
...
+			},
+		}
+
+		reconciler := &GatewayReconciler{
+			Client: k8sClient,
+			Scheme: k8sClient.Scheme(),
+			Driver: driver,
+		}
+
+		err := reconciler.validateGateway(ctx, gw)
```
jonstacks: This test feels a bit strange to me. validateGateway should already be getting called by the reconciler. IMO, We should prefer to test the reconcile loops by validating the CRs rather than directly calling methods.

internal/controller/gateway/httproute_controller_test.go
```diff
@@ -274,7 +277,7 @@ var _ = Describe("HTTPRoute controller", Ordered, func() {
 
 		When("the parent ref is an unsupported type", func() {
 			var (
-				service *v1.Service
+				service *corev1.Service
```
jonstacks: Any reason for changing to corev1?

internal/controller/gateway/httproute_controller_test.go
```diff
@@ -380,3 +383,79 @@ var _ = Describe("HTTPRoute controller", Ordered, func() {
...
+		Expect(k8sClient.Create(ctx, route)).To(Succeed())
+		defer func() { _ = k8sClient.Delete(ctx, route) }()
+
+		reconciler := &HTTPRouteReconciler{
+			Client: k8sClient,
+			Log:    logr.Discard(),
+		}
+
+		// When triggered by the gateway in the OTHER namespace, it should NOT match
+		requests := reconciler.findHTTPRouteForGateway(ctx, &gwOther)
```
jonstacks: Same logic here as for the above test. I'm not sure we get value by setting up the controllers to run and process CRs and then individually call methods on the reconcilers. I think for consistency, we should test from the outside(using CRs, mocking other services, etc) and then just validating that our code coverage covers the lines we want.

jonstacks (changes requested):
I'm open to other ways if there are strong benefits of doing it differently, but I think changing the tests to not call methods on the reconcilers is preferable.

internal/controller/gateway/httproute_controller_test.go
alex-bezek: I think i just picked this change up because the httproute_controller.go uses corev1. 
however we aren't actually consistent anywhere so I'll revert this

Image

jonstacks (approved):

PR #800, opened 2026-04-09
Fix/k8sop 249 manager driver fixes

alex-bezek:
related: K8SOP-240

## What
Four fixes to the manager driver addressing lost status conditions, reconcile churn from nondeterministic ordering, and informer cache corruption.

- **Bug 2.6:** `for _, listener := range newStatus.Listeners` created value copies — `SetStatusCondition` modified copies and the Programmed condition was silently discarded every iteration.
- **Bug 2.10:** Gateway status addresses built from map iteration had random order. `reflect.DeepEqual` saw false diffs, triggering spurious `Status().Update()` every sync cycle.
- **Bug 2.11:** Same pattern as 2.10 for Ingress LoadBalancer status — hosts deduped into map, iterated nondeterministically.
- **Bug 2.16:** `certRef.Namespace` mutated on cached gateway object from store in translation code (separate instance from the predicate bug fixed in the gateway controller PR).

## How
- 2.6: Index-based iteration with pointer: `for i := range` + `listener := &newStatus.Listeners[i]`
- 2.10: `sort.Slice(newStatus.Addresses, ...)` by Value before comparison
- 2.11: `sort.Slice(status, ...)` by Hostname before return
- 2.16: Compute namespace into local `certNs` variable instead of mutating cached ref

Tests added: Programmed condition persistence test, deterministic ordering test (20 iterations)

## Breaking Changes
None


jonstacks (approved):

alex-bezek:
@copilot resolve the merge conflicts in this pull request

PR #801, opened 2026-04-09
Fix/k8sop 250 bindings safety

alex-bezek:
related: K8SOP-240

## What
Three safety fixes to the bindings package addressing operator crashes, data races, and nil panics.

- **Bug 2.8:** Four `panic()` calls in `port_allocator.go` could crash the entire operator pod if a BoundEndpoint had a port outside the configured range. For a long-running service maintaining connections, crash-looping is unacceptable.
- **Bug 2.25:** `startPollingAPI` reassigned `r.portAllocator` while `deleteBinding` called methods on it concurrently — unsynchronized data race.
- **Bug 1.7:** `*op.Spec.Binding.IngressEndpoint` dereferenced without nil check in forwarder controller logging.

## How
- 2.8: All panic calls → error returns. `IsSet` now returns `(bool, error)`, `Unset` now returns `error`. All callers updated.
- 2.25: Added `sync.Mutex` protecting portAllocator reassignment and all method calls.
- 1.7: Added `derefString()` helper for safe *string dereference.

Tests added: `Test_portBitmap_outOfRange` (error returns for out-of-range ports), `TestDerefString`

## Breaking Changes
`portBitmap.IsSet` signature changed from `bool` to `(bool, error)`. `portBitmap.Unset` signature changed from `void` to `error`. Internal API only — not exposed to users.


internal/controller/bindings/boundendpoint_poller.go
```diff
@@ -244,7 +249,9 @@ func (r *BoundEndpointPoller) reconcileBoundEndpointsFromAPI(ctx context.Context
 	}
 
 	// reassign port allocations
+	r.portAllocatorMu.Lock()
 	r.portAllocator = currentPortAllocations
+	r.portAllocatorMu.Unlock()
```
jonstacks: **Non-blocking**: Maybe move the mutex into the portAllocator itself so that a portAllocator is safe for concurrent use

jonstacks (approved):

internal/controller/bindings/boundendpoint_poller.go
alex-bezek: addressed in this commit https://github.com/ngrok/ngrok-operator/pull/801/changes/8941c1941a6cf0bc3bcfe0a8ca68990ab68c8356 

We created a new function Replace for setting the ports that handles locking and unlocking the mutex. 

It also removed multiple other lock/unlock calls that surrounded some other calls that it added now that we understand the package already had its own mutex internally already
image

jonstacks (approved):

PR #805, opened 2026-04-16
DRAFT: v1 Spec

jonstacks:
## What

We are working toward a v1 of the ngrok-operator. This PR introduces a formal specification (`/specs`) that defines the operator's public-facing surface area as we stabilize it for a 1.0 release.

The spec covers:

### CRDs (`specs/crds/`)
Stable API contracts for each Custom Resource:
- `AgentEndpoint`, `CloudEndpoint`, `BoundEndpoint` — endpoint types and their schemas
- `TrafficPolicy`, `Domain`, `IPPolicy`, `KubernetesOperator` — supporting resources
- Shared patterns: conditions, finalizers, and common types

### Annotations (`specs/annotations.md`)
A central reference for all `ngrok.com/` annotations — what they do, which resources they apply to, and how they interact (e.g. `pooling-enabled`, mapping strategy).

### Helm Chart (`specs/helm/`)
Desired state for `values.yaml` as a first-class API:
- Global config, credentials, and the config file system
- Per-component values: `apiManager`, `agent`, `bindingsForwarder`
- Feature flags, drain/domain policies, and the cleanup hook

### Features (`specs/features/`)
Behavior specifications for cross-cutting capabilities:
- **Draining** — graceful resource cleanup on uninstall
- **Bindings** — projecting external ngrok endpoints into the cluster as Kubernetes Services
- **Ingress & Gateway API** — how those controllers work and how they map to ngrok endpoints
- **High Availability** — replica counts, leader election, PDBs
- **Traffic Policy** — resolution and precedence across controllers
- **Multi-install** — running multiple operator instances in the same cluster
- **Namespace watching** — scoping the operator to specific namespaces

### Controllers (`specs/controllers/`)
Reconciliation behavior for each controller, including error handling, drain awareness, and how individual controllers interact with the Driver pattern (Ingress/Gateway API).

### RBAC (`specs/rbac/`)
Permission sets for each component (operator, agent, bindings forwarder) and aggregation roles.

### Authentication (`specs/authentication.md`)
Credentials, secrets management, API key vs. authtoken handling.

### Design Decisions (`specs/design-decisions.md`)
Settled architectural trade-offs — notably:
- Following Kubernetes API conventions throughout
- Deferring field validation to the ngrok API rather than duplicating it in admission webhooks

## Why

The operator's API has grown organically. Before cutting a v1, we need a stable, documented contract so that:
1. We can distinguish bugs from unspecified behavior
2. Breaking changes are intentional and tracked
3. Contributors have an authoritative reference before opening PRs

The spec directory acts as a **living document** — expect iterations, comments, and open questions as we work toward sign-off.

## How

This is a review-first PR. The spec was generated from the current operator state and then hand-edited to reflect the desired v1 behavior. Feedback via PR comments is the primary workflow — treat each spec file as its own review surface.

No code changes are included in this PR.

## Breaking Changes

None — this PR contains specifications only. Breaking changes may be introduced in subsequent implementation PRs that align the code to the spec for v1.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive specs covering controller reconciliation flows, new CRD schemas and status/conditions, feature behaviors (Ingress, Gateway API, Bindings, Draining, HA, TrafficPolicy, Namespace scoping, Multi-install), Helm chart values and hooks, RBAC proposals, and design/migration guidance for the ngrok-operator.


specs/controllers/agentendpoint.md
```diff
@@ -0,0 +1,60 @@
...
+## Created Resources
+
+- ngrok agent endpoint (via AgentDriver API)
+- `Domain` CR (via DomainManager)
+
+## Status
+
+| Field                    | Description                              |
+|--------------------------|------------------------------------------|
+| `assignedURL`            | The URL assigned by ngrok                |
```
jonstacks: These descriptions are a bit short. Should we include more details. For example, if setting `url: https://`, the assignedURL will be a fully expanded URL based on our rules. Worth noting here?

specs/features/bindings.md
```diff
@@ -0,0 +1,50 @@
+# Endpoint Bindings Feature
+
+## Overview
+
+The endpoint bindings feature allows ngrok endpoints to be "bound" into a Kubernetes cluster, projecting external ngrok endpoints as local Kubernetes services. This enables traffic from ngrok to flow directly to services inside the cluster.
```
jonstacks, line 5: This is a bit misleading. This is for projecting endpoints that exist in ngrok into you cluster. For example, you could have an endpoint `https://service-a.namespace-a` that will appear as a k8s service named `service-a` in `namespace-a`.

The benefit to this is that pods in your cluster talk directly to a kubernetes service and that traffic is then securely tunneled to ngrok without needing to expose the target service on the internet.

specs/features/bindings.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Status:** In development (`features.bindings.enabled: false` by default).
```
jonstacks: This feature is no longer in development for v1. We should also look if anything else is marked as in development across the rest of the specs.

Open Question: For v1, do we want this to be on by default or keep it as opt-in still? We're leaning towards on by default.

specs/features/bindings.md
```diff
@@ -0,0 +1,50 @@
...
+| `features.bindings.serviceLabels`       | Labels applied to projected services                  | `{}`                                      |
+| `features.bindings.ingressEndpoint`     | Hostname of the bindings ingress endpoint             | `kubernetes-binding-ingress.ngrok.io:443` |
+
+## Components
+
+| Component            | Description                                              |
+|----------------------|----------------------------------------------------------|
+| BoundEndpoint CRD    | Represents an endpoint bound from ngrok to a service     |
+| BoundEndpoint controller | Creates target and upstream services for each binding |
+| Bindings Forwarder   | Bridges connections via mTLS to the ngrok ingress        |
```
jonstacks, line 25: Would be nice to link to other parts of the spec here.

specs/features/bindings.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Flow
+
+1. The operator registers with the ngrok API as a Kubernetes operator with bindings enabled.
+2. The ngrok API pushes bound endpoint information to the operator via a poller.
```
jonstacks: "pushes" & "poller" are a bit in conflict here.

specs/README.md
```diff
@@ -0,0 +1,89 @@
...
+1. **Implementation planning**: Provides a baseline for planning changes as we work toward v1.
+2. **Bug vs. unspecified behavior**: Defines expected behavior so we can distinguish bugs from unspecified behavior.
+3. **Living documentation**: Serves as the canonical reference for the operator's public surface area.
+
+## Glossary
+
+| Term | Definition | Details |
+|------|-----------|---------|
+| **Agent Endpoint** | An endpoint backed by an ngrok agent tunnel running inside the cluster. Created as an `AgentEndpoint` CR. | [crds/agentendpoint.md](crds/agentendpoint.md) |
+| **Bindings** | A feature that projects external ngrok endpoints into the cluster as local Kubernetes Services, enabling ngrok-to-cluster traffic. | [features/bindings.md](features/bindings.md) |
```
alex-bezek:
```suggestion
| **Bindings** | A feature that projects private ngrok endpoints into the cluster as local Kubernetes Services, enabling pods to send requests to them without those endpoints being on the public internet. | [features/bindings.md](features/bindings.md) |
```

alex-bezek: The other items in this table say how they are created via specific CRs or annotations. It might be worth noting how here as well. Something like "Both Agent Endpoints and Cloud Endpoints can be bound to kubernetes with a bindings=kubernetes value and can be created via CRs, the api, dashboard, etc with the same k8s cluster or other cluster, etc" probably more concisely though :)

alex-bezek: they also all link to more detailed docs so maybe the docs cover it fine

specs/annotations.md
```diff
@@ -0,0 +1,99 @@
...
+| Detail          | Value                                                  |
+|-----------------|--------------------------------------------------------|
+| Applies to      | `Service` (LoadBalancer), `Ingress`, `Gateway` routes  |
+| Allowed values  | `endpoints`, `endpoints-verbose`                       |
+| Default         | `endpoints`                                            |
+
+- `endpoints`: Creates only an `AgentEndpoint`.
+- `endpoints-verbose`: Creates both a `CloudEndpoint` and an internal `AgentEndpoint` (URL ending in `.internal`).
+
+See: [controllers/service.md](controllers/service.md), [controllers/ingress.md](controllers/ingress.md)
```
alex-bezek, line 34: The mapping strategy might warrant its own spec/doc. While the verbose one is clear what it does, not supplying one and letting it be dynamic is a bit confusing. And the underlying reason for it existing is somethign we've talked about but would be good to write down

specs/README.md
```diff
@@ -0,0 +1,89 @@
...
+### [features/](features/) — Cross-Cutting Features
+
+- [draining.md](features/draining.md) — Drain policy, cleanup hook, uninstall behavior
+- [multi-install.md](features/multi-install.md) — Multiple operator installations in one cluster
+- [ingress.md](features/ingress.md) — Kubernetes Ingress feature
+- [gateway-api.md](features/gateway-api.md) — Kubernetes Gateway API feature
+- [bindings.md](features/bindings.md) — Endpoint bindings feature
+- [high-availability.md](features/high-availability.md) — Replicas, leader election, PDB
+- [traffic-policy.md](features/traffic-policy.md) — Traffic policy resolution across controllers
+- [namespace-watching.md](features/namespace-watching.md) — Namespace scoping configuration
```
alex-bezek, line 57: I think this is good for now, and this is just a hunch/though, but I'm thinking long term we'll want more of the cross-cutting specs and less "helm" specs. Like most of these features have some impact across the stack from the helm values to install, how the crd is used, and how the reconciler works and deals with different scenarios. It feels like those should be documented mostly in here. 

It probably still makes sense to have some helm specs, but I think it might not get into every helm value for the agent-manager for example, but instead document how we want something like "Global helm values that can be overriden per component. and separatation of k8s configs from controller configs" and might use some examples

specs/README.md
```diff
@@ -0,0 +1,89 @@
...
+### [crds/](crds/) — Custom Resource Definitions
+
+- [common.md](crds/common.md) — Shared patterns: conditions, finalizers, shared types
+- [agentendpoint.md](crds/agentendpoint.md) — AgentEndpoint (`ngrok.com`)
+- [cloudendpoint.md](crds/cloudendpoint.md) — CloudEndpoint (`ngrok.com`)
+- [kubernetesoperator.md](crds/kubernetesoperator.md) — KubernetesOperator (`ngrok.com`)
+- [trafficpolicy.md](crds/trafficpolicy.md) — TrafficPolicy (`ngrok.com`)
+- [domain.md](crds/domain.md) — Domain (`ngrok.com`)
+- [ippolicy.md](crds/ippolicy.md) — IPPolicy (`ngrok.com`)
+- [boundendpoint.md](crds/boundendpoint.md) — BoundEndpoint (`ngrok.com`)
```
alex-bezek, line 68: Nit: i was confused what all the ngrok.com stuff was for. I'm guessing this is supposed to be the api group and is shown here as a relic of the fact that the existing system is not consistent? might be worth removing if they are consistent in this new world

specs/README.md
```diff
@@ -0,0 +1,89 @@
...
+| **Drain / Draining** | The process of gracefully removing ngrok API resources when the operator is uninstalled. Triggered by deletion of the `KubernetesOperator` CR. | [features/draining.md](features/draining.md) |
+| **Driver** | An internal pattern used by Ingress and Gateway API controllers that collects state from multiple resources and materializes the combined state as ngrok endpoints. | [controllers/ingress.md](controllers/ingress.md) |
+| **Endpoint Pooling** | Allows multiple endpoints to share the same public URL, distributing traffic across them. Controlled via the `ngrok.com/pooling-enabled` annotation. | [annotations.md](annotations.md) |
+| **Mapping Strategy** | Controls which ngrok endpoint resources are created: `endpoints` (AgentEndpoint only) or `endpoints-verbose` (CloudEndpoint + internal AgentEndpoint). | [annotations.md](annotations.md) |
+| **Traffic Policy** | A set of rules (rate limiting, header manipulation, authentication, etc.) applied to ngrok endpoints. Defined as an `TrafficPolicy` CR or inline JSON. | [features/traffic-policy.md](features/traffic-policy.md) |
+
+## Directory Structure
```
alex-bezek, line 24: Gonna leave a list of notes here for additional docs we might consider adding that I think of while reading. 

The first that comes to mind is something documenting the high level architecture of the 3 different deployable components and their differences in usage of leader election, namespace scoping, etc

specs/authentication.md
```diff
@@ -0,0 +1,75 @@
+# Authentication
+
+## Overview
+
+The ngrok-operator authenticates with the ngrok API using two credentials:
```
alex-bezek:
```suggestion
The ngrok-operator authenticates with ngrok using two credentials:
```

specs/authentication.md
```diff
@@ -0,0 +1,75 @@
...
+- The **agent-manager** uses `NGROK_AUTHTOKEN` for establishing agent tunnels.
+- The **bindings-forwarder** uses `NGROK_AUTHTOKEN` for its connections.
+
+## mTLS for Bindings
+
+When the bindings feature is enabled, the operator generates a self-signed TLS certificate and creates a Certificate Signing Request (CSR) with the ngrok API. This certificate is stored in a Secret (default name: `default-tls`) in the operator's namespace and is used for mTLS communication between the bindings forwarder and ngrok's ingress endpoint.
+
+## One-Click Demo Mode
+
+When `oneClickDemoMode: true` is set, the operator starts without requiring credentials. It will report as Ready but will not actually connect to the ngrok API. This mode is intended for demonstration purposes only.
```
alex-bezek: I wonder if this deserves its own page as well. I wouldn't say for example that its for demonstration purposes only. My understanding was that it was for marketplaces that require a way of installing an operator without any user provided indput. 

This doc reads kind of like human documentation vs a requirement bullet that is like "The operator MUST have the 1 of the 3 helm properties provided: demo mode, credential name, or the 2 credentials". maybe its just the terse-ness of old specs/reqs I was used to vs these. I'm not sure how to suggest a solid change though, maybe somethign we can iterate on. Or if we end up finding a good format we like we can put it into an agent.md, custom agent, or skill. There probably are open source skills for writing specs, so we could try taking these and transforming them into different flavors and seeing what we like

specs/rbac/README.md
```diff
@@ -0,0 +1,43 @@
+# RBAC
+
+## Overview
+
+The ngrok-operator uses Kubernetes Role-Based Access Control (RBAC) to authorize its components to interact with the Kubernetes API. Three separate components each have their own ServiceAccount and RBAC configuration:
+
+| Component            | Scope          | Purpose                                      |
+|----------------------|----------------|----------------------------------------------|
+| Operator (api-manager) | Cluster + Namespace | Main controller managing all CRDs and resources |
+| Agent                | Cluster        | Agent tunnel management                       |
```
alex-bezek: A random thought since i saw the word `tunnel`, maybe the glossary should have legacy terms with instructions on how to migrate them? so tunnels would be endpoints perhaps, kubernetes-ingress-controller would be ngrok-operator, etc

specs/rbac/README.md
```diff
@@ -0,0 +1,43 @@
...
+| Bindings Forwarder   | Cluster + Namespace | Endpoint binding forwarding                   |
+
+## Design Principles
+
+- **Least privilege**: Each component only has the permissions it needs.
+- **Cluster-scoped for CRDs**: CRDs and cross-namespace resources require ClusterRoles.
+- **Namespace-scoped for internal state**: Leader election and secret management use namespaced Roles.
+- **Conditional permissions**: Some permissions (e.g., secret management for bindings) are only granted when the corresponding feature is enabled.
```
alex-bezek, line 73: I like this break out of the main readme having high level info and cross cutting concerns and then having the discrete permission lists in sub folders. Probably can just refactor my newly merged requirements.md file into this format

specs/rbac/aggregation.md
```diff
@@ -0,0 +1,41 @@
...
+| `agentendpoint-viewer-role`              | `agentendpoints`, `agentendpoints/status` |
+| `ippolicy-viewer-role`                   | `ippolicies`, `ippolicies/status` |
+| `kubernetesoperator-viewer-role`         | `kubernetesoperators`, `kubernetesoperators/status` |
+| `trafficpolicy-viewer-role`         | `trafficpolicies`, `trafficpolicies/status` |
+
+**Verbs:** get, list, watch
+
+## Custom Annotations
+
+The `clusterRole.annotations` Helm value allows adding custom annotations to all ClusterRoles. This is commonly used for RBAC aggregation — for example, aggregating the editor role into the built-in `admin` or `edit` roles so that users with those roles automatically gain CRD access without needing `system:masters`.
```
alex-bezek: what are your thoughts on including shortcomings, things we aren't handling but have talked about in here? like we discussed how we support this annotation right now, but its not that helpful since you probably would want to assign the edit roles to the admin group and the viewer ones to something else.

alex-bezek: this value changed btw

specs/helm/common.md
```diff
@@ -0,0 +1,114 @@
...
+## Chart Structure
+
+The ngrok-operator ships as two Helm charts:
+
+| Chart             | Purpose                          |
+|-------------------|----------------------------------|
+| `ngrok-operator`  | Operator deployments and RBAC    |
+| `ngrok-crds`      | Custom Resource Definitions      |
+
+The CRD chart can be installed automatically via `installCRDs: true` (default) or separately.
```
alex-bezek, line 12: this level of documentation feels a bit separate. the below stuff primarily seems to document how in the main operator chart, how we want to handle globals and service specific overrides

specs/helm/common.md
```diff
@@ -0,0 +1,114 @@
...
+| `global.affinity`                        | Affinity rules                                   | `{}`           |
+| `global.topologySpreadConstraints`      | Topology spread constraints                      | `[]`           |
+| `global.priorityClassName`              | Pod priority class                               | `""`           |
+| `global.resources`                       | Container resource requests/limits               | `{}`           |
+| `global.extraVolumes`                    | Additional volumes                               | `[]`           |
+| `global.extraVolumeMounts`              | Additional volume mounts                         | `[]`           |
+| `global.extraEnv`                        | Additional environment variables                 | `{}`           |
+| `global.lifecycle`                       | Container lifecycle hooks                        | `{}`           |
+| `global.terminationGracePeriodSeconds`  | Graceful shutdown time                           | `30`           |
+
```
alex-bezek, line 59: Might want to include some strong language here driving the requirement that any setting that is in all 3 components should be here as well. Like podDisruptionBudget is one thats present in them all and could be here as well

specs/helm/operator.md
```diff
@@ -0,0 +1,44 @@
...
+| `apiManager.updateStrategy.type`               | Update strategy type                                 | `RollingUpdate` |
+| `apiManager.podDisruptionBudget.create`        | Enable PDB creation                                  | `false`         |
+| `apiManager.podDisruptionBudget.maxUnavailable`| Max unavailable pods                                 | `"1"`           |
+| `apiManager.podDisruptionBudget.minAvailable`  | Min available pods                                   | (unset)         |
+| `apiManager.serviceAccount.create`             | Create a ServiceAccount                              | `true`          |
+| `apiManager.serviceAccount.name`               | ServiceAccount name (auto-generated if empty)        | `""`            |
+| `apiManager.serviceAccount.annotations`        | ServiceAccount annotations                           | `{}`            |
+| `apiManager.clusterRole.annotations`           | Annotations for all ClusterRoles                     | `{}`            |
+
+A minimum of 2 replicas is recommended in production for high availability.
```
alex-bezek: seems a bit random

specs/helm/operator.md
```diff
@@ -0,0 +1,44 @@
...
+
+## App Config
+
+Component-specific app config rendered into the api-manager ConfigMap. Overrides values from the common ConfigMap (`ngrok.*`).
+
+| Parameter                                        | Description                                   | Default  |
+|--------------------------------------------------|-----------------------------------------------|----------|
+| `apiManager.config.oneClickDemoMode`             | Start without credentials for demo purposes   | `false`  |
```
alex-bezek, line 42: i was curious why this was such a small list, but i guess once you extract out the common pieces, there are relatively few specific options here. This did make me think though, things like "region" or "api-url" we pulled up to top level configs, should this oneClickDemo be pulled up to? it feels like if there is no valid reason you should/could configure these differently, then we shouldn't let you. But we also don't have to scope the config to only the component that uses it

specs/helm/features.md
```diff
@@ -0,0 +1,74 @@
...
+| Parameter                                | Description                                                      | Default    |
+|------------------------------------------|------------------------------------------------------------------|------------|
+| `features.drainPolicy`                   | Drain policy on uninstall: `"Delete"` or `"Retain"`             | `"Retain"` |
+| `features.defaultDomainReclaimPolicy`    | Default reclaim policy for Domains: `"Delete"` or `"Retain"`   | `"Delete"` |
+
+## Cleanup Hook
+
+| Parameter                              | Description                                  | Default              |
+|----------------------------------------|----------------------------------------------|----------------------|
+| `cleanupHook.enabled`                  | Enable the pre-delete cleanup hook           | `true`               |
```
alex-bezek, line 58: probably can also nest the cleanup hook under features

specs/features/traffic-policy.md
```diff
@@ -0,0 +1,60 @@
+# Traffic Policy
+
+## Overview
+
+TrafficPolicy is a cross-cutting resource that defines traffic handling rules (rate limiting, header manipulation, authentication, etc.) applied to ngrok endpoints. It can be referenced by multiple controllers and applied to endpoints in several ways.
```
alex-bezek, line 5: thoughts on leaving links to ngrok docs pages for these to lead AI to sources of truth for reference details https://ngrok.com/docs/traffic-policy/index#traffic-policy-overview ? my only hesitation is they could change but we could deal with this by having something routinely check that links are good

specs/features/traffic-policy.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Application Methods
+
+### 1. CRD Reference
+
+Traffic policies can be referenced directly on endpoint CRDs:
+
+- **AgentEndpoint**: `spec.trafficPolicy.targetRef` (K8sObjectRef) or `spec.trafficPolicy.inline` (raw JSON)
+- **CloudEndpoint**: `spec.trafficPolicyName` (string) or `spec.trafficPolicy` (inline TrafficPolicySpec)
```
alex-bezek: this is documenting the current state, but i think we actually want this to document the final state that https://linear.app/ngrok/issue/K8SOP-87/cloud-endpoint-crd-should-support-inline-and-ref-for-traffic-policy will transition us to

specs/features/traffic-policy.md
```diff
@@ -0,0 +1,60 @@
...
+The `ngrok.com/traffic-policy` annotation on parent resources (Service, Ingress, Gateway routes) references a TrafficPolicy by name in the same namespace:
+
+```yaml
+annotations:
+  ngrok.com/traffic-policy: "my-policy"
+```
+
+### 3. Inline on Parent Resources
+
+Some parent controllers support inline traffic policy configuration via their own mechanisms (e.g., the Gateway API's extensionRef).
```
alex-bezek: I'm not very familiar with the gateway api extensionRef, but it feels like this section should be documenting what we plan to do for https://linear.app/ngrok/issue/K8SOP-166/annotation-for-inline-traffic-policy-for-ingress-and-httproute

specs/annotations.md
```diff
@@ -0,0 +1,99 @@
...
+
+### `ngrok.com/traffic-policy`
```
alex-bezek, line 36: we plan on adding another annotation in this ticket https://linear.app/ngrok/issue/K8SOP-166/annotation-for-inline-traffic-policy-for-ingress-and-httproute

specs/features/traffic-policy.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Resolution with Mapping Strategy
+
+The `ngrok.com/mapping-strategy` annotation affects where the traffic policy is applied:
+
+| Mapping Strategy     | Policy Applied To   |
+|----------------------|---------------------|
+| `endpoints`          | `AgentEndpoint`     |
+| `endpoints-verbose`  | `CloudEndpoint`     |
```
alex-bezek, line 46: maybe would be better to link to a common guide on how these mapping strategies work

specs/features/traffic-policy.md
```diff
@@ -0,0 +1,60 @@
...
+## Validation
+
+The TrafficPolicy controller performs:
+
+1. **JSON syntax validation**: Ensures the policy field contains valid JSON.
+2. **Deprecation warnings**: Emits events for deprecated features:
+   - Legacy `directions` field
+   - `enabled` field on rules
+
+The policy content is **schemaless** — the operator does not enforce structure beyond valid JSON. Schema validation is performed by the ngrok API when the policy is applied to an endpoint.
```
alex-bezek, line 61: Alan was talking about in his experience explaining the why is very helpful here. We explain much more of the reason why in `specs/design-decisions.md`, maybe we just refer back to that?

specs/features/traffic-policy.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Events
+
+| Event                     | Description                         |
+|---------------------------|-------------------------------------|
+| `TrafficPolicyParseFailed`| Emitted when JSON parsing fails     |
+| `PolicyDeprecation`       | Emitted when deprecated features are used |
```
alex-bezek, line 68: maybe worth adding more color here that its not just deprecated features, but more specific to the deprecated phases `"Traffic Policy is using legacy directions: ['inbound', 'outbound']. Update to new phases: ['on_tcp_connect', 'on_http_request', 'on_http_response']"`

specs/features/namespace-watching.md
```diff
@@ -0,0 +1,38 @@
...
+| Ingress                   | Yes                      |
+| Gateway API (all)         | Yes                      |
+| Service LoadBalancer       | Yes                      |
+| AgentEndpoint (agent)     | Yes                      |
+| KubernetesOperator        | No (always watches only its own CR) |
+| BoundEndpoint             | No (watches operator namespace only) |
+
+## Caveats
+
+- If the watch namespace does not match the operator's namespace, the KubernetesOperator CR may not be visible to the cache. The KubernetesOperator controller uses a name+namespace predicate that is independent of the cache scope, but other controllers that need to read the KubernetesOperator may be affected.
```
alex-bezek: i might rephrase this as its just not supported. I think its bigger than just things "may be effected", i think if the operator can't register then a handful of things won't work. IMO we just say its definitely not supported right now

specs/features/namespace-watching.md
```diff
@@ -0,0 +1,38 @@
...
+- Domain and IPPolicy controllers operate on resources created by other controllers, so they inherit the effective scope of the controllers that create them.
```
alex-bezek: this statement is a bit weird to me. Domains are created by the driver but IP Policies aren't right? and they still get scoped to just watch specific namespaces

specs/features/multi-install.md
```diff
@@ -0,0 +1,72 @@
...
+
+Each operator instance maintains independent drain state:
+
+- `DrainOrchestrator` is configured with the specific `K8sOpName` and `K8sOpNamespace`
+- `StateChecker` only monitors its own KubernetesOperator CR
+- Draining one installation does not affect others
+
+### Namespace Watching
+
+Each deployment can be scoped to watch different namespaces:
```
alex-bezek:
```suggestion
Each deployment MUST be scoped to watch different namespaces:
```

specs/features/ingress.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Behavior
+
+When enabled, the operator:
+
+1. Creates an `IngressClass` resource (if `ingress.ingressClass.create` is true) with the configured controller name.
+2. Watches Ingress resources that reference the operator's IngressClass.
+3. For each matching Ingress, creates `AgentEndpoint` and/or `CloudEndpoint` resources based on the mapping strategy.
+4. Manages Domain resources for the hostnames specified in the Ingress rules.
+5. Updates the Ingress status with endpoint information.
```
alex-bezek, line 26: It feels like this should be fleshed out more. I'm not sure if this is the best document to put it in, but it seems like we should document how the load balancer status stuff gets set based on your different domain type. unless thats something more common that we should document somewhere more like domains

specs/features/high-availability.md
```diff
@@ -0,0 +1,47 @@
...
+## Overview
+
+The ngrok-operator supports running multiple replicas for high availability. Only one replica actively reconciles at a time via leader election; standby replicas take over if the leader fails.
+
+## Replica Configuration
+
+| Component          | Helm Value                            | Default | Recommendation       |
+|--------------------|---------------------------------------|---------|----------------------|
+| API Manager        | `apiManager.replicaCount`             | `1`     | 2+ in production     |
+| Agent              | `agent.replicaCount`                  | `1`     |                      |
```
alex-bezek: This should be documented to be higher. honestly we don't even run more than 1 in prod for API manager. seems like this document is primarily focused on the api manager's leader election usage but isn't talking about HA for the agent pods and doesn't meantion HA for the bindings pods either. maybe separate docs for each?

specs/features/high-availability.md
```diff
@@ -0,0 +1,47 @@
...
+| Bindings Forwarder | `bindingsForwarder.replicaCount`      | `1`     |                      |
+
+## Leader Election
+
+Leader election ensures only one operator replica actively reconciles at a time.
+
+| Setting         | Description                               | Default                     |
+|-----------------|-------------------------------------------|-----------------------------|
+| `--election-id` | ConfigMap/Lease name for leader election  | `ngrok-operator-leader`     |
```
alex-bezek, line 23: we do document these flags that we pass to controllers, but we do plan to get rid of those in favor of config files. It might be fine to keep the spec focused on the current usage of cobra cli flags though since we haven't really gone all the way through designing what those config files would be like. or we can do that up front here and document the pattern now

specs/features/high-availability.md
```diff
@@ -0,0 +1,47 @@
...
+- **Leader loss:** When the leader pod is lost, the lease expires (~15 seconds default TTL) and a standby replica acquires leadership.
+- **Graceful shutdown:** Signal handlers allow cleanup before relinquishing leadership.
+
+## Pod Disruption Budget
+
+| Helm Value                                    | Description                                    | Default |
+|-----------------------------------------------|------------------------------------------------|---------|
+| `apiManager.podDisruptionBudget.create`       | Enable PDB creation                            | `false` |
+| `apiManager.podDisruptionBudget.maxUnavailable` | Max unavailable pods                         | `"1"`   |
+| `apiManager.podDisruptionBudget.minAvailable`   | Min available pods                           | (unset) |
```
alex-bezek: we do have these for agent and bindings as well (although no global one which i think i commented about on that file). might be worth documenting the others as well

specs/features/high-availability.md
```diff
@@ -0,0 +1,47 @@
...
+Anti-affinity is configured via the standard `affinity` field on each component (or `global.affinity` for all components). There are no preset helpers — write affinity rules directly.
+
+| Helm Value                | Description                                    | Default |
+|---------------------------|------------------------------------------------|---------|
+| `global.affinity`         | Affinity rules for all components              | `{}`    |
+| `apiManager.affinity`     | Affinity rules for the api-manager (overrides global) | `{}`    |
+
+## Drain State Across Replicas
+
+When draining is initiated, the drain state propagates across replicas via the KubernetesOperator CR's `status.drainStatus` field. This ensures all replicas observe the drain state regardless of which replica is the leader.
```
alex-bezek, line 61: hmm, i guess this is technically correct. But also if you have multiple api managers, only 1 would be leader elected to be watching the operator anyways

specs/features/gateway-api.md
```diff
@@ -0,0 +1,60 @@
...
+| `GatewayClass`   | `gateway.networking.k8s.io/v1`       | Defines the controller type        |
+| `Gateway`        | `gateway.networking.k8s.io/v1`       | Configures listeners and addresses |
+| `HTTPRoute`      | `gateway.networking.k8s.io/v1`       | HTTP routing rules                 |
+| `TCPRoute`       | `gateway.networking.k8s.io/v1alpha2` | TCP routing rules                  |
+| `TLSRoute`       | `gateway.networking.k8s.io/v1alpha2` | TLS routing rules                  |
+| `ReferenceGrant` | `gateway.networking.k8s.io/v1beta1`  | Cross-namespace reference grants   |
+
+## Behavior
+
+When enabled and Gateway API CRDs are detected:
```
alex-bezek, line 29: might be worth calling out specifically that we don't install the gateway CRDs for you and the user has to bring their own

specs/features/gateway-api.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. The operator registers a `GatewayClass` with its controller name.
```
alex-bezek: I don't think this is the case and I don't see a ticket to do this in 1.0, is doing this something you are thinking we should add or did it slip into the docs because thats how ingress works?

specs/features/ingress.md
```diff
@@ -0,0 +1,45 @@
...
+The following annotations on Ingress resources influence behavior:
+
+- `ngrok.com/mapping-strategy` — Controls endpoint creation strategy
+- `ngrok.com/traffic-policy` — References a TrafficPolicy
+- `ngrok.com/pooling-enabled` — Enables endpoint pooling
+- `ngrok.com/description` — Sets endpoint description
+- `ngrok.com/metadata` — Sets endpoint metadata
+
+See [annotations.md](../annotations.md) for details.
+
```
alex-bezek, line 39: the gateway specs point out how the driver is used for it, might add the same thing for ingress

specs/features/gateway-api.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Annotations
+
+The following annotations on Gateway API resources influence behavior:
+
+- `ngrok.com/mapping-strategy` — Controls endpoint creation strategy
+- `ngrok.com/traffic-policy` — References a TrafficPolicy
+- `ngrok.com/pooling-enabled` — Enables endpoint pooling
+- `ngrok.com/description` — Sets endpoint description
+- `ngrok.com/metadata` — Sets endpoint metadata
```
alex-bezek, line 52: are these only supported for specific gateway resource types like gateways vs routes?

specs/features/traffic-policy.md
```diff
@@ -0,0 +1,60 @@
+# Traffic Policy
```
alex-bezek, line 1: I'm not sure if this is the best spot, but I think we should document how the operator adds its own specifc parts of the traffic policy for you that does stuff like forward internal. There is code we have that has to specify phases and doesn't follow our design decision of being transparent for this reason

specs/crds/common.md
```diff
@@ -0,0 +1,98 @@
...
+
+All ngrok-operator CRDs belong to a single API group `ngrok.com` at version `v1`:
+
+| API Group   | Version | CRDs                                                                             |
+|-------------|---------|----------------------------------------------------------------------------------|
+| `ngrok.com` | `v1`    | AgentEndpoint, CloudEndpoint, KubernetesOperator, TrafficPolicy, Domain, IPPolicy, BoundEndpoint |
+
+### Migration from Previous Versions
+
+Previous releases used three separate API groups (`ngrok.k8s.ngrok.com`, `ingress.k8s.ngrok.com`, `bindings.k8s.ngrok.com`) at version `v1alpha1`. A conversion webhook must be provided to handle conversion from the previous group/version combinations to `ngrok.com/v1`.
```
alex-bezek: I don't think we need to talk about the past in our specs

specs/crds/common.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Field Naming
+
+All JSON field names use **camelCase** (e.g., `trafficPolicy`, `resolvesTo`, `poolingEnabled`). No other conventions (snake_case, PascalCase) are permitted.
+
+## Serialization (`omitempty`)
+
+All fields must follow the Kubernetes API conventions for `omitempty`:
+
+- **Optional fields** use `omitempty`. Pointer types, slices, and maps must always include `omitempty` to avoid serializing `null` or zero values.
```
alex-bezek, line 25: we have like 1 scenario where this isn't true right? the domain status part for agent endpoints i think?

specs/crds/common.md
```diff
@@ -0,0 +1,98 @@
...
+- **Required fields** do not use `omitempty`. The field is always present in the serialized form.
+- **Bools** that distinguish between unset and `false` must be pointer types with `omitempty`.
+
+## Status Conditions
+
+All CRDs use `[]metav1.Condition` for status reporting with the following constraints:
```
alex-bezek: technically traffic policy won't have them. we did talk about adding a ready condition just for consistency but it seemed silly since it doesn't have a status and only makes sense in the future when we have a traffic policy api

specs/crds/common.md
```diff
@@ -0,0 +1,98 @@
...
+
+The common condition type is `Ready`, which summarizes the overall health of the resource. Individual controllers may define additional condition types specific to their resource.
+
+Condition type constants must be defined in the API package, not in internal controller code. The API package is the public contract — condition types are part of that contract since users depend on them for `kubectl wait`, health checks, and GitOps tooling.
+
+The API package must only contain API types and constants. Internal implementation details (controller interfaces, reconciliation helpers, etc.) must not be defined in the API package.
+
+## Finalizers
+
+The operator adds the finalizer `ngrok.com/finalizer` to resources it manages. This ensures that:
```
alex-bezek, line 48: specifically resources with ngrok api resources as well as agentendpoints, but isn't used for traffic policy technically

specs/crds/common.md
```diff
@@ -0,0 +1,98 @@
...
+
+1. The operator can clean up remote ngrok API resources before the Kubernetes resource is deleted.
+2. During drain, finalizers are removed to allow Kubernetes garbage collection.
+
+## Owner References
+
+Controllers set owner references on child resources they create. For example, the Service LoadBalancer controller sets an owner reference on created AgentEndpoint/CloudEndpoint resources pointing back to the parent Service.
```
alex-bezek, line 55: maybe caveat here how they are only for the same namespace. unless you think thats common knowledge in k8s and thus the models have it in their training data and know this

specs/crds/common.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Default Field Values
+
+Most CRDs that correspond to ngrok API resources include:
+
+| Field         | Default Value                         |
+|---------------|---------------------------------------|
+| `description` | `"Created by the ngrok-operator"`     |
+| `metadata`    | `{"owned-by": "ngrok-operator"}`      |
```
alex-bezek, line 68: TBD if we include the ticket to do this, but if we do, the metadata would include the operator id as well in theory. but we can hold on this till we make a decision there

specs/controllers/trafficpolicy.md
```diff
@@ -0,0 +1,37 @@
+# TrafficPolicy Controller
+
+## Executive Summary
```
alex-bezek: nit: i don't think these need executive summaries lol. we can have a small abstract/tldr/summary if you want

specs/controllers/kubernetesoperator.md
```diff
@@ -0,0 +1,55 @@
...
+| Resource              | Relation   | Predicate                                    |
+|-----------------------|------------|----------------------------------------------|
+| `KubernetesOperator`  | Primary    | Name+namespace predicate (only its own CR) + AnnotationChanged or GenerationChanged |
+
+The controller only reconciles the specific KubernetesOperator CR matching the operator's Helm release name and namespace.
+
+## Reconciliation Flow
+
+1. Generate or fetch a self-signed TLS certificate for mTLS (bindings feature).
+2. Find or create the KubernetesOperator remote resource in the ngrok API.
```
alex-bezek, line 18: might be good to add more details on how it does this using the default namespace id and why that is good for prod but how it leaks api resources in local clusters if you just recreate them

specs/controllers/kubernetesoperator.md
```diff
@@ -0,0 +1,55 @@
...
+3. Update the remote resource with feature configuration (`enabledFeatures`, `binding`, `deployment`).
+4. Store the bindings ingress endpoint in status.
```
alex-bezek, line 20: doesn't it do a bunch of stuff with a cert that would be good to add?

specs/controllers/kubernetesoperator.md
```diff
@@ -0,0 +1,55 @@
+# KubernetesOperator Controller
```
alex-bezek, line 1: Meta thought for these controllers: should they state what rbac they need in here, or document rbac concerns in a common rbac file?

specs/features/ingress.md
alex-bezek: FWIW though the ingress controller doc mentions it

specs/design-decisions.md
```diff
@@ -0,0 +1,16 @@
+# Design Decisions
+
+This document captures architectural trade-offs that have been weighed and decided. These are not part of the spec itself, but provide context for why the spec is shaped the way it is. Future contributors should understand these decisions before proposing changes that revisit them.
```
alex-bezek, line 3: I wonder if we should also have an architecture section or something that isn't necessarily tradeoffs. something that high level explains how we have 3 different types of pods. and maybe a dedicated file to talking about the driver? i don't see anything for the driver

specs/controllers/ingress.md
```diff
@@ -0,0 +1,50 @@
...
+4. Add finalizer.
+5. Update the Ingress in the Driver store.
+6. Call `Driver.Sync()` to materialize endpoints.
+
+## Driver Pattern
+
+The Ingress controller uses a Driver that collects state from multiple Ingress resources and produces the desired set of ngrok endpoints:
+
+1. Each Ingress is stored in the driver's internal state.
+2. `Driver.Sync()` considers all stored Ingresses, Services, and Domains to generate the correct set of `AgentEndpoint` and/or `CloudEndpoint` resources.
```
alex-bezek, line 31: seems like we should have a whole document on the driver and store?

specs/controllers/domain.md
```diff
@@ -0,0 +1,56 @@
...
+| Error Code | Description                | Behavior                     |
+|------------|----------------------------|------------------------------|
+| 446        | Domain attached to edge    | Retryable (exponential backoff) |
+| 511        | Dangling CNAME             | Retryable (exponential backoff) |
+| Default    |                            | Via `CtrlResultForErr`       |
+
+## Special Cases
+
+- **Internal domains**: Domains with URLs ending in `.internal` are not managed in the ngrok API. The controller removes the finalizer and takes no further action.
+- **DNS provisioning**: Custom domains may take time for DNS to propagate. The exponential backoff rate limiter handles retries during this period.
```
alex-bezek: might also be worth stating that if dns is never setup for a custom domain, it will requeue forever and show not ready

specs/controllers/domain.md
```diff
@@ -0,0 +1,56 @@
+# Domain Controller
+
+## Executive Summary
```
alex-bezek: should probably talk about the reclaim policy in here

specs/controllers/common.md
```diff
@@ -0,0 +1,73 @@
+# Controllers — Common Behavior
+
+## Base Controller Pattern
+
+All ngrok-operator controllers use `BaseController[T]`, a generic base type that implements the standard reconciliation pattern.
```
alex-bezek: most

specs/controllers/cloudendpoint.md
```diff
@@ -0,0 +1,48 @@
...
+|-----------------------|------------|----------------------------------------------|
+| `CloudEndpoint`       | Primary    | AnnotationChanged or GenerationChanged       |
+| `TrafficPolicy`  | Secondary  | Indexed by `spec.trafficPolicyName`; DELETE events filtered |
+| `Domain`              | Owned      | All events                                   |
+
+## Reconciliation Flow
+
+1. Ensure the associated Domain exists via `DomainManager.EnsureDomainExists()`.
+2. Fetch the traffic policy (inline or by name).
+3. Create or update the cloud endpoint via the ngrok API.
```
alex-bezek: might be specific that we do this regardless of if the domain is ready since a not ready domain because the cert is bad is still usable

jonstacks: We should update this to specify that we create/update the cloud endpoint via the ngrok API regardless of if the domain is ready or not. We also want the base controller behavior to only update a ngrok API resource if has a material change in the resource(Don't do updates if it would result in no change to the resource (terraform semantics))

specs/controllers/common.md
jonstacks: Good catch. Will update with some or most.

specs/controllers/domain.md
jonstacks: Good catch. Will update with details about how that works.

specs/controllers/domain.md
alex-bezek: Chatted, we shouldn't focus specifically on "if dns is not setup" but rather just document the requeue behavior more generally in that we will requeue continually as its not ready with exponential backoff

specs/controllers/ingress.md
jonstacks: We talked synchronously, and are going to leave this out since its an internal implementation detail and not a user facing concern.

specs/controllers/kubernetesoperator.md
jonstacks: Seems reasonable to maintain this and list why each controller needs that access.

alex-bezek: For a bit more detail, I think the rbac should cover EVERYTHING but in the controller files it can obviously assume it needs some rbac for the crds, but should focus on calling out additional things it needs like secret access or pods

specs/controllers/kubernetesoperator.md
jonstacks: I agree. I believe the logic is that we use the `uid` of the default namespace to determine if its the same cluster or not. Since that does impact how it reconcile, that makes sense. TODO: Call out the decision logic the k8s op reconciler uses here

specs/controllers/kubernetesoperator.md
jonstacks: Good call out.

TODO: We should make sure it calls out the TLS secret that it creates or fetches and how that process works.

specs/controllers/agentendpoint.md
```diff
@@ -0,0 +1,60 @@
+# AgentEndpoint Controller
```
jonstacks, line 1: TODO: We recently added the ability to do Zero Knowledge(ZK) TLS similar to the ngrok agent where an AgentEndpoint can specify a TLS termination to happen at the agent, including mTLS. We should update this with that change

specs/controllers/trafficpolicy.md
jonstacks: TODO: Just call these sections summaries

specs/crds/common.md
jonstacks: My thought was we wanted a way to automatically migrate these for our users and wanted to call this out in the 1.0.0 spec as desired behavior. Are you thinking we'd leave that up to the user to do a manual migration?

specs/README.md
jonstacks: TODO: remove the `ngrok.com` here

specs/README.md
jonstacks: Resolving as it was called out good for now. I'm leaning towards we'll have to keep helm specs around as its an API of its own, but we'll likely adjust and course correct as needed.

specs/README.md
jonstacks: I'm not keen on calling out the components in the spec for "how" it should work. Feels like an implementation detail IMO. Wdyt?

specs/design-decisions.md
jonstacks: The driver is an implementation detail IMO and we could easily replace it with something else while still complying with what we decide on for the spec. Seems like something that might make it into architecture docs and not necessarily the spec?

specs/authentication.md
jonstacks: You're correct that it was implemented for marketplace installs. If we ever moved to an operator model, where agents are CRDs, like Datadog's, we could largely drop this.

I think to resolve this, we just remove that is for demonstration purposes only from the spec. The reason doesn't much matter for the spec. Just that when setting `oneClickDemoMode` that it doesn't fail for lack of credentials, becomes ready, and waits until credentials are supplied.

specs/annotations.md
jonstacks: From that ticket, we are still in the process of deciding on `ngrok.com/traffic-policy` supporting a named ref and inline traffic policy vs `ngrok.com/traffic-policy-inline`. We probably either need to make a decision on that or loop in product to help unblock

specs/annotations.md
jonstacks: Maybe I'm thinking about the purpose of the spec for an agent incorrectly. I'm thinking it might be best to just describe what the software should do and not the why or the how(maybe best reserved for docs?). I think some examples here might be helpful though to illustrate *what* should be done under different scenarios.

specs/rbac/README.md
jonstacks: TODO: merge the requirements.md file into this one

specs/rbac/README.md
jonstacks: That seems like migration instructions and not something to call out in the spec? I like the idea, but maybe we add it to the AGENTS.md when we start the 1.0 migration efforts?

specs/rbac/aggregation.md
jonstacks: TODO: update with the new value that is in the main branch.

jonstacks: Are you saying that we should support annotations on each of the viewer and editor roles so that they can be aggregated with `aggregationRule` on a custom ClusterRole?

specs/helm/operator.md
jonstacks: I'm fine if we want to move this up. I'm indifferent. Just noting we'll have to update all of the one-click install marketplaces if/when we change it

specs/helm/operator.md
jonstacks: TODO: This feels out of place in the spec. Remove it

specs/helm/features.md
jonstacks: How strongly do you feel it should be moved? IMO, it feels unrelated to how the ngrok-operator works with with ngrok API resources.

specs/helm/common.md
jonstacks: TODO: specify that global section should contain things that are common to all 3 components

specs/helm/common.md
jonstacks: Can you dive into what makes it feel separate? To me it feels like its specifying the default value on the helm chart values(public API). Do you think it should be worked differently or just called out below in the top level structure better?

specs/features/traffic-policy.md
jonstacks: It seems like the event might be more general and we are just using to for deprecated phases today, but that it could also be emitted and should be reserved for general deprecations. Do we want to scope this tighter to only phases? If so, I'd think we'd need to add a different Event in the future which seems annoying.

specs/features/traffic-policy.md
jonstacks: TODO: CloudEndpoint's `spec.trafficPolicy` will match the AgentEndpoints structure.

specs/features/traffic-policy.md
jonstacks: My initial hunch is leaning towards updating the docs based on the specs. My worry is that while the docs are correct for the current version, they are not aligned with the spec for 1.0 and will cause internal inconsistencies.

specs/features/traffic-policy.md
jonstacks: TODO: specify that forwarding from Cloud -> Agent Endpoint happens via manipulation of the supplied traffic policy.

specs/features/namespace-watching.md
jonstacks: TODO: Remove this line.

specs/features/namespace-watching.md
jonstacks: Open Question: Should this be an error condition?

specs/features/ingress.md
jonstacks: IIRC, we talked about the driver and removing it from the spec as an internal implementation detail.

specs/features/ingress.md
jonstacks: TODO: Consider linking to K8s Ingress Status Spec.

We don't own this and have to fill it out to play nicely with other tools like external-dns, etc.

specs/features/high-availability.md
jonstacks: TODO: Add spec to global helm values to support global podDisruptionBudget. Also add to the spec for agent and bindings as well.

specs/features/high-availability.md
jonstacks: I think we want to do some more exploration work first, before finalizing this. The barrier to exploring different options is pretty low now.

specs/features/high-availability.md
jonstacks: TODO: Document the recommendation for Agent & Bindings Forwarder in prod. This will involve specifying cost considerations as well.

specs/features/gateway-api.md
jonstacks: I'll have to ask an agent and get back to you on it.

specs/features/gateway-api.md
jonstacks: I think it slipped in because of how Ingress works. TODO: Remove this and document that a gateway class is required to be created by a user.

specs/features/gateway-api.md
jonstacks: TODO: Add to the spec that Gateway API is installed separately. Might also be worth defining in the spec which versions of the Gateway API we intend to support.

specs/design-decisions.md
alex-bezek: I think you are right. In my head an end goal is that we have enough context here for AI agents to work on features without breaking existing things, so a few comments are calling out concepts that are important for an AI or person to understand when changing the code. But I think you are right in that we could have separate Architecture docs or docs about specific implementations like the driver without those being specs and we can fill those in later

specs/crds/common.md
alex-bezek: we pulled this out of 1.0 so we can disregard this

specs/crds/common.md
alex-bezek: Thinking about this again, it doesn't seem like something we need to document. we could go down rabbit holes if we try to document k8s things

specs/crds/common.md
alex-bezek: no, i agree we want to automate as much as possible and give a specific migration guide. That just felt like different documentation than the "specs". The specs I think should stand the test of time and not be about specific migrations or changes from 1 to another, instead they just document the desired state

specs/features/gateway-api.md
alex-bezek: I think i only see them on the gateway itself right now which makes the most sense. Trying to support per route seems like a merge hell nightmare https://github.com/ngrok/ngrok-operator/blob/v1-spec/pkg/managerdriver/translate_gatewayapi.go#L174-L186

specs/features/namespace-watching.md
alex-bezek: You mean like a status condition on the k8s operator crd? or even the operator not going ready? potentially? IMO if we are going to put effort into it, it might be better suited to just make it work for real

specs/features/traffic-policy.md
alex-bezek: That makes sense to me, seems fine as is

specs/features/traffic-policy.md
alex-bezek: i meant linking to non-operator ngrok docs like traffic policy. probably not something to go wild with though or we risk links getting stale and broken

specs/helm/features.md
alex-bezek: hmm, if i look at the other features I also see config values that are coupled to that feature and only affect the internal k8s setup, not the ngrok api calls like the ingress class controller name, bindings service labels/annotations, gateway reference grants etc. 
I was thinking whether the operator cleans up resources, and how it does that, are a "feature". The drain policy is another configuration of this cleanup. It seems like in this document the split of whats a feature vs config is based on if its a k8s setting vs another type of config? I think I was envisioning a slightly different split where we moved away from these "features" sections and instead the bindings helm section would contain an enabled field, its k8s config options, and a config section that takes all these bindings feature configs. 

we should probably talk about this synchronously or just table the helm values refactor as do that as its own tech design as there are probably a lot of ways to break that down and options to talk through

specs/helm/operator.md
alex-bezek: I'm a bit torn i guess. If it was also a config option for another component like the agent or bindings, then I would say definitely hoist it up so you can only set it in 1 spot 1 way. But i believe this is only used by the api manager pod. It does feel like something that we might end up using in other components, and fundamentally is a config that affects the whole install so I think it makes sense to pull it up or put it as a feature. 

And yeah, we can do it passively in 1 release, but before we completely rip things out we may have to go update all of those. It wouldn't be terrible though to leave a shim in place for this specifically

specs/helm/common.md
alex-bezek: yeah, it was just that everything else is lists of helm values and this just felt out of place calling out that there are 2 charts and then doesn't say which one we are describing below. Its probably fine though

specs/rbac/aggregation.md
alex-bezek: eventually, probably if anyone asks for it or wants it. But my question was about if we should document that here. shortcomings, things we haven't done yet specifically, etc. However, I think it was just on my mind at the time, and we could go crazy trying to document everything we don't do...

specs/rbac/README.md
alex-bezek: I don't think its migration instructions for the user, I meant the appearance of the word tunnel in the code, docs, and these specs. I think you are right though, not something to include in the specs. Might be a neat github agentic workflow to read the code each week and look for old terms like edges, tunnels, kubernetes-ingress, etc and clean them up

specs/README.md
alex-bezek: yep, I think from reading through all this that we just want a separate set of internal/architecture docs for specific things

specs/annotations.md
alex-bezek: I agree with not putting the "how" as that should be able to change. I could see value in the "whys" being supplied though, especially when the time comes when we have to make a change to the specs. I think in this case its just that we want to give users the flexibility to define their endpoints how they want and the default is set to minimize active endpoints. Examples though of how it works in different scenarios though would be super valuable, i'm not even sure I know all the ways it does things dynamically (for the uninstall chainsaw tests i ended up just making everything verbose since they worked different in different scenarios)

specs/features/bindings.md
alex-bezek: on by default makes sense to me since I don't think you need an account flag or anything and doesnt require additional setup

specs/annotations.md
jonstacks: Yeah, after talking about it, I think we should at least have a doc that explains the mapping strategy, gives plenty of examples(good for tests too), and probably at that point also the why.

specs/helm/common.md
jonstacks: I think we'll just call it out below like you mentioned so it doesn't feel so much like an aside.

specs/helm/operator.md
jonstacks: TODO: Spike on this

specs/helm/features.md
jonstacks: TODO: We have common enough ground that it feels weird, so its probably something we should look into to compare.

specs/features/traffic-policy.md
jonstacks: I missed that it was applying to non ngrok-operator concepts. I think that makes sense to embed those. If link get stale or broken, we can have checks for those.

TODO: Add links for traffic policy and probably ngrok API resources.

specs/features/traffic-policy.md
jonstacks: I think this is just something where we don't want to use "inline" here.

TODO: `reference` is probably a better word here. Also we're not going to pick this up for 1.0.

specs/annotations.md
jonstacks: After talking sync, I think we are going to punt on this for now and not pick up that ticket right now

specs/features/traffic-policy.md
jonstacks: TODO: already discussed and we'll have a separate doc section for it, link to it from here

specs/features/namespace-watching.md
jonstacks: TODO: Just make sure to call this out as not supported and be more explicit

specs/features/high-availability.md
jonstacks: Technicallly Correct

TODO: Also, I don't know how much this particular example adds versus talking about leader election in general and just linking out to kubebuilder's/controller-runtime docs

specs/features/gateway-api.md
jonstacks: Yeah. I think just specifying it on the gateway for now makes the most sense to not have to deal with that nightmare for now.

specs/crds/common.md
jonstacks: TODO: Talked sync, probably makes sense to call out the migration in its own file so we can delete it later. Ex: 1.2 drops support for old types. Also need to drop the wording around webhook conversion as we are moving groups.

specs/crds/common.md
jonstacks: TODO: Double check on this and add a carveout for AgentEndpoints if necessary

specs/crds/common.md
jonstacks: TODO: Add carve out like "All CRDS use it, unless otherwise noted" -> then note it on the traffic policy spec.

specs/features/bindings.md
jonstacks: There is a bit more setup for this. The default endpoint selector is `true` which will then attempt to project every service into your k8s cluster, which may not be what you want. It seems likely to me that it will fail to create namespaces for many of the bound endpoints if they don't exist.

@iambmelt, do you have any thoughts on this?

iambmelt: Reading over this, I'm inclined to think we want to keep bindings opt-in at v1. THought process being, if I have multiple clusters (e.g., dev/stage/prod), every instance could end up racing to project every binding (which doesn't seem like what users would expect/want?). It also strikes me that auto-creating namespaces feels like a privileged activity that folks may be uneasy about allowing through this mechanism

My $0.02 is seems like `enabled: false` is the safe default, I'd want to see `enabled: true` to at least emit a warning if paired with the `true` selector

alex-bezek (approved):

PR #806, opened 2026-04-20
fix: skip redundant ngrok API updates for CloudEndpoint

jonstacks:
## What

The update path unconditionally called the ngrok API to update the endpoint on every reconciliation, even when nothing had changed. This was most visible when a domain was not ready: the controller requeues every 10 seconds, generating unnecessary API write calls for the lifetime of the unready domain.

## How

Fetch the current endpoint via Get before issuing an Update and compare the API state against the desired spec. The Update call is now skipped entirely when the endpoint already matches. This also moves the not-found detection from after Update to after Get, catching missing endpoints earlier.

## Breaking Changes

No

### DIFF
--- a/internal/controller/ngrok/cloudendpoint_controller.go
+++ b/internal/controller/ngrok/cloudendpoint_controller.go
@@ -28,6 +28,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
@@ -232,6 +233,28 @@ func (r *CloudEndpointReconciler) update(ctx context.Context, clep *ngrokv1alpha
 		return r.updateStatus(ctx, clep, nil, domainResult, err)
 	}
 
+	// Fetch current endpoint state from the ngrok API so we can compare
+	// before issuing an update. This avoids redundant API writes on every
+	// requeue cycle (e.g. while waiting for a domain to become ready).
+	currentEndpoint, err := r.NgrokClientset.Endpoints().Get(ctx, clep.Status.ID)
+	if ngrok.IsNotFound(err) {
+		r.Recorder.Eventf(clep, nil, v1.EventTypeWarning, "EndpointNotFound", "Reconcile", fmt.Sprintf("Failed to find endpoint %s by ID. Creating a new one", clep.Status.ID))
+		clep.Status.ID = ""
+		if err := r.controller.ReconcileStatus(ctx, clep, nil); err != nil {
+			return err
+		}
+		return r.create(ctx, clep)
+	}
+	if err != nil {
+		return r.updateStatus(ctx, clep, nil, domainResult, err)
+	}
+
+	// Skip the API update if nothing has changed
+	if !endpointNeedsUpdate(currentEndpoint, clep.Spec, policy) {
+		setCloudEndpointCreatedCondition(clep, true, ReasonCloudEndpointCreated, "CloudEndpoint updated successfully")
+		return r.updateStatus(ctx, clep, currentEndpoint, domainResult, nil)
+	}
+
 	updateParams := &ngrok.EndpointUpdate{
 		ID:             clep.Status.ID,
 		Url:            &clep.Spec.URL,
@@ -243,15 +266,6 @@ func (r *CloudEndpointReconciler) update(ctx context.Context, clep *ngrokv1alpha
 	}
 
 	ngrokClep, err := r.NgrokClientset.Endpoints().Update(ctx, updateParams)
-	if ngrok.IsNotFound(err) {
-		// Couldn't find endpoint by ID to update, so blank it out and create a new one
-		r.Recorder.Eventf(clep, nil, v1.EventTypeWarning, "EndpointNotFound", "Reconcile", fmt.Sprintf("Failed to update endpoint %s by ID because it was not found. Creating a new one", clep.Status.ID))
-		clep.Status.ID = ""
-		if err := r.controller.ReconcileStatus(ctx, clep, nil); err != nil {
-			return err
-		}
-		return r.create(ctx, clep)
-	}
 	if err != nil {
 		setCloudEndpointCreatedCondition(clep, false, ReasonCloudEndpointCreationFailed, fmt.Sprintf("Failed to update cloud endpoint: %v", err))
 		return r.updateStatus(ctx, clep, nil, domainResult, err)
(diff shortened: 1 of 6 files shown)


internal/controller/ngrok/cloudendpoint_controller.go
```diff
@@ -232,6 +233,28 @@ func (r *CloudEndpointReconciler) update(ctx context.Context, clep *ngrokv1alpha
...
 	}
 
+	// Fetch current endpoint state from the ngrok API so we can compare
+	// before issuing an update. This avoids redundant API writes on every
+	// requeue cycle (e.g. while waiting for a domain to become ready).
+	currentEndpoint, err := r.NgrokClientset.Endpoints().Get(ctx, clep.Status.ID)
+	if ngrok.IsNotFound(err) {
+		r.Recorder.Eventf(clep, nil, v1.EventTypeWarning, "EndpointNotFound", "Reconcile", fmt.Sprintf("Failed to find endpoint %s by ID. Creating a new one", clep.Status.ID))
+		clep.Status.ID = ""
+		if err := r.controller.ReconcileStatus(ctx, clep, nil); err != nil {
```
alex-bezek, line 243: A pattern I was trying to do is most controllers have an `updateStatus` function that handles setting all the status fields and deriving the conditions from that status and setting those too, and then it always calls the .controller.ReconcileStatus. In this case you could pass it the nil currentEndpoint and it would set the status.ID to "", set conditions, and call ReconcileStatus. 

That being said, setting the conditions may not be that important since we immediately call `.create` and don't rely on a re-reconcile. That create does the same status setting so I don't think you get much out of the conditions other than more updates. 
So that means all it would do differently is dry up setting the status id which isn't really worth much. 

Fully up to you, just thought i'd drop the thought.

internal/controller/ngrok/cloudendpoint_controller.go
```diff
@@ -232,6 +233,28 @@ func (r *CloudEndpointReconciler) update(ctx context.Context, clep *ngrokv1alpha
...
+		}
+		return r.create(ctx, clep)
+	}
+	if err != nil {
+		return r.updateStatus(ctx, clep, nil, domainResult, err)
+	}
+
+	// Skip the API update if nothing has changed
+	if !endpointNeedsUpdate(currentEndpoint, clep.Spec, policy) {
+		setCloudEndpointCreatedCondition(clep, true, ReasonCloudEndpointCreated, "CloudEndpoint updated successfully")
```
alex-bezek, line 254: is this necessary since updateStatus should do the same thing? 
https://github.com/ngrok/ngrok-operator/blob/b91752ab6f003691b0579d00905ea708eeb7785a/internal/controller/ngrok/cloudendpoint_controller.go#L291-L292

internal/controller/ngrok/cloudendpoint_controller_unit_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
+	"testing"
+
+	"github.com/ngrok/ngrok-api-go/v7"
+	ngrokv1alpha1 "github.com/ngrok/ngrok-operator/api/ngrok/v1alpha1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestEndpointNeedsUpdate(t *testing.T) {
```
alex-bezek, line 11: This makes sense to me that its done separately. Is it just because to test this many edge cases via the full reconcile would be way more verbose and this strikes a good balance?

internal/controller/ngrok/cloudendpoint_controller.go
```diff
@@ -344,6 +358,31 @@ func (r *CloudEndpointReconciler) findCloudEndpointsForDomain(ctx context.Contex
 	return requests
 }
 
+// endpointNeedsUpdate compares the current endpoint state from the ngrok API
+// against the desired state derived from the CloudEndpoint spec and the resolved
+// traffic policy. Returns true if an API update call is necessary.
+func endpointNeedsUpdate(current *ngrok.Endpoint, spec ngrokv1alpha1.CloudEndpointSpec, policy string) bool {
```
alex-bezek, line 364: For some reason I thought we had some pattern of putting a `.Equals` method on a crd type that could accept another type it correlated to and do all these checks so the checks for the fields lived close to where the fields would get added. 

I can't find this pattern though in this project or our internal one so I might just be misremembering. Curious your thoughts on the idea though. I wonder if we just talked about it in the past but chose not to do it as it would make our crd package types rely on the ngrok package and might make circular deps

alex-bezek (approved):
just some minor questions/nits about status. Looks good

internal/controller/ngrok/cloudendpoint_controller.go
jonstacks: I believe we did have it at one point on a few things. Don't know if we standardized it across our API though:
* https://github.com/ngrok/ngrok-operator/blob/22cc736ec3108b995a40556d17ba84956e334fb8/api/ingress/v1alpha1/domain_types.go#L111-L119
* https://github.com/ngrok/ngrok-operator/blob/22cc736ec3108b995a40556d17ba84956e334fb8/api/ingress/v1alpha1/httpsedge_types.go#L157-L179

Might be worth considering for 1.0.0 though.

internal/controller/ngrok/cloudendpoint_controller_unit_test.go
jonstacks: I went back and forth with including a test for this directly. I've generally been preferring that we use the suite tests to test the behavior of our controllers from the outside so we don't have a bunch of tests that churn when we do refactors. This was also a weird thing to test when we had methods on the Reconcilers and we're calling them during the suite tests. It just felt wrong to directly call a Reconciler method instead of verifying the results of the reconcile.

As you mentioned, I think there are a lot of paths through this function though that it would be hard to get good coverage via envtest. Since it is a pure function, I decided a unittest was pretty simple.

internal/controller/ngrok/cloudendpoint_controller.go
jonstacks: Looks like `updateStatus` only sets the overall `ReadyCondition` based on other conditions. I assumed we still needed to set those other conditions.

internal/controller/ngrok/cloudendpoint_controller.go
jonstacks: Similar to the below comment, I don't think `updateStatus` does that today. From what I can tell, it only sets the overall `ReadyCondition`. I think I'd need to refactor it to have it derive all of the conditions.

internal/controller/ngrok/cloudendpoint_controller.go
jonstacks: Talked offline, it only sets the `ReadyCondition`. Might be good for a future refactor.

PR #809, opened 2026-05-08
Release operator helm chart 0.23.0-rc1

alex-bezek:
## What
Releasing as an RC for testing

## How
Trying out new ai based relese process https://github.com/ngrok/ngrok-operator/compare/main...alex/claude-release-skill

## Breaking Changes

  - URI → URL rename (#779): URI field renamed to URL across CRDs (AgentEndpoint, CloudEndpoint, etc.) for consistency
  - Bindings port allocator (#801): panic() calls replaced with error returns — callers that relied on panic recovery behavior may need updates
  - RBAC overhaul (#804): All RBAC is now hand-managed in Helm templates. Controller-gen kubebuilder:rbac markers removed. ClusterRoles restructured
  per-component (api-manager, agent-manager, bindings-forwarder). Existing role bindings will be replaced on upgrade.


jonstacks (approved):
**Non-blocking** We might want to aggregate the breaking changes on the PR description for when we cut the actual release. Since this is an RC, less important IMO. We'll want it for the actual release though

PR #814, opened 2026-05-20
feat(agent): Support agent TLS termination

jonstacks:
## What

The ngrok agent supports [terminating TLS at the agent](https://ngrok.com/docs/agent/agent-tls-termination), allowing users to manage their own PKI infrastructure and using ngrok as the transport(ngrok reads the SNI and routes appropriately). The ngrok-operator does not yet support this today. This PR adds support for AgentEndpoints to specify their own TLS configuration(server certificate and key) or mTLS configuration and offload this to the agent.

## How
* Adds new fields to AgentEndpoint CRD to specify TLSTermination and mTLS configuration
* passes those to the ngrok-go v2 SDK

## Breaking Changes
No, this is a net new capability

## Summary by CodeRabbit

* **New Features**
  * Agent-side TLS termination for AgentEndpoint: supply a server certificate and optional mutual‑TLS (modes: require | request). CRD requires referenced certificate and CA names and enforces that endpoints using TLS termination use "tls://".

* **Tests**
  * Added tests using self-signed TLS secrets covering server-certificate resolution, mTLS modes, and related error conditions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ngrok/ngrok-operator/pull/814?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


internal/controller/agent/agent_endpoint_controller.go
```diff
@@ -448,6 +486,95 @@ func (r *AgentEndpointReconciler) getClientCerts(ctx context.Context, aep *ngrok
...
+
+// clientAuthForMode maps the CRD mTLS mode to tls.ClientAuthType. Defaults to
+// RequireAndVerifyClientCert when the mode is empty (matches the CRD default).
+func clientAuthForMode(mode ngrokv1alpha1.EndpointMutualTLSMode) tls.ClientAuthType {
+	switch mode {
+	case ngrokv1alpha1.EndpointMutualTLSModeRequest:
+		return tls.VerifyClientCertIfGiven
+	default:
+		return tls.RequireAndVerifyClientCert
+	}
```
alex-bezek, line 574: is there any risk that we would add another type here in the future and should use an explicit case with the default or logging?

internal/controller/agent/agent_endpoint_controller.go
```diff
@@ -448,6 +486,95 @@ func (r *AgentEndpointReconciler) getClientCerts(ctx context.Context, aep *ngrok
...
+
+// getServerCert fetches the agent's server certificate (tls.crt + tls.key) for
+// TLS termination from the referenced kubernetes.io/tls Secret.
+func (r *AgentEndpointReconciler) getServerCert(ctx context.Context, aep *ngrokv1alpha1.AgentEndpoint) (*tls.Certificate, error) {
+	ref := aep.Spec.TLSTermination.ServerCertificateRef
+	key := ref.ToClientObjectKey(aep.Namespace)
+
+	secret := &v1.Secret{}
+	if err := r.Client.Get(ctx, key, secret); err != nil {
+		r.Recorder.Eventf(secret, nil, v1.EventTypeWarning, "SecretNotFound", "Reconcile", fmt.Sprintf("Failed to find Secret %s for tlsTermination.serverCertificateRef", ref.Name))
```
alex-bezek: if we failed to get the secret, this event won't be tied to it right? should we record the event an the AEP or just log instead?

alex-bezek (approved):

internal/controller/agent/agent_endpoint_controller.go
jonstacks: If you opt-in to mutual TLS for the AgentEndpoint, this seems like a reasonable default to me. Presumably, you are trying lock down access with a client certificate. The full list of consts we can pass are in https://pkg.go.dev/crypto/tls#ClientAuthType.

If anything, I think there is a likely chance of someone saying they want `RequestClientCert` or `RequireAnyClientCert` and us adding those cases into this switch statement.

Are you saying that you want more logging around which case was selected?

internal/controller/agent/agent_endpoint_controller.go
jonstacks: That is a good point, I think we do probably want to record this as an event on the AEP. I'll dig into it

jonstacks:
@coderabbitai review

internal/controller/agent/agent_endpoint_controller.go
jonstacks: Fixed in https://github.com/ngrok/ngrok-operator/pull/814/commits/ef06e48c5fcd299b99800362160df3278ff655ac to add more logging when it becomes the default

internal/controller/agent/agent_endpoint_controller.go
jonstacks: Fixed in https://github.com/ngrok/ngrok-operator/pull/814/commits/ef06e48c5fcd299b99800362160df3278ff655ac

internal/controller/agent/agent_endpoint_controller.go
```diff
@@ -236,9 +266,15 @@ func (r *AgentEndpointReconciler) update(ctx context.Context, endpoint *ngrokv1a
...
+	agentTLS, err := r.getAgentTLSTermination(ctx, endpoint)
+	if err != nil {
+		setEndpointCreatedCondition(endpoint, false, ReasonConfigError, fmt.Sprintf("Failed to get TLS termination config: %v", err))
+		return r.updateStatus(ctx, endpoint, nil, trafficPolicy, domainResult, err)
+	}
+
 	// Create the endpoint
 	tunnelName := r.statusID(endpoint)
-	result, err := r.AgentDriver.CreateAgentEndpoint(ctx, tunnelName, endpoint.Spec, trafficPolicy, clientCerts)
+	result, err := r.AgentDriver.CreateAgentEndpoint(ctx, tunnelName, endpoint.Spec, trafficPolicy, clientCerts, agentTLS)
```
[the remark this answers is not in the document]
jonstacks: We explicitly call out on the CRD that this is not supported. We may decide to follow up on this in a future PR, but for now we aren't going handle this.

PR #821, opened 2026-06-02
Migrate internally written labels and annotations from k8s.ngrok.com …

alex-bezek:
…to ngrok.com

## What
For 1.0, we are changing various things from k8s.ngrok.com to ngrok.com . This PR specifically updates the annotations and labels that it writes to CRDs internally

## How
To support a passive rollout and possible passive rollbacks, these changes take a 3 phase approach:
- we initially read from both and write both. we don't just write the new one as that can make rolling back more difficult. It would be possible in some cases to just write the new one and rollbacks would still work for anything that previously existed and had its old labels. This is difficult though with our driver pattern where we are reconciling things and would force it to match what we want
- future rollouts will then stop reading the old value, stop writing it, and delete it if its present. then a later release will remove this code
- Affected the following spots
  - Labels written on domains, cloud, and agent endpoints to denote them as owned by the driver not the user
  - service computed url
  - bindings labels

## Breaking Changes
Not yet

## Release Notes

* **New Features**
  * Added zero-downtime support for migrating legacy `k8s.ngrok.com/*` label/annotation keys to `ngrok.com/*`, using a staged rollout approach that maintains dual compatibility during upgrades and rollbacks.
* **Documentation**
  * Expanded the migration and rollback guidance (including “passivity shims”) and provided component-specific behavior details.
* **Bug Fixes / Tests**
  * Improved computed-url and label/selector reconciliation to correctly prefer new keys while safely backfilling/cleaning legacy keys, with added unit and integration test coverage.


internal/domain/manager.go
```diff
@@ -210,10 +210,20 @@ func (m *Manager) ensureControllerLabels(ctx context.Context, log logr.Logger, d
...
+	// gets the legacy pair backfilled. The read-side cleanup drops the
+	// legacy-pair check; the early-return then collapses back to
+	//   if hasControllerNameLabel && hasControllerNamespaceLabel { return nil }
+	_, hasLegacyName := l[labels.LegacyControllerName]
+	_, hasLegacyNamespace := l[labels.LegacyControllerNamespace]
+
+	if hasControllerNameLabel && hasControllerNamespaceLabel && hasLegacyName && hasLegacyNamespace {
 		return nil
 	}
+	// LEGACY-PREFIX-MIGRATION: END
```
jonstacks: **Non-Blocking** I know we're in the process of doing a migration, but I think we might be able to clean this(L210-L226) up with:

```suggestion
    if !m.controllerLabels.EnsureLabels(domainObj.DeepCopy()) {
        return nil
    }
```

If we are okay with the extra allocation from DeepCopy, this clone-and-probe strat allows us one less place of having to do the cleanup.

jonstacks (approved):

jonstacks (approved):

PR #823, opened 2026-06-02
setup passive migrations to let cloud endpoints reference traffic pol…

alex-bezek:
…icies the same way as agent endpoints

## What
We are cleaning up our CRDs and making things consistent for a 1.0 release. This change updates the cloud endpoint CRD to support a better Traffic Policy Ref like the agent endpoint has. I first made cloud endpoint a long time ago and just had it reference the traffic policy CRD name via a string. It also added another field to support adding the policy inline as yaml/json. however the agent endpoint used a more formal ref approach and also allowed inlining it via the same attribute.

This updates cloud endpoint to support the same approach, but does it passively

## How
- follow a similar approach as how agent endpoint and cloud endpoints have a shared interface and package to dry things up
- add a temporary type for now that supports the new and old ways for passivity
- dries up the code that handles watching for updates
- handles errors to avoid infinite requeues

## Breaking Changes
Not Yet

## Summary by CodeRabbit

* **New Features**
  * CloudEndpoints now support traffic policies defined inline or referenced by name.
  * Traffic policies can be applied consistently across CloudEndpoints and AgentEndpoints, with same-namespace reference resolution.
* **Bug Fixes**
  * Improved traffic-policy status reporting for invalid, missing, or rejected policies.
  * Prevented repeated reconciliation when the API rejects a traffic policy.
* **Compatibility**
  * Existing traffic-policy fields remain readable during migration, with canonical configuration taking precedence.
  * Deprecated fields now provide migration guidance and warnings where appropriate.
* **Documentation**
  * Added migration, rollback, and traffic-policy configuration guidance.


alex-bezek:
Discussed this with @jonstacks . We initially came up with the idea of allowing the cross namespace traffic policy lookup incase someone wanted to have a namespace filled with re-useable traffic policies. We talked about it though, and while we don't think there are any security issues, it does raise eyebrows and is just a bit weird and nobody asked for it. So we are deciding to remove the cross namespace functionality

alex-bezek:
Note: also ensure the TrafficPolicyApplied condition is made consistent and correct

internal/controller/agent/agent_endpoint_controller.go
```diff
@@ -242,47 +235,51 @@ func (r *AgentEndpointReconciler) update(ctx context.Context, endpoint *ngrokv1a
...
 	}
 
-	// Mark the endpoint as successfully created
+	// Downstream create succeeded — record the policy as applied (true)
+	// and mark the endpoint as created.
+	r.TrafficPolicyManager.MarkApplied(endpoint)
 	setEndpointCreatedCondition(endpoint, true, ReasonEndpointCreated, "Endpoint successfully created")
-	if trafficPolicy != "" {
-		setTrafficPolicyCondition(endpoint, true, "TrafficPolicyApplied", "Traffic policy successfully applied")
-	}
```
jonstacks, line 283: Do we still want this guard for if the traffic Policy is not empty?

jonstacks (approved):

internal/controller/agent/agent_endpoint_controller.go
alex-bezek: the same type of check exists in the `MarkApplied` function, its just inverted. In there it checks if the endpoint has a traffic policy and if not it early returns. Otherwise it sets the condition

PR #842, opened 2026-07-08
Make the secret name used for mtls with the operator and bindings con…

alex-bezek:
…figurable via helm

Resolves K8SOP-278

## What

On startup the operator is responsible for creating its KubernetesOperator CRD. On that CRD is a `TlsSecretName`. the KubernetesOperator controller makes a cert in this secret and the bindings forwarder uses the operator field to find the secret name to lookup and use when connecting to the bindings endpoint. 

Previously this was just hardcoded to `ngrok-operator-default-tls`

## How

Add a helm value to make it configurable. use the full name override as a prefix to avoid secret collisions. 

## Breaking Changes

If you are overriding the release name to be something different than `ngrok-operator` then your "default" value for this secret name will change but should reconcile and be eventually consistent. Users can set the secret name to the old secret value for passivity if they want. 

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the TLS secret used by the Bindings feature via Helm/flags.
  * When not set, the operator now uses a release-scoped default secret name instead of a fixed value.

* **Bug Fixes**
  * Removed the implicit default for the bindings TLS secret field.
  * Enforced validation to reject empty secret names.

* **Documentation**
  * Updated Bindings and authentication docs, plus CRD/reference text, to reflect the new release-based TLS secret naming and validation behavior.

### DIFF
--- a/api/ngrok/v1alpha1/kubernetesoperator_types.go
+++ b/api/ngrok/v1alpha1/kubernetesoperator_types.go
@@ -50,7 +50,7 @@ type KubernetesOperatorBinding struct {
 
 	// TlsSecretName is the name of the k8s secret that contains the TLS private/public keys to use for the ngrok forwarding endpoint
 	// +kubebuilder:validation:Required
-	// +kubebuilder:default="default-tls"
+	// +kubebuilder:validation:MinLength=1
 	TlsSecretName string `json:"tlsSecretName"`
 }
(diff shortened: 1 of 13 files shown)


alex-bezek:
We decided to hold on this as its not required for 1.0 and may change with private dial coming soon

PR #844, opened 2026-07-08
tighten rbac further and update specs

alex-bezek:
Resolves K8SOP-280

## What

From auditing everything and reviewing permissions we found a few spots that were overly permissive that we could clean up. 

Additionally while here, updated the specs a bit

## How

removing permissions we don't have code to use. 

## Breaking Changes

no

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened operator permissions by removing several unnecessary write/delete actions from cluster- and namespace-scoped access.
  * Reduced access for select resources to read-only or limited update paths where appropriate.

* **Documentation**
  * Reworked RBAC docs into clearer, component-specific guides.
  * Simplified the main RBAC overview and updated permission descriptions for operator, agent, and bindings forwarding components.

PR #852, opened 2026-07-15
NgrokTrafficPolicy: replace status.policy with Ready/Valid conditions + printer columns

alex-bezek:
Fixes [K8SOP-288](https://linear.app/ngrok/issue/K8SOP-288/ngroktrafficpolicy-status-remove-statuspolicy-add-conditions-printer)

## What

- **Remove `status.policy`** (breaking): it only mirrored `spec.policy` and nothing ever wrote it; `observedGeneration` on conditions is the correct "what did the controller see" signal.
- **Add `Ready`/`Valid` conditions** reflecting parse/validation of `spec.policy`. Both share the same reason so deprecation warnings surface in the Ready-based printer columns. Reasons: `TrafficPolicyValid`, `TrafficPolicyParseFailed`, `LegacyPolicyFormat`, `EnabledFieldDeprecated`.
- **Printer columns**: Ready, Reason (priority=1), Age.
- Spec `policy` field now documents why it is intentionally schemaless (kept `Type=object`).

## Behavior changes

- Parse failures and legacy-format warnings were Events only — expired after ~1h, invisible in `kubectl get`, unusable with `kubectl wait`/GitOps health checks. They now surface on the resource (Events kept too).
- A malformed policy no longer returns an error (endless requeue); it sets `Ready=False` and waits for a spec change (generation-changed predicate re-triggers).
- `client.IgnoreNotFound` on the initial Get.

## RBAC

api-manager role regains `ngroktrafficpolicies/status` get/patch/update. #844 removed it as a dead permission (correct at the time — the controller never wrote status); this PR adds the first real status write.

## Docs/specs

- `docs/v1-migration-guide.md`: new section on the `status.policy` removal.
- `specs/`: crds/trafficpolicy.md, controllers/trafficpolicy.md, features/traffic-policy.md, rbac/operator.md updated to match.

## Verify

- New table-driven unit test `TestSetTrafficPolicyConditions`: invalid JSON → `Ready=False`/`TrafficPolicyParseFailed`, valid → `Ready=True`, legacy directions → `LegacyPolicyFormat`, `enabled` set → `EnabledFieldDeprecated`.
- `make test` (full suite), `make lint` (0 issues), `make helm-test` (131/131, snapshots updated), `make generate manifests manifest-bundle` regenerated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * TrafficPolicy now reports policy parsing/validation through `Ready` and `Valid` status conditions.
  * Added `Ready`, `Reason`, and `Age` printer columns for TrafficPolicy resources.
* **Bug Fixes**
  * Malformed or deprecated policies now surface outcomes via `status.conditions` (and related warnings) without blocking reconciliation.
* **Documentation**
  * Updated the v1 migration guide and TrafficPolicy/cloud endpoint docs to reflect the `status.conditions` model and schemaless reconcile-time parsing.
* **Chores**
  * Updated CRD schema/behavior for `status.conditions` (max 8), RBAC for status updates, and enhanced condition-copy semantics; expanded controller tests.


internal/controller/ngrok/ngroktrafficpolicy_controller.go
```diff
@@ -68,13 +68,20 @@ func (r *NgrokTrafficPolicyReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	policy := &ngrokv1alpha1.NgrokTrafficPolicy{}
 	if err := r.Get(ctx, req.NamespacedName, policy); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	parsedTrafficPolicy, parseErr := util.NewTrafficPolicyFromJson(policy.Spec.Policy)
+	setTrafficPolicyConditions(policy, parsedTrafficPolicy, parseErr)
+	policy.SetObservedGeneration(policy.Generation)
+	if err := r.Status().Update(ctx, policy); err != nil {
```
jonstacks: This looks like we're updating the status on every reconcile even if nothing has changed. Should we conditionally update only if it would result in a status change?

Also curious on your thoughts about setting this before its handled by the manager driver

internal/controller/ngrok/ngroktrafficpolicy_conditions.go
```diff
@@ -0,0 +1,45 @@
...
+	ReasonTrafficPolicyValid       = "TrafficPolicyValid"
+	ReasonTrafficPolicyParseFailed = "TrafficPolicyParseFailed"
+	ReasonLegacyPolicyFormat       = "LegacyPolicyFormat"
+	ReasonEnabledDeprecated        = "EnabledFieldDeprecated"
+)
+
+// setTrafficPolicyConditions sets the Valid and Ready conditions from the
+// result of parsing spec.policy. Both conditions share the same reason so
+// deprecation warnings surface in the Ready-based printer columns.
+func setTrafficPolicyConditions(tp *ngrokv1alpha1.NgrokTrafficPolicy, parsed util.TrafficPolicy, parseErr error) {
```
jonstacks, line 26: What do you think about having this return a bool for whether or not the conditions were changed? I think we can pipe that through. It might help determine if we need to make an update later or not.

jonstacks (approved):

internal/controller/ngrok/ngroktrafficpolicy_conditions.go
alex-bezek: Seems like it should be pretty doable since `metav1.SetStatusCondition` returns that bool, we just swallow it in our wrapper. If we want to do this though, it seems like it would be better to pull out into its own change we apply across all the controllers since right now none of them do this and we'd have to update the shared helper function.

internal/controller/ngrok/ngroktrafficpolicy_controller.go
alex-bezek: I can diff the status before updating 👍 

As for setting the status before the managerDriver.SyncEndpoints, I think this way makes sense. The status information is just about if its able to parse it as json. One day if we have an api for validating these, it would do that as well. But since these can be used by multiple ingress/crds, I don't think their status has anything to do with how they are used. You could make a standalone trafficpolicy CRD that shows as ready "to be used" was what i was thinking

jonstacks (approved):

PR #872, opened 2026-08-14
R2 write-side cleanup: k8s.ngrok.com -> ngrok.com prefix migration (K8SOP-303)

alex-bezek:
## Summary
Second-release (R2) write-side cleanup for the `k8s.ngrok.com/` -> `ngrok.com/` prefix migration shipped in 0.24 (R1). Per the two/three-release patterns in `docs/developer-guide/passivity-shims.md`, this stops writing the legacy-prefixed key on new/updated objects while keeping dual-read support so already-reconciled objects and rollback stay safe.

- Controller labels (`internal/controller/labels`): `EnsureControllerLabels` now deletes the legacy-prefixed pair instead of dual-writing it. Dual-read (`HasControllerLabels`/selectors) is untouched.
- `ngrok.com/computed-url` Service annotation: write-side now writes only the new key and deletes the legacy one if present.
- BoundEndpoint-owned Service labels/annotations (bindings): dual-write dropped for the labels/annotations set on generated Services.
- Operator finalizer: `AddFinalizer` now writes only the new key and removes the legacy one (three-release finalizer pattern — `HasFinalizer`/`RemoveFinalizer` still bridge both keys, so a rollback to R1 stays safe).
- Helm: `ingress.controllerName` default flipped to `ngrok.com/ingress-controller` (IngressClass rollout-race deferral continues to recognize both values at read time).

## Test plan
- [x] `go build ./...`, `go vet ./...`, `golangci-lint run` all clean
- [x] `go test ./...` green, including updated fixtures/tests across `internal/controller/labels`, `internal/controller/service`, `internal/controller/bindings`, `internal/util`, `internal/domain`, `pkg/managerdriver` (64 golden fixtures updated), `internal/controller/ngrok`, `internal/controller/gateway`
- [x] `helm unittest` green, snapshots regenerated (`helm/ngrok-operator`)
- [x] `tests/chainsaw*` fixtures updated to assert the new-prefix keys where objects are freshly created

K8SOP-303

## Summary by CodeRabbit

- **Changed**
  - Standardized generated resource metadata on the `ngrok.com` label namespace.
  - Existing legacy controller labels are now cleaned up during reconciliation.
  - Finalizers now use `ngrok.com/finalizer`, with legacy finalizers removed.
  - Computed service URL annotations migrate to the canonical key and remove legacy keys.
  - Generated services no longer write legacy ownership labels or endpoint URL annotations.

- **Documentation**
  - Updated migration guidance for legacy metadata cleanup and removable-code markers.

### DIFF
--- a/helm/ngrok-operator/values.yaml
+++ b/helm/ngrok-operator/values.yaml
@@ -258,7 +258,7 @@ credentials:
 ##
 ingress:
   enabled: true # enabled by default
-  controllerName: "k8s.ngrok.com/ingress-controller"
+  controllerName: "ngrok.com/ingress-controller"
   watchNamespace: "" # default all
   ingressClass:
     name: ngrok
--- a/internal/controller/service/controller.go
+++ b/internal/controller/service/controller.go
@@ -381,18 +381,14 @@ func (r *ServiceReconciler) setComputedURLAnnotation(ctx context.Context, svc *c
 	if a == nil {
 		a = make(map[string]string)
 	}
-	// LEGACY-PREFIX-MIGRATION: BEGIN
-	// Dual-write both the new and legacy computed-url keys so a rollback to a
-	// pre-migration operator still sees the value it knows how to read. The
-	// write-side cleanup drops the legacy comparison and the legacy write; the
-	// body collapses back to the simple "skip if equal, otherwise write" form.
-	if a[annotations.ComputedURLAnnotation] == computedURL &&
-		a[annotations.LegacyComputedURLAnnotation] == computedURL {
+	// LEGACY-PREFIX-MIGRATION (read-side cleanup): drop this delete once no
+	// pre-migration-stamped Services remain.
+	_, hadLegacy := a[annotations.LegacyComputedURLAnnotation]
+	if a[annotations.ComputedURLAnnotation] == computedURL && !hadLegacy {
 		return nil
 	}
 	a[annotations.ComputedURLAnnotation] = computedURL
-	a[annotations.LegacyComputedURLAnnotation] = computedURL
-	// LEGACY-PREFIX-MIGRATION: END
+	delete(a, annotations.LegacyComputedURLAnnotation)
 	svc.SetAnnotations(a)
 	return r.Client.Update(ctx, svc)
 }
--- a/internal/domain/manager_test.go
+++ b/internal/domain/manager_test.go
@@ -153,10 +153,7 @@ func assertDomainLabels(t *testing.T, c client.Client, name, namespace string, e
 	t.Helper()
 	var domain ingressv1alpha1.Domain
 	require.NoError(t, c.Get(t.Context(), types.NamespacedName{Name: name, Namespace: namespace}, &domain))
-	domainLabels := domain.GetLabels()
-	for k, v := range expectedLabels {
-		assert.Equal(t, v, domainLabels[k])
-	}
+	assert.Equal(t, expectedLabels, domain.GetLabels())
 }
 
 // Tests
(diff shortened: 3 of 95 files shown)


internal/controller/service/controller.go
```diff
@@ -381,18 +381,14 @@ func (r *ServiceReconciler) setComputedURLAnnotation(ctx context.Context, svc *c
...
+	// LEGACY-PREFIX-MIGRATION (read-side cleanup): drop this delete once no
+	// pre-migration-stamped Services remain.
+	_, hadLegacy := a[annotations.LegacyComputedURLAnnotation]
+	if a[annotations.ComputedURLAnnotation] == computedURL && !hadLegacy {
 		return nil
 	}
 	a[annotations.ComputedURLAnnotation] = computedURL
-	a[annotations.LegacyComputedURLAnnotation] = computedURL
-	// LEGACY-PREFIX-MIGRATION: END
+	delete(a, annotations.LegacyComputedURLAnnotation)
```
jackmcclary-ngrok, line 391: I think we want the `// LEGACY-PREFIX-MIGRATION (read-side cleanup): END` here because we will eventually want to delete this line with cleanup as wll

internal/domain/manager_test.go
```diff
@@ -153,10 +153,7 @@ func assertDomainLabels(t *testing.T, c client.Client, name, namespace string, e
 	t.Helper()
 	var domain ingressv1alpha1.Domain
 	require.NoError(t, c.Get(t.Context(), types.NamespacedName{Name: name, Namespace: namespace}, &domain))
-	domainLabels := domain.GetLabels()
-	for k, v := range expectedLabels {
-		assert.Equal(t, v, domainLabels[k])
-	}
+	assert.Equal(t, expectedLabels, domain.GetLabels())
```
jackmcclary-ngrok, line 156: This implicitly restricts the test to require strict ordering on the elements. Is that intended?

jackmcclary-ngrok (approved):
2 comments, but LGTM

internal/domain/manager_test.go
alex-bezek: It looks like [assert.Equal](https://github.com/stretchr/testify/blob/959dbdacf1533e155162811ea90c90117a420463/assert/assertions.go#L496-L505) does a [reflect.DeepEqual](https://github.com/stretchr/testify/blob/959dbdacf1533e155162811ea90c90117a420463/assert/assertions.go#L63-L71) internally which shouldn't care about the ordering of the keys in the map

internal/controller/service/controller.go
alex-bezek: oops, good call, i'll add these for the next phase

helm/ngrok-operator/values.yaml
```diff
@@ -258,7 +258,7 @@ credentials:
 ##
 ingress:
   enabled: true # enabled by default
-  controllerName: "k8s.ngrok.com/ingress-controller"
+  controllerName: "ngrok.com/ingress-controller"
```
alex-bezek, line 261: I'm testing this more and realizing the IngressClass has this field as immutable and can't be edited. I don't have a super great answer for how we could do this change. I think we would have to require the user to halt their operator, delete the ingress class, remake it with the new value, restart the operator, and then do the rollout. If they don't stop the operator manager pods first, I believe it would remove all their infra tied to that ingress class. 

We could try to automate that with an upgrade hook but that only applies in certain situations. I wonder if the juice is worth the squeeze here as its a pretty manual step for every user, potentially dangerous, and buys pretty much nothing. @jonstacks curious your thoughts here on if this is worth pursuing for completeness or if we could leave it around.

PR #874, opened 2026-08-14
Cleanup: remove legacy IPPolicy RulesConfigured condition handling (K8SOP-307)

alex-bezek:
## Summary
Undocumented-in-`passivity-shims.md` but tagged `LEGACY-CONDITION-MIGRATION` cleanup for the `RulesConfigured` -> `IPPolicyRulesConfigured` status condition rename.

- Deleted the `legacyConditionIPPolicyRulesConfigured` const and the `meta.RemoveStatusCondition` call in `setIPPolicyRulesConfiguredCondition`.

Why this is safe now: `conditions.Set` (`meta.SetStatusCondition`) only upserts the condition type it's given — it doesn't prune the rest of the slice, so this isn't "self-healing" in a general sense. What actually did the cleanup was the explicit `RemoveStatusCondition` call added alongside the original rename: every IPPolicy had the legacy condition stripped on its first reconcile after that release. Enough reconciles have happened since that no live object should still carry it, so the removal call has served its purpose and can go.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `golangci-lint run` clean
- [x] `go test ./...` green (removed the now-redundant `TestSetIPPolicyRulesConfiguredCondition_ClearsLegacyCondition` test)

K8SOP-307

## Summary by CodeRabbit

* **Refactor**
  * Removed the deprecated legacy IP policy status condition.
  * IP policy status now uses only the current rules-configured condition.
  * Aligned IP policy status reporting with the current condition naming, providing a more consistent status experience for users and tools that monitor IP policy configuration.
  * Removed outdated references associated with the legacy condition.

### DIFF
--- a/internal/controller/ingress/ippolicy_conditions.go
+++ b/internal/controller/ingress/ippolicy_conditions.go
@@ -14,10 +14,6 @@ const (
 	ConditionIPPolicyCreated         = "IPPolicyCreated"
 	ConditionIPPolicyRulesConfigured = "IPPolicyRulesConfigured"
 
-	// LEGACY-CONDITION-MIGRATION: delete this once all IPPolicies have been
-	// reconciled at least once past the ConditionIPPolicyRulesConfigured rename.
-	legacyConditionIPPolicyRulesConfigured = "RulesConfigured"
-
 	// condition reasons for IPPolicy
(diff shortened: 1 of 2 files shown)


jackmcclary-ngrok (approved):
LGTM

PR #886, opened 2026-08-31
Upgrade ngrok-api-go to v9 and filter reserved domains server-side

jonstacks:
## What
Upgrades `ngrok-api-go` from v7 to v9, and uses ngrok's [API filtering](https://ngrok.com/docs/api/api-filtering) to look up reserved domains by hostname.

## How
- Bump the dependency to v9. `List()` on resources that support CEL filtering now takes `*ngrok.FilteredPaging`, so `ngrokapi.Lister` is split into `Lister` and `FilteredLister`, and the shared mock base client is generic over the paging type.
- v9 switches JSON tags from `omitempty` to `omitzero`, so a non-nil empty slice is serialized instead of dropped. Guard the few call sites that could send one on a KubernetesOperator or CloudEndpoint create/update, so behavior is unchanged.
- Regenerate the BoundEndpoint CRD. `ngrok.Ref` is inlined into `status.endpoints[]`, and since its fields now use `omitzero`, controller-gen marks `id` and `uri` as required. The operator always populates both from the ngrok API, so what it writes is unchanged.
- `findReservedDomainByHostname` now filters on an exact hostname match server-side, instead of paging through every reserved domain on the account and filtering client-side.

## Breaking Changes
None for users. The only schema change is the required `id`/`uri` noted above, on operator-owned status fields.

## Summary by CodeRabbit

* **Enhancements**
  * Updated the ngrok API integration to the latest version for improved compatibility.
  * Added efficient server-side filtering for domain, IP policy, and TCP address lookups.
  * Improved request handling by omitting empty binding and selector values when appropriate.

* **Bug Fixes**
  * Corrected empty-result handling for API requests.
  * Updated the BoundEndpoint status schema to require endpoint IDs and URIs.

### DIFF
--- a/api/bindings/v1alpha1/boundendpoint_types.go
+++ b/api/bindings/v1alpha1/boundendpoint_types.go
@@ -25,7 +25,7 @@ SOFTWARE.
 package v1alpha1
 
 import (
-	v6 "github.com/ngrok/ngrok-api-go/v7"
+	"github.com/ngrok/ngrok-api-go/v9"
 	ngrokv1alpha1 "github.com/ngrok/ngrok-operator/api/ngrok/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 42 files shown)


jackmcclary-ngrok (approved):
LGTM one clarifying question for myself: ngrok.Paging vs ngrok.FilteredPaging is decided based on CEL support. Is that correct?

jonstacks:
> LGTM one clarifying question for myself: ngrok.Paging vs ngrok.FilteredPaging is decided based on CEL support. Is that correct?

Thats correct. The ngrok-api-go client is generated and knows which types currently support filtered paging https://github.com/search?q=repo%3Angrok%2Fngrok-api-go+FilteredPaging&type=code

PR #887, opened 2026-09-02
Add wildcard domain support to the Domain controller

jonstacks:
## What
Today the Domain controller reserves every `Domain` resource individually with the ngrok API, even when a wildcard reservation (e.g. `*.example.com`) already covers that hostname (e.g. `a.example.com`). This wastes reservations and doesn't reflect that the wildcard's DNS record and certificate already serve the subdomain.

## How
- When reconciling a `Domain`, the controller now checks whether an existing wildcard reservation already covers the hostname. If so, it skips reserving the domain individually and instead records `status.coveredByWildcardDomain` with the covering wildcard (e.g. `*.example.com`).
- Exact reservations always win over a covering wildcard, and any `Domain` with an explicit `resolvesTo` target always gets its own reservation regardless of wildcard coverage.
- Wildcard-covered `Domain`s are rechecked hourly (in case the wildcard is later removed) and are never deleted on their own, since the underlying reservation is shared across every subdomain under the wildcard.
- Domain lookups now issue a single server-side CEL filter (enabled by the ngrok-api-go v9 upgrade) to find both the exact and wildcard candidates in one `List` call, instead of separate client-side lookups.
- Added a real CEL filter implementation to the `nmockapi` test doubles so this filtered `List` behavior can be exercised in unit tests without hitting the live ngrok API.
- New CRD field `status.coveredByWildcardDomain` plus a new printer column (`Wildcard`) on `Domain`, and a CEL validation rule ensuring `status.id` and `status.coveredByWildcardDomain` are never both set.

## Breaking Changes
None. This is additive: existing `Domain`s with individual reservations are unaffected, and the new status field/printer column are purely additive to the CRD.

## Summary by CodeRabbit

* **New Features**
  * Added support for hostnames covered by existing wildcard reservations without creating separate reservations.
  * Covered domains now report wildcard ownership, DNS, certificate, and readiness status.
  * Added a `Wildcard` column to Domain listings.
  * Ingress, Gateway, AgentEndpoint, CloudEndpoint, and Service resources continue reporting covered hostnames correctly.

* **Bug Fixes**
  * Domain hostnames are now immutable after creation.

* **Documentation**
  * Documented wildcard matching, readiness, reservation, and status behavior.


internal/mocks/nmockapi/filter_internal_test.go
```diff
@@ -0,0 +1,142 @@
...
+
+	"github.com/ngrok/ngrok-api-go/v9"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPagingFilter(t *testing.T) {
+	t.Run("FilteredPaging with a filter", func(t *testing.T) {
+		assert.Equal(t, `obj.domain == "a.com"`,
+			pagingFilter(&ngrok.FilteredPaging{Filter: new(`obj.domain == "a.com"`)}))
```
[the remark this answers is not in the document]
jonstacks: This is valid go syntax as of version 1.26.

internal/controller/ingress/domain_controller.go
```diff
@@ -230,27 +283,74 @@ func (r *DomainReconciler) delete(ctx context.Context, domain *v1alpha1.Domain)
...
 func (r *DomainReconciler) updateStatus(ctx context.Context, domain *v1alpha1.Domain, ngrokDomain *ngrok.ReservedDomain, createErr error) error {
 	if ngrokDomain != nil {
 		domain.Status.ID = ngrokDomain.ID
 		domain.Status.Domain = ngrokDomain.Domain
+
+		// This Domain now holds a reservation of its own, so any previously
+		// recorded wildcard coverage no longer applies. Leaving it set would both
+		// misreport the domain and, via the delete() guard, orphan this
+		// reservation in the account when the CR goes away.
+		domain.Status.CoveredByWildcardDomain = ""
```
alex-bezek, line 352: Woult it be worth doing this outside the IF check so it always clears the wildcard status even if the new domain wasn't able to be created? Thats a bit of an edge case, but in theory we should only enter this block if the domain is determined to no longer be covered by a wildcard right?

internal/controller/ingress/domain_conditions.go
```diff
@@ -76,22 +82,33 @@ func updateDomainConditions(domain *ingressv1alpha1.Domain, ngrokDomain *ngrok.R
...
 
-	setDomainCreatedCondition(domain, true, ReasonDomainCreated, "Domain successfully reserved")
+	// The domain is provisioned in the platform either way: via its own
+	// reservation, or via the wildcard that covers it.
+	readyReason := ReasonDomainActive
+	readyMessage := "Domain ready for use"
+	if coveredByWildcard {
+		readyReason = ReasonCoveredByWildcardDomain
+		readyMessage = fmt.Sprintf("Domain ready for use (covered by wildcard domain %s)", domain.Status.CoveredByWildcardDomain)
+		setDomainCreatedCondition(domain, true, ReasonCoveredByWildcardDomain,
```
alex-bezek, line 92: nit: what do you think about making the Created condition false with the message that its covered so not created in the ngrok api?

api/ingress/v1alpha1/domain_types.go
```diff
@@ -112,6 +114,16 @@ type DomainStatus struct {
...
 
+	// CoveredByWildcardDomain is the wildcard reservation that already serves
+	// this hostname, e.g. "*.example.com" for "a.example.com". When set, the
+	// operator intentionally did not reserve this domain in the ngrok API
+	// because the wildcard's DNS record and certificate already cover it.
+	//
+	// ID stays empty in this case: it is the handle the operator would delete,
+	// and the wildcard reservation is shared with every other subdomain under it.
+	// +optional
+	CoveredByWildcardDomain string `json:"coveredByWildcardDomain,omitempty"`
```
alex-bezek, line 125: would there be any value in adding the ngrok wildcard domain id here, maybe if it were an object with the domain string and id? 
I'm not sure what we would use the ID for, maybe simpler lookups but we do have cel filtering now. An object would let us add things in the future more passively. Just a thought

internal/controller/ingress/domain_controller.go
```diff
@@ -156,33 +164,70 @@ func (r *DomainReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{Requeue: true}, nil //nolint:staticcheck
 	}
 
+	// A wildcard-covered Domain holds no reservation of its own, so nothing in
+	// Kubernetes changes if the wildcard reservation disappears from the account,
+	// and the watch predicate only fires on annotation/generation changes.
+	// Re-check on an interval: status.id is empty, so the recheck goes back
+	// through create(), which reserves the domain itself once the wildcard is gone.
+	if domain.Status.CoveredByWildcardDomain != "" {
+		return ctrl.Result{RequeueAfter: wildcardCoverageRecheckInterval}, nil
```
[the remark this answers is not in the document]
alex-bezek: hmm I think this is right and unfortunate. I believe fixing those create events in the base controller is something we have on our Events cleanup ticket since we also get them just on startup i believe. 
We could probably use a runnable to poll for a missing wildcard domain and update the Domain CRs if a wildcard goes away triggering a reconcile of them. Or we can ship this and come back to the events as a whole.

specs/controllers/domain.md
```diff
@@ -70,8 +82,60 @@ The `spec.reclaimPolicy` field controls what happens to the ngrok domain reserva
...
+
+The Domain CR is still created — it is the object other controllers read
+`status.cnameTarget` from — but the reconciler skips the ngrok API reservation.
+
+**Matching is single-label**, mirroring DNS and TLS wildcard semantics:
+
+| Hostname | Covered by | Notes |
+|----------|-----------|-------|
+| `a.mydomain.com` | `*.mydomain.com` | direct child |
+| `a.b.mydomain.com` | `*.b.mydomain.com` | **not** `*.mydomain.com` |
```
alex-bezek, line 111: is there a weird edge case here where they could have the TLD wildcard reserved and use nested subdomains and just won't care that certs aren't available and use http? I'm just remembering when i made the assumption that a domain needed its cert to be "ready and useable" and that was wrong

internal/controller/ingress/domain_controller.go
```diff
@@ -156,33 +164,70 @@ func (r *DomainReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{Requeue: true}, nil //nolint:staticcheck
 	}
 
+	// A wildcard-covered Domain holds no reservation of its own, so nothing in
+	// Kubernetes changes if the wildcard reservation disappears from the account,
+	// and the watch predicate only fires on annotation/generation changes.
+	// Re-check on an interval: status.id is empty, so the recheck goes back
+	// through create(), which reserves the domain itself once the wildcard is gone.
```
alex-bezek, line 171: Is this just to make sure the status correctly matches and you auto-reserve a bunch of domains if the wildcard goes away, or is there a user experience workflow you are envisioning that this fixes?
Is this something where if you make the wildcard and then a domain, delete the wildcard, and then try to create an Ingress using that domain, nothing is going to cause a re-reconcile of the domain and make it actually create the domain, causing the endpoint to not create?

Depending on the workflow, I wonder if the polling every hour is worth it. they should get caught by the sweep the operator does every 8 hours. And if the workflow is something quick like delete then create, 1 hour might not help much. 1 api call every hour though is pretty cheap so its probably fine

alex-bezek (approved):
Some nits and minor thoughts/suggestions, nothing blocking

PR #889, opened 2026-09-08
Make Domain.spec.domain immutable

jonstacks:
## What

Changing `spec.domain` on an existing, reserved `Domain` is silently ignored.

`update()` fetches the reservation by `status.ID` — still the old domain — and its change detection compares only `description`, `metadata`, and `resolvesTo`, never `spec.domain` against the response. It concludes "no changes needed" and writes the old reservation back to status, so the resource reports `Ready=True` with `observedGeneration` stamped to the new generation. The new domain is never reserved and `spec.domain != status.domain` is the only tell.

This can't be honored in the reconciler: `ngrok.ReservedDomainUpdate` has no domain field, because reserved domains can't be renamed via the ngrok API. Supporting it would mean releasing the reservation and re-reserving, risking losing the domain in the gap.

## How

Reject the change at admission with a CEL transition rule on `spec.domain`. Nothing guarded it before — the generated CRD had no `x-kubernetes-validations` at all, and no admission webhook is registered for `Domain`.

Only direct CR edits are affected. The operator's own paths name the CR after `HyphenatedDomainNameFromURL(domain)`, so a different hostname is a different object — a create, not a mutation.

That rule is also what the driver's own `spec.domain` rewrite hits, if a Domain drifted before it shipped. `applyDomains` now treats an `Invalid` rejection on a single Domain as non-fatal; otherwise one unfixable Domain fails the errgroup, returns early from `Sync`, and wedges reconciliation for every Ingress and Gateway in the cluster. All other error classes stay fatal.

Specs updated, including a scope note on the existing "Defer validation to the ngrok API" decision, which explicitly ruled out CEL x-validation. That decision covers whether a *value* is acceptable to ngrok; deferring can't work here, because the API exposes no rename, so no request is ever sent and no error ever comes back to surface.

Tests: envtest spec for the rejection (against the real generated CRD), and two `applyDomains` unit tests for the guard. Both verified to fail without their fix.

## Breaking Changes

Yes, for anyone editing `spec.domain` today — that edit is now rejected instead of silently dropped. Moving to a different domain means creating a new Domain resource for it; the old one does not need to be deleted, and if it is, `spec.reclaimPolicy` governs only what happens to its reservation in ngrok.

A Domain already drifted before this ships stays valid: CEL transition rules compare against the same object's previous value, so an unchanged drifted resource passes and keeps reporting `Ready` against the old reservation. To find any:

```
kubectl get domains -A -o json | jq -r '.items[] | select(.status.domain != "" and .spec.domain != .status.domain) | "\(.metadata.namespace)/\(.metadata.name): spec=\(.spec.domain) status=\(.status.domain)"'
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Domain values can no longer be changed after creation; invalid updates are rejected before being applied.
  * Invalid domain updates no longer block synchronization of other valid domains.

* **Documentation**
  * Clarified that changing a domain requires creating a replacement resource; deleting the existing resource is optional.
  * Documented how reclaim policy affects reservation cleanup.
  * Added guidance on domain resource naming and immutability.


jackmcclary-ngrok:
So previously it worked such that updating the CR spec.Domain would make the user think that it was changing the domain that it was connected to in the ngrok backend API, but it wouldn't actually change anything. Now it fails loudly? Did I understand the scenario correctly?

jackmcclary-ngrok (approved):
If my understanding from my previous comment is correct then looks good!

jonstacks:
Yes, that is correct. We could allow changing the domain and handle it in the controller by checking what the existing status ID points to. I think that opens up questions on what we do with the old domain. Do we follow the reclaimPolicy and either delete it or retain it and then create a new one? Do we always delete it?

It seemed more correct to me that the reclaim policy is on the domain. In either case, it is subtly bugged right now.

